PR #2, opened 2015-02-10
test travis pr

actionless:


ioanpocol:
:+1: 
OK, so the test on server&client runs on every PR.

petrjasek:
yep

PR #9, opened 2015-02-16
(Fix)[Ingest] Added FTP passive switch for connection to the server.

mdhaman:
@petrjasek ftp code change did not work expected. I was getting error as "200 Type is Image (Binary)" on line https://github.com/superdesk/superdesk/blob/2bca01c2999b01da03a9c32a764a90a2e9d73e95/server/superdesk/io/ftp.py#L88.

The ftp.retrbinary command does the following in passive mode:
- send "Type I"
- send PASV 
The response of the PASV command is "200 Type is Image (Binary)" instead of "227 Entering Passive Mode" Hence response validation of PASV fails. FTP download in active mode work fine. 

Moreover, I have modified the client and server so that the FTP connection type is configurable.


petrjasek:
:+1:

amagdas:
:+1:

PR #11, opened 2015-02-16
chore(travis): add e2e tests

actionless:


actionless:
@petrjasek @amagdas 
can u take a look on build logs?
i tried with both chrome and firefox but anyway from 2 to 5 e2e tests are failing from time to time

screenshots of failed tests:
http://curl.io/get/cyusjvxu/0dbc509e10c16dc820f5609137c39cd72d7f4e5c

client/package.json
```diff
@@ -38,7 +38,7 @@
     "karma-ng-html2js-preprocessor": "^0.1.0",
     "lodash": "^2.4.1",
     "karma-coverage": "^0.2.0",
-    "protractor": "~1.6.1",
+    "protractor": "actionless/protractor#old-chrom",
```
amagdas: With latest protractor are we having more errors?

actionless: it's the latest protractor but the thing what chromedriver version is hardcoded and in order to use travis we need older version of chromedriver (travis have chromium 37 installed while newer chromedriver requires 38)

and i docker-based travis workers we can't do operations which require super user rights so it's not possible to install newer chrome

actionless:
@plamut come and join us

client/spec/setup.js
```diff
@@ -10,7 +10,7 @@ beforeEach(function(done) {
     getToken(function() {
         resetApp(function() {
             browser.driver.getCurrentUrl().then(function(url) {
-                if (url.indexOf('data:') !== 0) {
+                if (url.indexOf('http') === 0) {
```
actionless: that's for compatibility with firefox

plamut:
Can this be reproduced locally? Or does the issue only appear on Travis?

actionless:
only on travis, but now i am trying to fully reproduce the environment (chrome and chromedriver versions at least) to check if the problem can be related to any of them

actionless:
@plamut from first checks i have an impression what it can be chromedriver version, but an exception is different than here:

```
17-Feb-2015 10:35:11 	Uncaught exception: Timed out waiting for Protractor to synchronize with the page after 11 seconds. Please see https://github.com/angular/protractor/blob/master/docs/faq.md
```

if it would be clear that chromedriver is the root cause when it seems like we will have no other choice than stick back to openVZ-based builds (they're slower a bit)

actionless:
i will merge Petr's changes and try again

plamut:
OK, thanks for the update(s).

EDIT: it seems tha the build has just passed?

petrjasek:
I think it would be good to put this shell code from `.travis.yml` into `.sh` files so we could run it locally

actionless:
oops, just noticed what because of some networking glitch my comment was not being sent:

sequence for running e2e tests relies on travis build lifetime hooks (running xvfb in before_install and starting servers in before_script) -- moving it outside as separate scripts will not be very helpful so i better recommend using this tool for local reproduce of travis build: 
https://github.com/grosser/wwtd

actionless:
@petrjasek also as a separate script it's presented as part of docker pr:
https://github.com/actionless/superdesk/blob/add-docker-ci-scripts/docker/run_e2e_tests.sh

petrjasek:
ok makes sense

petrjasek:
:+1:

actionless:
@amagdas please review as well, it need to be merged asap

amagdas:
:+1:

PR #15, opened 2015-02-16
fix(archive): Headline and container values for images in grid view

sjunaid:
SD-1454

### DIFF
--- a/client/app/scripts/superdesk/filters.js
+++ b/client/app/scripts/superdesk/filters.js
@@ -77,6 +77,11 @@ define([
                 }
             };
         })
+        .filter('truncateString', function() {
+            return function(inputString, limit, postfix) {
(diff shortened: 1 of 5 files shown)


eozan:
:+1:

client/app/scripts/superdesk/filters.js
```diff
@@ -77,6 +77,11 @@ define([
                 }
             };
         })
+        .filter('truncateString', function() {
+            return function(inputString, limit, postfix) {
+                return _.trunc(inputString, {'length': limit, 'omission': postfix});
```
amagdas: We can supply a default value for 'omission' so we don't have to supply a value for it each time...

amagdas:
:+1:

mugurrus:
:+1:

ioanpocol:
:+1:

PR #16, opened 2015-02-17
feat (routing rules) - Backend: Endpoint to manage the routing schemes

sivakuna-aap:


petrjasek:
:+1: 
this unique name check could be case insensitive but not sure if that's required

amagdas:
:+1:

PR #18, opened 2015-02-17
Routing UI

eozan:


eozan:
todo:
- markup and style fix

client/app/scripts/superdesk-ingest/module.js
```diff
@@ -543,9 +742,13 @@ define([
 
     app
         .service('ingestSources', IngestProviderService)
+        .service('subjectService', SubjectService)
```
petrjasek: it seems more like a factory and not service

petrjasek:
there are some strings in templates without translate annotation

PR #19, opened 2015-02-17
Add RSS ingest provider (SD 1525)

plamut:


server/superdesk/io/rss.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def _create_item(self, data):
+        """Create a new content item from RSS feed data.
+
+        :param dict data: parsed data of a single feed entry
+        :return: created content item
+        :rtype: dict
+        """
+        # TODO: what keys should this item contain?
+        item = dict()
```
petrjasek: there should be also `item['type'] = 'text'`

plamut: Added.

petrjasek:
:+1:

plamut:
UPDATE: I refined the error types raised by the RSS ingest service.

IMO work is now done, except for the required CSS changes of course. It should be easy for the design team to add those tomorrow when they are back to work.

server/superdesk/errors.py
```diff
@@ -312,3 +312,33 @@ def ftpUnknownParserError(cls, exception, provider, filename):
             logger.exception("Provider: {} - File: {} unknown file format. "
                              "Parser couldn't be found.".format(provider.get('name', 'Unknown provider'), filename))
         return IngestFtpError(5001, exception, provider)
+
+
+class IngestRssFeedError(SuperdeskIngestError):
```
petrjasek: not sure we need rss specific error class, eg. auth error would make sense also for reuters where you have to authenticate. also connection timeout would work for http apis, not found too.

plamut: I can move the errors to one of the existing classes, the `IngestApiError` class seems like the best candidate to me?

petrjasek: :+1:

plamut:
RSS ingest-specific errors now moved/merged with `IngestApiError` class.

petrjasek:
:+1:

amagdas:
Rebase from master is needed...

server/superdesk/io/rss.py
```diff
@@ -0,0 +1,128 @@
...
+
+PROVIDER = 'rss'
+
+utcfromtimestamp = datetime.utcfromtimestamp
+
+
+class RssIngestService(IngestService):
+    """Ingest service for providing feeds received in RSS 2.0 format.
+
+    (NOTE: it should also work with other syndicated feeds fromats, too, since
```
amagdas: Typo, s/fromats/formats

plamut:
@amagdas  Heh, I'm rebasing this almost all the time. ;)
I'll rebase again, along with the fix for that typo you found in the comment.

plamut:
Can we remove the WIP label here? Because the only thing missing are a few CSS changes, and that is covered by another ticket.

amagdas:
:+1:

PR #21, opened 2015-02-17
fix(celery): make sure workers won't hang with emails

petrjasek:
without configured mail server celery workers get stuck with email
sending tasks. this will make such tasks fail after 10s


amagdas:
:+1:

ioanpocol:
:+1:

PR #24, opened 2015-02-17
[SD-1586] - Improve visual feedback for the fetch action

akintolga:
* Now the fetch button will be replaced by the progress icon during the fetching process both in ingest and in content.
* If the fetch fails, user will be notified with the error message
* Progress icon has also been used in spike action in content 
* A bug is fixed in sending the notification from the archive_ingest end point
* Some re-factoring in archive_ingest end point is done to improve the performance of fetch action
* Another bug is fixed: "seeing the fetch watermark in content items"


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -79,8 +80,12 @@ define([
         $rootScope.currentModule = 'ingest';
 
         this.fetchItems = function(criteria) {
+            $scope.loading = true;
             api.ingest.query(criteria).then(function(items) {
                 $scope.items = items;
+                $scope.loading = false;
+            }, function(response) {
```
petrjasek: fyi you can use [promise.finally](https://docs.angularjs.org/api/ng/service/$q#the-promise-api)

client/app/scripts/superdesk-archive/module.js
```diff
@@ -79,6 +83,7 @@ define([
                 .activity('spike', {
                     label: gettext('Spike Item'),
                     icon: 'remove',
+                    monitor: 'true',
```
petrjasek: seems like this could be boolean and not string

server/apps/archive/archive_ingest.py
```diff
@@ -110,7 +113,8 @@ def create(self, docs, **kwargs):
                 superdesk.get_resource_service(ARCHIVE).post([dest_doc])
                 insert_into_versions(dest_doc.get('guid'))
 
-                push_notification('item:fetch', item=str(ingest_doc.get('_id')))
+        if(kwargs.get('notify', True)):
```
petrjasek: hey it's not javascript ;)

petrjasek:
:+1:

server/apps/archive/archive_ingest.py
amagdas: :smile: This was bound to happen :smile_cat:

akintolga: It is all single repo problem. Before that I would never us js in python :bowtie:

amagdas: Yup, and it's gonna happen again :smile: Probably it will be me, we can do a bet on this :8ball:

server/apps/archive/common.py
```diff
@@ -184,7 +184,7 @@ def remove_unwanted(doc):
     """
 
     # _type attribute comes when queried against Elastic and desk comes while fetching an item from ingest
-    for attr in ['_type', 'desk']:
+    for attr in ['_type', 'desk', 'archived']:
         if attr in doc:
```
sivakuna-aap: We might not need this change as this attribute is removed in `archive_ingest` module already.

akintolga: This was needed to stop existing content items having 'archived' field duplicated with that field again.

sivakuna-aap: True but this is already taken care in archive_ingest (line: 72) module right?

akintolga: I want to avoid the existing 'wrong' data in content in master-sd being proliferated by duplication. After all the already fetched data with wrong value disappear we can revisit this.

sivakuna-aap: :+1:

amagdas:
:+1:

petrjasek:
:+1:

PR #27, opened 2015-02-18
feat(publish) [SD-1549] Backend: Create a publish endpoint for publishing items.

sivakuna-aap:
Also a user shouldn't be allowed to update after publishing.
[SD-1545] Added a scenario for testing uniqueness of routing scheme names

### DIFF
--- a/client/app/scripts/superdesk-desks/directives.js
+++ b/client/app/scripts/superdesk-desks/directives.js
@@ -358,8 +358,11 @@ define([
                 desk: '=',
                 stage: '='
             },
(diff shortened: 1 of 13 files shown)


petrjasek:
I think it might be better to make publish a separate module.. archive gets a bit fat (and it shouldn't be archive anyhow)

amagdas:
I agree with @petrjasek  :+1:

sivakuna-aap:
Refactored publish endpoint to a new package

amagdas:
:+1:

petrjasek:
:+1:

PR #28, opened 2015-02-18
Sd 1558

sjunaid:


eozan:
:+1:

client/app/scripts/superdesk-desks/directives.js
```diff
@@ -189,21 +187,29 @@ define([
...
-                        }, function(response) {
-                            scope.message = gettext('There was a problem, stage was not saved.');
-                        });
+                        }, errorMessage);
                     }
                 };
 
+                function errorMessage(response) {
+                    if (response.data && response.data._issues && response.data._issues.name && response.data._issues.name.unique) {
+                        scope.message = gettext(
```
petrjasek: this message can't be translated - it would require a defined translation for each stage name..
better set some `scope._error = true;` and in template:
```
<span ng-show="_error" translate>Stage with name "{{ editStage.name }}" already exists.</span>
```
translate will first replace the string with translation and then it will eval `{{ editStage.name }}`

sjunaid:
@petrjasek ,can you please review the suggested changes. thanks

ioanpocol:
In order to pass all tests a rebase is needed.

petrjasek:
:+1:

amagdas:
:+1:

PR #33, opened 2015-02-18
fix(ingest): set 5m soft time limit for all tasks

petrjasek:
but keep it 10s for emails, and make it 30m for ingest.update_provider
+ set timeouts for reuters api requests


amagdas:
:+1:

ioanpocol:
:+1:

mugurrus:
:+1:

PR #44, opened 2015-02-19
chore(travis, client): install patched protractor for travis on-demand

actionless:
should speed-up client build (not installing protractor from git) and hopefully should fix e2e on bamboo (it seems chromedriver 2.13 works there better than 2.12 on bamboo)


amagdas:
:+1:

petrjasek:
:+1:

PR #47, opened 2015-02-19
Local demo

actionless:


actionless:
i am not sure about moving vhost from docker to client or not

petrjasek:
I think it would make sense to run single nginx as proxy for server and also serving client files, then it should be outside both client and server

actionless:
@petrjasek but currently it's actually like that

petrjasek:
ah right so then imo it shouldn't be part of client

actionless:
so what is your idea, to swap dockerfiles back and make nginx-less Dockerfile default?

petrjasek:
I'm not sure, we can keep it like this.

PR #50, opened 2015-02-19
packages - add liveblog extensibility point

amagdas:


amagdas:
@ancafarcas This is the one :smile_cat:

petrjasek:
just wondering - why do we this linked in packages field?

amagdas:
We need to show if an item is part of a package or not, this is to be easily available, in some designs/stories is to be shown on each item.
We can do it like the current implementation or check on each fetch action, I think the current implementation is more efficient...

petrjasek:
I was thinking if we have to have this `linked_in_packages` given that we can search by `groups.refs.residRef` if that would be properly mapped..

amagdas:
We can search, but as I said, that would mean for each item fetch from archive to do this search and this will hit on performance quite a bit...

petrjasek:
I see didn't realise we do it in a loop :+1:

ioanpocol:
:+1:

PR #54, opened 2015-02-19
Fix bower for docker

actionless:
\+ changes from https://github.com/superdesk/superdesk/pull/47


petrjasek:
does it work for you @vied12 ?

vied12:
It works ! :+1:

PR #55, opened 2015-02-19
Highlights: Fixes and Improvements

nidzix:
SD-1244
+ Add ui for managing highlights configurations
+ Fix styling issues with highlight dropdowns
+ Improve hilight inclusion list directive for content items

### DIFF
--- a/client/app/scripts/superdesk/ui/ui.js
+++ b/client/app/scripts/superdesk/ui/ui.js
@@ -252,6 +252,29 @@ define([
 		};
 	}
 
+	DropdownPositionDirective.$inject = ['$document'];
+	function DropdownPositionDirective($document) {
+		return {
+			link: function(scope, element) {
+
+				var tolerance = 250;
+
+				element.bind('click', function(event) {
+					if (closeToBottom(event)) {
+						element.addClass('dropup');
+					} else {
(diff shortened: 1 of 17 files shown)


amagdas:
:+1:

client/app/scripts/superdesk/ui/ui.js
```diff
@@ -252,6 +252,29 @@ define([
...
 	}
 
+	DropdownPositionDirective.$inject = ['$document'];
+	function DropdownPositionDirective($document) {
+		return {
+			link: function(scope, element) {
+
+				var tolerance = 250;
+
+				element.bind('click', function(event) {
```
petrjasek: not sure here, but don't you have to unbind it?

petrjasek:
:+1:

client/app/scripts/superdesk/ui/ui.js
nidzix: I think that only those handlers attached to the element **diferent than the one linked to directive**, should be unbinded. For example if I attach `click` event to `document`, here,  in this directive, I would need to unbind it.

petrjasek: ok couldn't find it in docs so I'm just checking

PR #61, opened 2015-02-20
SD-1620 : Wrong error message when ingesting to personal

ioanpocol:
Added support to fetch into Personal


mugurrus:
:+1:

amagdas:
:+1:

petrjasek:
I don't think it should be supported.. There is a requirement to prevent this.

sivakuna-aap:
I agree with @petrjasek . The ticket is related to changing the error message language.

ioanpocol:
The user is not allowed to fetch in personal space

PR #64, opened 2015-02-20
(feat) Routing Rules - Implementation for routing rules for ingest.

mdhaman:
SD-1548


server/apps/rules/routing_rules.py
```diff
@@ -69,10 +72,10 @@ class RoutingRuleSchemeResource(Resource):
                             'allowed': ['SUN', 'MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT']
```
amagdas: This list should be declared only once and reused.

sivakuna-aap:
:+1:

amagdas:
Rebase needed.

server/apps/rules/routing_rule_validator.py
```diff
@@ -0,0 +1,62 @@
...
+import logging
+import re
+
+logger = logging.getLogger(__name__)
+
+
+class RoutingRuleValidator():
+
+    def is_valid_rule(self, item, rule_filter):
+        if not filter:
```
petrjasek: seems like it should be `rule_filter`, `filter` is always defined

petrjasek:
:+1: but would be good to catch some specific exceptions in those try/catch blocks

server/apps/rules/routing_rule_validator.py
sivakuna-aap: `filter` is not mandatory property. When `filter` is not specified it the rule will match all the ingested articles associated with an ingested provider. For example, a rule can be something like this All articles from AAP with an email id should go to email desk.

mdhaman: You are right ```filter``` is always defined. It should be ```rule_filter```.  

@petrjasek  which exceptions are you referring?

server/apps/rules/routing_rules.py
```diff
@@ -155,20 +181,35 @@ def __validate_routing_scheme(self, routing_scheme):
...
+        if schedule:
+            day_of_week = [str(week_day).upper() for week_day in schedule.get('day_of_week', [])]
+            if not (len(set(day_of_week) & set(self.day_of_week)) == len(day_of_week)):
+                return False
+
+            if schedule.get('hour_of_day_from') or schedule.get('hour_of_day_to'):
+                try:
+                    return datetime.strptime(schedule.get('hour_of_day_from'), '%H%M') < \
+                        datetime.strptime(schedule.get('hour_of_day_to'), '%H%M')
+                except:
```
petrjasek: exceptions like this

mdhaman: done..

petrjasek:
:+1:

amagdas:
:+1:

mugurrus:
:+1:

PR #66, opened 2015-02-20
[SPIKE] content macros

petrjasek:
initial implementation.. let me know if it makes sense.
- [x] server side macro registration
- [x] `/macros` api
- [x] client widget macros
- [x] shortcuts for macros on client - it's `ctrl + macro.shortcut`


server/apps/macros.py
```diff
@@ -0,0 +1,60 @@
...
+class MacrosResource(superdesk.Resource):
+    resource_methods = ['GET', 'POST']
+    item_methods = []
+    privileges = {'POST': 'archive'}
+
+    schema = {
+        'macro': {
+            'type': 'string',
+            'required': True,
+            'allowed': (macro['name'] for macro in macros)
```
amagdas: I don't think this really works as the macros are registered after the linking so here `macros` will be empty.
Probably doing a behave test with a wrong macro name will show this.

amagdas:
Good job :heart_decoration: :100:

server/apps/macros.py
petrjasek: that's why it's using generator ;) it's only evaluated when there is a post call

amagdas: That I get :) Still the schema is registered when the Eve app is created, so this lazy evaluation I don't think that it has the effect it should.

petrjasek: it's registered but until there is some iteration over allowed it's only passing references around.

amagdas: If that's the case, I learned something today :+1:

server/features/macros.feature
```diff
@@ -0,0 +1,28 @@
...
+            {"item": {"body_html": "$12"}}
+            """
+
+    @auth
+    Scenario: Return an error when triggering unknown macro
+        When we post to "/macros"
+            """
+            {"macro": "this does not exist!", "item": {"body_html": "test"}}
+            """
+        Then we get response code 400
```
amagdas: What's the error message?

server/settings.py
```diff
@@ -143,7 +143,10 @@ def env(variable, fallback_value=None):
     'apps.packages',
     'apps.privilege',
     'apps.rule_sets',
-    'apps.highlights'
+    'apps.highlights',
+    'apps.macros',
+
+    'macros.currency',
```
akintolga: It looks like we need to restart the server when there's a new macro, is that right?

ioanpocol:
:+1:

server/settings.py
amagdas: We can load new modules dynamically if needed.

petrjasek: yep we can instead specify a folder from where it would autoload all macros

akintolga: I think it needs to be dynamic. So I guess settings.py doesn't need to know about the individual modules.

amagdas:
Rebase needed...

akintolga:
Happy :+1: 

I have updated the wiki page with what we could gather from Michael. I added the "Additional Requirements" section. I think (maybe) we can start using the current implementation and then we can attempt to address the remaining requirements in later milestones. This needs to be discussed with POs, PLs and ORPs (other responsible people :sweat_smile: ).

PR #75, opened 2015-02-24
Item list

eozan:


client/app/scripts/superdesk/itemList/itemList.js
```diff
@@ -0,0 +1,282 @@
...
+var DEFAULT_OPTIONS = {
+    endpoint: 'search',
+    pageSize: 25,
+    page: 1,
+    sort: [{versioncreated: 'desc'}]
+};
+
+angular.module('superdesk.itemList', [])
+.service('itemListService', ['api', function(api) {
+    var getQuery = function(options) {
```
petrjasek: it would be better like `function getQuery()`, seems like there is no need to assign it

client/app/scripts/superdesk/itemList/itemList.js
```diff
@@ -0,0 +1,282 @@
...
+        options = _.extend(DEFAULT_OPTIONS, options);
+        var query = getQuery(options);
+        return api(options.endpoint, options.endpointParam || undefined)
+            .query(query)
+            .then(function(result) {
+                return result;
+            });
+    };
+}])
+.provider('ItemList', function() {
```
petrjasek: why provider?

client/app/scripts/superdesk/itemList/itemList.js
```diff
@@ -0,0 +1,282 @@
...
+                } else {
+                    scope.itemListOptions.types.push(itemType);
+                }
+            };
+
+            scope.isItemTypeEnabled = function(itemType) {
+                return scope.itemListOptions.types.indexOf(itemType) > -1;
+            };
+
+            scope.pin = function(item) {
```
petrjasek: how does this work with different lists?

petrjasek:
pls add some tests..

client/app/scripts/superdesk/itemList/itemList.js
eozan: done

client/app/scripts/superdesk/itemList/itemList.js
eozan: it works fine with content and ingest items. it was an existing widget functionality, so it makes sense to put it in sd-item-list-widget directive.

client/app/scripts/superdesk/itemList/itemList.js
eozan: this is the best way to inject an instantiatable function (without instantiating, so that multiple lists can be created after injection) in my opinion.

client/app/scripts/superdesk/itemList/itemList.js
petrjasek: I'm just wondering how you get only pinned items for ingest in ingest. can't find any filtering..

client/app/scripts/superdesk/itemList/itemList.js
```diff
@@ -0,0 +1,287 @@
...
+        link: function(scope, element, attrs) {
+            scope.items = null;
+            scope.maxPage = 1;
+            scope.pinnedItems = [];
+            scope.selected = null;
+
+            var pinnedList = {};
+            var itemList = new ItemList();
+
+            itemList.addListener(function() {
```
petrjasek: this should be somewhere unregistered

client/app/scripts/superdesk/itemList/itemList.js
petrjasek: :ok:

client/app/scripts/superdesk/itemList/itemList.js
eozan: Added itemPinService to manage pinning items.

client/app/scripts/superdesk/itemList/itemList.js
eozan: there is removeListener, and it can be used to unregister listeners. But in most cases it won't be needed, as it will die with the scope anyway.

petrjasek: but you're using scope in handler, so there is a reference and thus scope can't die..
looks like memory leak..

eozan: now removing listeners on scope's $destroy event.

petrjasek:
:+1:

amagdas:
:+1:

PR #77, opened 2015-02-24
fix(docker): prevent creation of new files via touch

petrjasek:
if there was a file with space in name it would split it
for xargs and touch every part of that filename, creating new files,
and not touching the original file at all

but not sure why it's there @actionless ?

### DIFF
--- a/docker/container_build.sh
+++ b/docker/container_build.sh
@@ -38,9 +38,7 @@ mkdir -p $CLIENT_RESULTS_DIR/unit &&
 mkdir -p $SCREENSHOTS_DIR
 
 # reset repo files' dates:
-cd $BAMBOO_DIR/server/
-find -print0 ./ | grep -vzZ .git/ | xargs -0 touch -t 200001010000.00
-cd $BAMBOO_DIR/client/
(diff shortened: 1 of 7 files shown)


actionless:
it's to reset mtime

docker/container_build.sh
```diff
@@ -39,9 +39,9 @@ mkdir -p $SCREENSHOTS_DIR
 
 # reset repo files' dates:
 cd $BAMBOO_DIR/server/
-find ./ | grep -v .git/ | xargs touch -t 200001010000.00
+find -print0 ./ | grep -vzZ .git/ | xargs -0 touch -t 200001010000.00
 cd $BAMBOO_DIR/client/
-find ./ | grep -v .git/ | xargs touch -t 200001010000.00
+find -print0 ./ | grep -vzZ .git/ | xargs -0 touch -t 200001010000.00
```
actionless: actually it can be deduplicated, ie `cd` to whole `$BAMBOO_DIR` instead of each client and server

actionless:
LGTM

amagdas:
:+1:

PR #86, opened 2015-02-25
[SD-889] - Content expiry per ingest provider

akintolga:


client/app/scripts/superdesk-desks/views/content-expiry.html
```diff
@@ -0,0 +1,11 @@
+<label translate ng-if="preview">{{ContentExpiry.Header}}
```
petrjasek: it would be better like `<label ng-if="preview">{{ ContentExpiry.Header | translate }}`

petrjasek:
:+1:

ioanpocol:
:+1:

PR #92, opened 2015-02-26
doc(install): add details about installation

petrjasek:
add info about installing docker.io, virtualenv and docker-composer


actionless:
LGTM

PR #93, opened 2015-02-26
fix(docker): fix docker check in demo

petrjasek:
there was no output above the message

### DIFF
--- a/scripts/docker-local-demo.sh
+++ b/scripts/docker-local-demo.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 (
-	docker ps >/dev/null && which docker-compose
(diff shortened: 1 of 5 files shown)


scripts/docker-local-demo.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 (
-	docker ps >/dev/null && which docker-compose
+	docker --version && docker-compose --version
```
actionless: it should also check if docker daemon is running, that's why `ps>/dev/null`

petrjasek: I see. well issue was with docker-compose, which returns nothing if it's not there

petrjasek:
part of #94

PR #99, opened 2015-03-02
[SD-810] Email ingest

marwoodandrew:
Given a simple email it will create an item with the headline set to the subject line of the email and the body set to either the plain text or html text of the email.

If the email has image attachments these are ingested and a composite item of the email and images is created.


server/superdesk/io/email.py
```diff
@@ -0,0 +1,56 @@
...
+        self.parser = rfc822Parser()
+
+    def _update(self, provider):
+        server = provider.get('config', {}).get('server', None)
+        if not server:
+            server = ''
+        port = provider.get('config', {}).get('port', None)
+        if not port:
+            port = 993
+
```
sivakuna-aap: We can have something like this `port = provider.get('config', {}).get('port', 993)`

server/superdesk/io/email.py
```diff
@@ -0,0 +1,56 @@
...
+        if rv == 'OK':
+            rv, data = Imap.search(None, provider.get('config', {}).get('filter', None))
+            if rv == 'OK':
+                new_items = []
+                for num in data[0].split():
+                    rv, data = Imap.fetch(num, '(RFC822)')
+                    if rv == 'OK':
+                        new_items.append(self.parser.parse_email(data))
+            Imap.close()
+        Imap.logout()
```
sivakuna-aap: I think variable name should start with small, probably flake might complain...not sure though

server/superdesk/io/email.py
marwoodandrew: We could but when I tried, I failed, will try again

server/superdesk/io/email.py
marwoodandrew: flak was happy, but will change!

server/superdesk/io/rfc822.py
```diff
@@ -0,0 +1,147 @@
...
+                        comp_item['groups'] = groups
+
+        if html_body is not None:
+            item['body_html'] = html_body
+        else:
+            item['body_html'] = text_body
+            item['type'] = 'preformatted'
+
+        # if there is more than 1 ref then it because we added an image
+        if len(refs) > 1:
```
ioanpocol: > 1 is not required

marwoodandrew: It is not a test for an empty list,  "> 1" is good, as it would have a single entry for the text item if it had no image. Any way have re-factored it in the hope it is clearer.

amagdas:
:+1:

sivakuna-aap:
:100:

marwoodandrew:
Cane we remove the WIP tag now! and consider it for a merge?

PR #101, opened 2015-03-02
SD-292: Create new version of a package - protractor tests

mugurrus:


actionless:
@mugurrus firefox failure in e2e is ok

ioanpocol:
:+1:

amagdas:
Cherry-picked this changes to https://github.com/superdesk/superdesk/pull/108, I think we can close this.

ioanpocol:
This is not fixed yet and there will be a lot some changes to fix it.

amagdas:
The tests on firefox still fail, exactly as before, now the build passes as we ignore failures on firefox

mugurrus:
The tests on FF fail because of bad driver, we agreed to ignore the FF tests run for now.

amagdas:
I know, my point was this PR is only actually enabling back the test, the state is the same as before, works on Chrome and fails on FF, not sure the latest added changes are needed.

PR #106, opened 2015-03-03
macros spike

petrjasek:
will load macros from given module
it will reload the module whenever using it to keep the list up to date


ioanpocol:
:+1:

mugurrus:
:+1:

PR #107, opened 2015-03-03
fix(e2e): wait for angular.bootstrap call

petrjasek:
@actionless


amagdas:
:+1:

actionless:
it looks fine but better create branch in superdesk repo to check how it will run in bamboo

petrjasek:
duplicate

PR #108, opened 2015-03-03
refactor(packages) - remove "packages" endpoint, use "archive" instead

amagdas:


ioanpocol:
:+1: 
There are 2 behave tests that should be updated/fixed

server/apps/archive/archive.py
```diff
@@ -128,6 +130,8 @@ def update_word_count(doc):
 
 
 class ArchiveService(BaseService):
+    packageService = PackageService()
```
sivakuna-aap: It might be better to move this to `__init__()`

ioanpocol: The PackageService is not public, it is used only here.

mugurrus:
:+1:

ioanpocol:
Disabled e2e test related to package and created a new PR:
https://github.com/superdesk/superdesk/pull/137

The package related e2e will be fixed by:
https://github.com/superdesk/superdesk/pull/101

PR #112, opened 2015-03-04
fix e2e setup

petrjasek:
a few more fixes for e2e. should fix logout test.
includes #107


ioanpocol:
:+1:

actionless:
it seems to be stailized for chrome@bamboo

actionless:
but i can't get why it fails on firefox on this line:
https://github.com/superdesk/superdesk/blob/fix-e2e-setup/client/spec/login_spec.js#L41

`btn` element is resolved only when and waitForAngular is being called from `element` again?
but that's very strange because it being called only from here (https://github.com/angular/protractor/blob/6c24a11c3549c3fa222a52309633e05d768c428f/lib/element.js#L136) and anyway it should happen only after waitForSuperdesk is resolved

btw, about waitForSuperdesk, https://github.com/superdesk/superdesk/blob/fix-e2e-setup/client/spec/helpers/utils.js#L57, it returns a promis which returns a promise. Are both promises being resolved here (https://github.com/superdesk/superdesk/blob/fix-e2e-setup/client/spec/login_spec.js#L39) or only first one (ie it returns unresolved promise which was returned by waitForAngular)?

petrjasek:
not sure what was the firefox issue, I've rewrote the test not to use waitForAngular after reloading the page, seems to work @actionless

mugurrus:
:+1:

PR #114, opened 2015-03-04
[SD-1682] - Display the expiry field in metadata tab for ingest items

akintolga:
Changes:
- Started using the 'expiry' field in ingested items
- Elastic filter in remove_expired_content has changed to use 'expiry' field (it should also be backward compatible)


petrjasek:
:+1:

ioanpocol:
:+1:

PR #120, opened 2015-03-06
(fix) SD-1721 and errors in console while creating a new user

sivakuna-aap:
[SD-1721] User with no save search permission gets improper error.
Modified edit-form.html and users.js:
 1. To display the "Default Desk" field only if the user is member of some desk.
 2. Fixed error thrown while creating a new user.


petrjasek:
:+1:

ioanpocol:
:+1:

PR #128, opened 2015-03-06
fix(e2e) - Fix firefox e2e tests

ioanpocol:


client/spec/content_spec.js
```diff
@@ -12,8 +12,8 @@ describe('Content', function() {
 
     beforeEach(function() {
         openUrl('/#/workspace/content')();
-        workspace.openPersonal();
-        expect(element.all(by.repeater('items._items')).count()).toBe(3);
+        workspace.switchToDesk('PERSONAL');
+        expect(element.all(by.repeater('items._items')).count()).toBe(6);
```
petrjasek: fyi you can use `.toBeGreaterThan(x)` so we don't need to update specs when there is a new item added

petrjasek:
can you push it also to `fix-e2e-setup` branch so we can test it on bamboo before merging to master?

mugurrus:
:+1:

petrjasek:
:+1:

PR #132, opened 2015-03-09
fix(ingest expiry) Email and Teletype parsers to include version created in all items

marwoodandrew:
The changes to the ingest expiry requires that all items are created with the versioncreated key in the item dictionary.


amagdas:
:+1:

ioanpocol:
:+1:

PR #133, opened 2015-03-09
fix(desks): Fixed displaying only a limited number of users when adding ...

eozan:
...users.

fix(groups): Fixed displaying only a limited number of users when adding users.

SD-1681


amagdas:
:+1:

ioanpocol:
:+1:

PR #134, opened 2015-03-09
(fix) SD-1723

sivakuna-aap:
[SD-1723] A combination of privileges gives unclear error.
Modified SuperdeskAPIError.__str__() in errors.py module to return error code and message, which is sometimes sent to client as part of response. This will enhance the user experience to see what exactly went wrong. To put it in simple words the error message is changed from "Error:403" to "Error: 403: Unauthorized to modify unique name".

### DIFF
--- a/server/superdesk/errors.py
+++ b/server/superdesk/errors.py
@@ -81,7 +81,7 @@ class SuperdeskApiError(SuperdeskError):
         return rv
 
     def __str__(self):
-        return repr(self.status_code)
+        return "{}: {}".format(repr(self.status_code), self.message)
 
     @classmethod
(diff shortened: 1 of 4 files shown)


server/superdesk/errors.py
```diff
@@ -81,7 +81,7 @@ def to_dict(self):
         return rv
 
     def __str__(self):
-        return repr(self.status_code)
+        return "{}: {}".format(repr(self.status_code), self.message)
```
amagdas: :100:

amagdas:
:+1:

ioanpocol:
:+1:

PR #137, opened 2015-03-09
 refactor(packages) - remove "packages" endpoint, use "archive" instead

ioanpocol:
On PR #108 disabled package related e2e test, until it will be fixed by https://github.com/superdesk/superdesk/pull/101


ioanpocol:
:+1:

amagdas:
Actually only one more test was failing, this is good as well.

petrjasek:
:+1:

PR #138, opened 2015-03-09
fix(travis): ignore firefox build results

petrjasek:
we don't want to spend too much time fixing firefox webdriver issues, so if build fails on firefox we still merge.
@superdesk/developers 
@marwoodandrew @akintolga @sivakuna-aap @mdhaman @sjunaid

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+sudo: false
+
 env:
     - TARGET=client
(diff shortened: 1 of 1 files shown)


ioanpocol:
:+1:

amagdas:
:+1:  :100:  :heart: :blue_heart: :yellow_heart:

PR #145, opened 2015-03-10
fix(highlights): Disable mark for highlights on Personal, show highlight...

ioanpocol:
...s with no desk

Disable mark for highlights on Personal
When a highlight has no desk, it will be showed for all desks
Change the filter from Personal in order to show only items not assigned
to a desk


amagdas:
:+1:

petrjasek:
:+1:

PR #147, opened 2015-03-10
(fix) Issues related lock

sivakuna-aap:
[SD-1651] Spiked item should not be editable from authoring
[SD-1718] Locked item being spiked
[SD-1719] Fetching a locked item creates another locked item
[SD-1724] User with no task management privilege is not able to send item to another desk
[SD-1728] It should not be possible to spike a locked item

### DIFF
--- a/client/app/scripts/superdesk-archive/module.js
+++ b/client/app/scripts/superdesk-archive/module.js
@@ -94,7 +94,10 @@ define([
                         });
                     }],
                     filters: [{action: 'list', type: 'archive'}],
-                    action: 'spike'
+                    action: 'spike',
+                    condition: function(item) {
+                        return item.lock_user === null || angular.isUndefined(item.lock_user);
+                    }
                 })
                 .activity('unspike', {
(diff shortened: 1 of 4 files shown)


ioanpocol:
:+1: 

The behave test re: macro USD to AUS exchange is failing. 
Looks like the exchange rate has changed dramatically :)
We will change the test to check that the values are different.

client/app/scripts/superdesk-archive/module.js
```diff
@@ -94,7 +94,10 @@ define([
                         });
                     }],
                     filters: [{action: 'list', type: 'archive'}],
-                    action: 'spike'
+                    action: 'spike',
+                    condition: function(item) {
+                        return item.lock_user === null || angular.isUndefined(item.lock_user);
```
petrjasek: fyi you can also use `item.lock_user != null`, it works for both `undefined` and `null`

petrjasek:
:+1:

petrjasek:
will fix the behave test

PR #148, opened 2015-03-10
fix(behave): use predefined USD/AUD rate

petrjasek:
there is another test using the real rate so there is no need
to test it there, where it's harder to make greaterThan asserts


ioanpocol:
:+1:

sivakuna-aap:
:+1:

PR #155, opened 2015-03-10
SD-1562 Add close button to edit package form

mugurrus:


client/spec/helpers/utils.js
```diff
@@ -1,15 +1,49 @@
...
 // construct url from uri and base url
 exports.constructUrl = function(base, uri) {
     return base.replace(/\/$/, '') + uri;
 };
 
-var LoginModal = require('./pages').login;
+var currBrowser = browser;
+
+function BrowserManager() {
+    this.setBrowser = function(userBrowser) {
```
amagdas: Is this ever used?

ioanpocol: It should be used on tests. A multi-browser test should be added.

petrjasek: yep with test it would be better

mugurrus:
it's just wip, sorry didn't flag it as such from the beginning

ioanpocol:
:+1:

petrjasek:
:+1:

PR #159, opened 2015-03-11
Routing

eozan:


petrjasek:
I think it was reviewed, so as long as it gets :+1: from qa we can merge this. Just pls fix this jshint issue.

PR #160, opened 2015-03-11
(fix) SD-1598

sivakuna-aap:
[SD-1598] Reopened because of the below scenario:

Go to any desk and click on the ingest tab.
*The page is populated with the ingest items
Now click on personal
*The page still gets poplulated with ingest items, instead it should switch to the content tab and show the personal items.


ioanpocol:
:+1:

amagdas:
:+1:

PR #163, opened 2015-03-11
Fix preferences

amagdas:
Refactor preferences to be stored inside user document instead of session, this helps with returning a valid ETAG.
- [x] Fix patching preferences
- [x] Fix session based preferences removal.


sivakuna-aap:
Looks like we are refactoring complete preferences. If possible can we map the URL to `users/#USER_ID#/preferences` instead of just `/preferences`?

amagdas:
It is still tied to the session, we get prefs like /preferences/#SESSION_ID#

server/apps/auth/db/commands.py
```diff
@@ -91,6 +91,7 @@ def run(self, username, password):
         }
         service = superdesk.get_resource_service('auth')
         id = str(service.post([credentials])[0])
+        print('Session ID:', id)
```
akintolga: This looks like a left-over bit of code...

mdhaman: print not needed.

amagdas: Actually is needed to know the session for which the token is generated, this is  a flask command.

mdhaman: :+1:

akintolga:
:+1:

PR #165, opened 2015-03-11
[SD-1683] - Transform ingested articles during routing

akintolga:
- [x] Register macros by desks
- [x] Made improvements in registering new macros and re-registering updated ones dynamically
- [x] List of registered macros in desk setting popup
- [x] Selection of a macro in fetch and publish actions in routing scheme set up
- [x] Execution of the macro while routing 
- [x] A new macro added to populate the abstract field
- [x] A behave/unit test to see macro in action in routing scheme


server/superdesk/macros.py
```diff
@@ -21,6 +22,13 @@ def load_macros():
     """
     try:
         module = app.config['MACROS_MODULE']
+        already_imported = sorted(sys.modules.keys())
+        for t in already_imported:
+            if t == module:
+                m = sys.modules[t]
+                imp.reload(m)
```
petrjasek: `imp` is deprecated, but you can use `importlib.reload` here

server/superdesk/macros.py
```diff
@@ -21,6 +22,13 @@ def load_macros():
     """
     try:
         module = app.config['MACROS_MODULE']
+        already_imported = sorted(sys.modules.keys())
+        for t in already_imported:
```
petrjasek: just wondering - is there a need for a loop? it should work with `if module in already_imported: ... ` no?

server/macros/__init__.py
```diff
@@ -4,5 +4,22 @@
 
 Use `superdesk.macros.macros.register` for registration.
 """
+import sys
+import imp
+macros = ['macros.currency', 'macros.abstract_populator']
 
-from . import currency  # noqa
+try:
```
petrjasek: I'm not sure this part is required. If you reload this module it will reload also modules within it no?

ioanpocol:
There is a small js error:
 app/scripts/superdesk-ingest/module.js

574 | }

^ Missing semicolon.

server/macros/__init__.py
akintolga: What I observed is that, we need to force reimporting the modules in a package, otherwise the package does not reflect the changes to the macros. It does add the new ones, because we would have a new import statement in __init__.py for the new macro. So (as it normally does) once a module is loaded it doesn't load again if it is not forced for reload.

server/superdesk/macros.py
akintolga: Yes, I'm getting rid of the loop.

server/superdesk/macros.py
akintolga: `importlib.reload` is only available from 3.4 forward. So it would fail in 3.3 which I use for development.

amagdas: Actually should we just go and support python 3.4 and  up? Python 3.5 will be here soon so we will support the latest 2 versions.
@petrjasek Any thoughts on this?

server/macros/currency.py
```diff
@@ -38,5 +47,6 @@ def convert(match):
     name='usd_to_aud',
     label='Convert USD to AUD',
     shortcut='c',
-    callback=usd_to_aud
+    callback=usd_to_aud,
+    desks=['SPORTS DESK', 'POLITICS']
```
ioanpocol: Whey here are hardcoded desk names?

ioanpocol:
:+1:

server/macros/__init__.py
petrjasek: :ok_hand:

server/macros/currency.py
petrjasek: I think it's more readable than desk ids

petrjasek:
:+1:

server/macros/currency.py
ioanpocol: Yes but why here and not in settings file?

petrjasek: because you want to assign this macro to given desks

ioanpocol: OK, probably on mongo is a better place then

petrjasek: yep at some point it could have some ui

server/apps/archive/archive_ingest.py
```diff
@@ -63,10 +64,15 @@ def create(self, docs, **kwargs):
...
                 ingest_doc = superdesk.get_resource_service('ingest').find_one(req=None, _id=doc.get('guid'))
+
                 if not ingest_doc:
                     msg = 'Fail to found ingest item with guid: %s' % doc.get('guid')
                     raise SuperdeskApiError.notFoundError(msg)
 
+                if doc.get('macro'):
+                    # there is a macro so transform it
+                    ingest_doc = get_resource_service('macros').execute_macro(ingest_doc, doc.get('macro'))
+
```
sivakuna-aap: Any specific reason for applying macro even before determining if the state transition is valid or not?

server/features/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'tolga'
```
sivakuna-aap: License Info pls.

akintolga: Done.

server/apps/archive/archive_ingest.py
akintolga: No, there wasn't. I have moved the transformation after the check of state transition.

amagdas:
:+1:

PR #166, opened 2015-03-11
fix(authoring): fix saving of packages

petrjasek:
keep original etag/last_updated when setting linked_in_packages

### DIFF
--- a/server/apps/archive/archive_composite.py
+++ b/server/apps/archive/archive_composite.py
@@ -118,7 +118,7 @@ class PackageService():
         item_id = assoc[ITEM_REF]
         item = get_resource_service(endpoint).find_one(req=None, _id=item_id)
         if not item:
-            message = 'Invalid item reference: ' + assoc['itemRef']
+            message = 'Invalid item reference: ' + assoc[ITEM_REF]
             logger.error(message)
             raise SuperdeskApiError.notFoundError(message=message)
         return item, item_id, endpoint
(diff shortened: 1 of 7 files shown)


server/apps/archive/archive_composite.py
```diff
@@ -134,8 +134,11 @@ def update_link(self, package_id, assoc, delete=False):
         if not delete:
             two_way_links.append({'package': package_id})
 
+        # do update without changing etag/last_modified
         updates = self.get_item_update_data(item, two_way_links, delete)
-        get_resource_service(endpoint).patch(item_id, updates)
+        updates['_etag'] = item['_etag']
```
amagdas: Why is this needed? It looks weird to have two different docs having the same etag...

petrjasek: it's same doc. but you modify the etag every time you save a package. and if you edit this somewhere else it would get etag conflict

petrjasek: i mean what it does is setting the etag to previous value to avoid autogenering

amagdas: I understood that, but maybe a better solution would be to notify the client that the document updated so it can refetch the new document with the latest changes, including _etag...
My worry is that having an entity in mongo with the same _etag although they are not having the same content, it will break concurrency control...

petrjasek: it's single item, you update that same item, change there linked_in_packages, but keep the etag.
so if user (it might be other user, item can be locked etc) will save item it will still work. we can't reload that item for user who is editing content because someone else might pick it into a package imo

amagdas: I agree, I think a better solution is to use the Eve support for this: https://github.com/nicolaiarocci/eve/commit/d7e9bd024d9c1a5faad0e31b737d119cf3ee1fc7

petrjasek: thing is this is after eve handling, updating a related item, not the package. it's making changes directly

ioanpocol: I think that a good approach is that the saved etag to have the same value as the one calculated on values, otherwise we will have some issues like those that we have now for user preferences that are failing on save because of this matter.

petrjasek: here the point is not to change it when we change values

ioanpocol:
:+1:

server/features/steps/steps.py
```diff
@@ -289,6 +289,21 @@ def step_impl_when_post_url(context, url):
...
 
 
+def get_response_etag(response):
+    return json.loads(response.get_data())['_etag']
+
+
+@when('we save etag')
+def step_when_we_save_etag(context):
+    context.etag = get_response_etag(context.response)
+    print(context.etag)
```
mugurrus: delete print

mugurrus:
:+1:

PR #167, opened 2015-03-11
refactor(ws): on server reconnect for every message

petrjasek:
instead of keeping single connection it will open one for every message, so behaving more like http.


ioanpocol:
:+1:

amagdas:
:+1:

PR #170, opened 2015-03-12
[SD-1329] Idle Channel notification

marwoodandrew:


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -217,6 +217,7 @@ define([
             link: function($scope) {
 
                 var DEFAULT_SCHEDULE = {minutes: 5, seconds: 0};
+                var DEFAULT_IDLE_TIME = {hours: 0, minutes: 0, seconds: 0};
```
amagdas: Do we need to support seconds? Having just minutes should be enough I think..

marwoodandrew: You're probably right, seconds are over kill!

amagdas:
:+1:

ioanpocol:
:+1:

PR #172, opened 2015-03-12
SD-1612 Define how to populate data

amagdas:
Phase 1: Add data initialization files and a script to bootstrap the mongo database.
In another pull request we will import editorial content and ingest sources.


ioanpocol:
:+1:

server/superdesk/data_initialization/import.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
petrjasek: would it be possible to make this manage command?

amagdas: It is just that it's a bit more work. This is on the plan just that I added this if we need a way to bootstrap a new env quickly.

amagdas:
@superdesk/developers @sivakuna-aap @marwoodandrew @mdhaman @akintolga

mdhaman:
:+1:

akintolga:
:+1:

sivakuna-aap:
:+1:

PR #176, opened 2015-03-12
fix(lightbox): Fix lightbox preview

nidzix:
SD-1737

### DIFF
--- a/client/app/scripts/superdesk-archive/directives.js
+++ b/client/app/scripts/superdesk-archive/directives.js
@@ -310,7 +310,11 @@ define([
         .directive('sdItemRendition', function() {
             return {
                 templateUrl: require.toUrl('./views/item-rendition.html'),
-                scope: {item: '=', rendition: '@'},
+                scope: {
+                    item: '=',
+                    rendition: '@',
+                    ratio: '=?'
+                },
                 link: function(scope, elem) {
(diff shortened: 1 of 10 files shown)


ioanpocol:
:+1:

client/app/scripts/superdesk-archive/directives.js
```diff
@@ -323,15 +328,63 @@ define([
...
                                 }
+                                _calcRatio();
                             };
 
                             img.onerror = function() {};
                             img.src = href;
                         }
                     });
+
+                    var stopRatioWatch = scope.$watch('ratio', function(val) {
```
petrjasek: seems complicated, if you have ratio set on scope from sdRatioCalc, why not use `ng-class="{portrait: ratio > 1.0}"` ?

nidzix: Because it's not using just scope.ratio, but also image ratio to get whether its portrait or not.

petrjasek: why it needs both then?

nidzix: You need to compare this, because thats the only way to properly style image with undefined ratio under the container with undefined ratio :)

amagdas:
:+1:

PR #177, opened 2015-03-12
fix(authoring): fix saving after autosave

petrjasek:
after autosave call it was extending the original item,
but later if you save it would compare again with original
and there were no changes. so now autosaving is only updating
$scope.item, also on load autosave is applied only to $scope.item
and not to $scope.origItem.


ioanpocol:
:+1:

nidzix:
:+1:

PR #180, opened 2015-03-12
fix(packages): Fix the search page on edit package

ioanpocol:


server/apps/archive/user_content.py
```diff
@@ -23,7 +23,7 @@ class UserContentResource(Resource):
     datasource = {
         'source': 'archive',
         'aggregations': aggregations,
-        'elastic_filter': {'not': {'exists': {'field': 'task.stage'}}}  # eve-elastic specific filter
+        'elastic_filter': {'not': {'exists': {'field': 'task.desk'}}}  # eve-elastic specific filter
```
amagdas: Why is this needed?

amagdas:
:+1:

server/apps/archive/user_content.py
ioanpocol: On personal we have to filter content created by current user and not assigned to any desk.

amagdas: And the previous filter didn't work?

ioanpocol: No

amagdas: :+1:

petrjasek:
:+1:

PR #182, opened 2015-03-13
fix(email ingest) handle variations of charsets in headers

marwoodandrew:
Have also re-factored the exception handling so that the exception is caught and reported in the parser and swallowed by the ingest provider.


amagdas:
:+1:

ioanpocol:
:+1:

PR #183, opened 2015-03-13
fix(idle channels) work around for the framework changing 0 values to ...

marwoodandrew:
datetimes

My zero hours and minutes were being changed to empty datetimes, I think this is a known "feature" :-1:


amagdas:
:+1:

petrjasek:
:+1:

ioanpocol:
:+1:

PR #184, opened 2015-03-13
fix(authoring): fix save when closing item

petrjasek:
it was missing orig param needed to compute the diff


ioanpocol:
:+1:

petrjasek:
@nidzix what do you think?

nidzix:
:+1:

PR #186, opened 2015-03-13
refactor(ingest): keep ingesting on single item error

petrjasek:
- instead of stopping on item failure, it will skip this very item
- removed catch all exception to get more info on sentry (temporary)

### DIFF
--- a/server/superdesk/io/reuters.py
+++ b/server/superdesk/io/reuters.py
@@ -65,10 +65,13 @@ class ReutersIngestService(IngestService):
         result_items = []
         while items:
             item = items.pop()
-            result_items.append(item)
-
             self.add_timestamps(item)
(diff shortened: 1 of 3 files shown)


amagdas:
:+1:

server/superdesk/io/reuters.py
```diff
@@ -128,7 +131,7 @@ def get_tree(self, endpoint, payload=None):
             raise IngestApiError(error, self.provider)
 
         if response.status_code == 404:
-            raise IngestApiError.apiNotFoundError(LookupError('Not found %s' % payload), self.provider)
+            LookupError('Not found %s' % payload)
```
petrjasek: looks like there should be `raise`

PR #195, opened 2015-03-16
fix(highlights): Revert button for export

nidzix:


ioanpocol:
:+1:

ioanpocol:
Added the Revert button for export to https://github.com/superdesk/superdesk/pull/205

PR #204, opened 2015-03-17
Revert "Revert "multiprocessing removed from backend server""

actionless:
Reverts superdesk/superdesk#202


ioanpocol:
I already done this:
https://github.com/superdesk/superdesk/pull/202

PR #209, opened 2015-03-17
SD-1566: Clicking on "View" from Content items' list and from preview of...

fritzSF:
... the item leads to different view


amagdas:
:+1:

PR #213, opened 2015-03-18
Handle missing dates on RSS

amagdas:
Looks like we get last_updated but not _created for some RSS feeds.
@superdesk/developers @sivakuna-aap @akintolga @marwoodandrew


akintolga:
:+1:

marwoodandrew:
:+1:

PR #217, opened 2015-03-18
fix(desks selection): When opening Content for a first time contents are...

ioanpocol:
... not match selected desk


mugurrus:
:+1:

nidzix:
:+1:

PR #220, opened 2015-03-18
continue e2e fixing for bamboo

actionless:
still some testcases are executed non-sequentially and sometimes that became noticeable when io is slow

TODO: use request-promise and make helpers with callbacks to return promises

### DIFF
--- a/client/spec/fetch_spec.js
+++ b/client/spec/fetch_spec.js
@@ -36,15 +36,19 @@ describe('Fetch', function() {
     it('can fetch from ingest with menu', function() {
         workspace.switchToDesk('SPORTS DESK').then(content.setListView);
         expect(element.all(by.repeater('items._items')).count()).toBe(2);
-        browser.get('/#/workspace/ingest');
-        workspace.switchToDesk('SPORTS DESK');
-        content.setListView();
-
-        content.actionOnItem('Fetch', 0);
-
-        browser.get('/#/workspace/content');
-        workspace.switchToDesk('SPORTS DESK');
-        content.setListView();
+        browser.get('/#/workspace/ingest').then(function() {
+            return workspace.switchToDesk('SPORTS DESK');
+        }).then(
+            content.setListView
+        ).then(function() {
+            return content.actionOnItem('Fetch', 0);
+        }).then(function() {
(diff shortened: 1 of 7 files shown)


amagdas:
:+1:

client/spec/fetch_spec.js
```diff
@@ -77,19 +81,26 @@ describe('Fetch', function() {
...
         expect(element.all(by.repeater('items._items')).count()).toBe(2);
 
-        browser.get('/#/workspace/ingest');
-        workspace.switchToDesk('SPORTS DESK');
-        content.setListView();
-        content.actionOnItem('Fetch', 0);
-        browser.get('/#/workspace/content');
-        workspace.switchToDesk('SPORTS DESK');
-        content.setListView();
+        browser.get('/#/workspace/ingest').then(function() {
```
ioanpocol: Based on the documentation from here: 
http://angular.github.io/protractor/#/control-flow
it is not needed to add "then" after every statement

actionless: between `expect` command all promises can be resolved unordered

ioanpocol: I think no. Based on documentation all statements are ordered in control flow queue, not only expect statements.

actionless: but from screenshots and logs it seems like that happens somehow

ioanpocol: We have to find the reason. I think that the code became too ugly by adding all those "then" statements

actionless: it was but the thing what no one of developers which were trying this issue before managed to find the reason of that "desynchronization", so if i won't manage to find the core problem of that than maybe we can consider moving to angular-independed e2e framework, like http://nightwatchjs.org/ or http://dalekjs.com/pages/getStarted.html

actionless:
can't reproduce the failure after that kernel tuning, let's see how it will be over the time

ioanpocol:
If after a few days it is working we should rolback "then" changes and check again for few days if it is working

actionless:
@ioanpocol master branch was still failing after write buffer was reduced, so it seems like they those fixes are work together

PR #221, opened 2015-03-19
[SD-1853] WENN/FTP Ingest errors

marwoodandrew:
Two problems, the subject codes parsed by the WENN parser where not in a list. Also the NITF parser can be used either by a File Ingestor or FTP, it returns a dictionary not a list.


sivakuna-aap:
:+1:

amagdas:
:+1:

PR #226, opened 2015-03-19
feat(desks) - Master desk view - Add Users tab

ioanpocol:


amagdas:
:+1:

ioanpocol:
There were some comments from Nikola that disappear after I performed an amend on commit.
Not fixed yet:
1. on user update refresh the view
2. refactor to use groupBy on finding roles available for a desk
3. when a role for a desk is hide, a small UI alignment issue appear

ioanpocol:
The refresh of view after user change will be added in a different PR

nidzix:
:+1:

PR #227, opened 2015-03-19
bamboo e2e fixes, pt.3 

actionless:


actionless:
so after some investigations i came to the fact what we had "desync" thingie because of 3 (at least) different reasons:
1) on unrelated failure reporter itself was trying to sync to page -- fixed in first PR
2) on slow io actions there executed non-sequentially (mb related to animations, i'm not totally sure) -- fixed in first and second PR
3) strange bug on navigation which reproduces on slow io -- hopefully fixed in this PR

amagdas:
:+1:

PR #232, opened 2015-03-20
[sd-1848]-Confirmation when deleting a rule

sjunaid:


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -436,8 +436,12 @@ define([
                     scope.schemes = result._items;
                 });
 
-                function confirm() {
-                    return modal.confirm(gettext('Are you sure you want to delete scheme?'));
+                function confirm(context) {
+                    if (context === 'scheme') {
+                        return modal.confirm(gettext('Are you sure you want to delete scheme?'));
```
amagdas: Maybe we should word it like: "Are you sure you want to delete this scheme?", there is "this" missing

ioanpocol:
:+1:

nidzix:
:+1:

PR #237, opened 2015-03-23
fix(ingest) - don't restart the ingest if an item fails.

amagdas:
SD-1869 - skip metadata load if we can't parse it
SD-1856 Investigate reuters image download triggering soft timeout


amagdas:
@superdesk/developers @sivakuna-aap @akintolga @marwoodandrew @mdhaman

marwoodandrew:
:+1:

akintolga:
:+1:

PR #239, opened 2015-03-23
feat - Ingest Dashboard

mdhaman:
- [x] Ingest Dashbaord (wip)
- [x] log messages 
- [x] Behave tests
- [x] e2e


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -77,6 +96,32 @@ define([
                 }
 
                 return this.fetched;
+            },
+            fetchDashboardProviders: function() {
+                var deferred = $q.defer();
```
petrjasek: looks like [forgotten promise](http://taoofcode.net/promise-anti-patterns/#theforgottenpromise). issue is that if the first api.query fails it won't reject it and it will leak some memory

petrjasek:
:+1:

ioanpocol:
:+1:

PR #240, opened 2015-03-23
[SD-1801] - Implementing filters for the search

akintolga:
- [x] Changed the search to have aggregations not affected by filter
- [x] Changed mark up to have check boxes
- [x] Created a tag service to represent the all types of filters as tags
- [x] UI: mark up and styling as well as chenges to the parameters pop-up (By Nikola)
- [x] Tag service tests
- [x] e2e tests


client/app/scripts/superdesk-search/search.js
```diff
@@ -203,6 +213,129 @@
...
+            'source': 1,
+            'day': 1,
+            'week': 1,
+            'month': 1,
+            'desk': 1,
+            'stage':1,
+            'state':1
+        };
+
+        var initSelectedParameters = function (parameters) {
```
petrjasek: hey, imo it would be better to define functions like `function initSelectedParameters(parameters) {` so in case of error (unlikely) you see its name in stack trace, also when profiling etc.

akintolga:
@petrjasek @ioanpocol @nidzix @mugurrus 

We discussed that search changes PR can be created against master today. What do you think? Should we keep it separate? We don't have an instance for it.

petrjasek:
master sounds good

ioanpocol:
OK
If you need a new instance we can create a new one based on a branch

PR #246, opened 2015-03-24
[SD-1655] Federated Search

marwoodandrew:
This PR is for discussion. 

It allows the AAP multimedia archive to be searched in from the global search. It has an additional check that if checked will search  the AAP multimedia archive only!

There is no attempt to merge the results from the elastic search and the multimedia archive as  a) this is hard, b) There is a suggestion that they should not be merged but displayed separately.

It sort of implements a DataLayer and returns a fabricated Eve elastic cursor.

Is this the correct way to architect this?

Should it be a separate end point?

Videos don't work, the item view doesn't work and there is no facility to fetch.

### DIFF
--- a/client/app/scripts/superdesk-search/views/item-search.html
+++ b/client/app/scripts/superdesk-search/views/item-search.html
@@ -2,7 +2,7 @@
     <fieldset ng-if="context">
         <label> 
             <input type="radio" value="local" ng-model="repo.search"> 
-            Local Repos
+            In Superdesk
         </label>
         <div ng-show="repo.search == 'local'">
             <div class="field">
--- a/client/app/scripts/superdesk-search/views/item-search.html
+++ b/client/app/scripts/superdesk-search/views/item-search.html
@@ -2,16 +2,24 @@
     <fieldset ng-if="context">
         <div class="field">
             <label for="ingest-collection">
-                {{:: 'Ingest' | translate}}
+                {{ 'Ingest' | translate}}
                 <input type="checkbox" id="ingest-collection" ng-model="repo.ingest" ng-checked="repo.ingest" tabindex="4">
             </label>
         </div>
         <div class="field">
             <label for="archive-collection">
-                {{:: 'Content' | translate}}
+                {{ 'Content' | translate}}
                 <input type="checkbox" id="archive-collection" ng-model="repo.archive" ng-checked="repo.archive" tabindex="5">
             </label>
         </div>
+    </fieldset>
+    <fieldset ng-if="context">
+        <div class="field">
+            <label for="archive-collection">
+            {{ 'AAP Multimedia' | translate}}
+            <input type="checkbox" id="aapmultimedia-collection" ng-model="repo.aapmultimedia" ng-checked="repo.aapmultimedia" tabindex="5">
+            </label>
+        </div>         
     </fieldset>
 	<fieldset>
         <div class="field">
--- a/client/app/scripts/superdesk-search/views/item-search.html
+++ b/client/app/scripts/superdesk-search/views/item-search.html
@@ -1,25 +1,31 @@
 <form>
     <fieldset ng-if="context">
-        <div class="field">
-            <label for="ingest-collection">
-                {{ 'Ingest' | translate}}
-                <input type="checkbox" id="ingest-collection" ng-model="repo.ingest" ng-checked="repo.ingest" tabindex="4">
-            </label>
-        </div>
-        <div class="field">
-            <label for="archive-collection">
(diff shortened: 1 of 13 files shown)


petrjasek:
hey - is the idea to have there read only access or also some editing? if it's read only i think we could define simpler api than datalayer

marwoodandrew:
Yep, read-only, I think the idea is that there should be something like the fetch button that will download the image from the image system and inject it into superdesk.

marwoodandrew:
After declaring it read only, I got jumped on in the stand-up, may need to update as well.

ioanpocol:
One option is to implemented a media storage like AmazonMediaStorage and index the images with elastic.

server/superdesk/aap_mm_datalayer.py
```diff
@@ -0,0 +1,92 @@
...
+                doc['mimetype'] = 'image/jpeg'
+
+            doc['slugline'] = doc.pop('Byline')
+        return ElasticCursor(docs=hits['docs'], hits={'hits': hits})
+
+    def find_all(self, resource, max_results=1000):
+        raise NotImplementedError
+
+    def find_one(self, resource, req, **lookup):
+        raise NotImplementedError
```
amagdas: Does the API support the implementation for `find_one`?

server/apps/search/__init__.py
```diff
@@ -22,7 +21,7 @@ class SearchService(superdesk.Service):
     It can search ingest/content/archive/spike at the same time.
     """
 
-    available_repos = ('ingest', 'archive')
+    available_repos = ('ingest', 'archive', 'aapmultimedia')
```
amagdas: This should be resolved after the application loads the modules, basically a module should register a search service in the form of a datalayer/or simpler API if we are going to do only search actions against it and SD datalayer should just dynamically use them.

server/superdesk/aap_mm_datalayer.py
marwoodandrew: Not yet, but it will, I think. It will be needed for the "fetch" type action.

amagdas:
@ioanpocol I don't think it's feasible to suggest to the clients to move all their data to AWS/S3 or Elastic, this PR actually adds the capability to query external resources and show the results in SD, the items will be indexed in SD only after fetching them.

ioanpocol:
@amagdas The proposal was not to move all data to amazon but to implement a media storage like AmazonMediaStorage (that is a python class).
From what I know the searching on current image storage system is very slow so it can be a solution to index it with Elastic.

marwoodandrew:
@ioanpocol Currently the [public image website](http://one.aap.com.au/#/) uses the same API as I'm using here, it is served by a SOLR index, it's reasonably quick.

ioanpocol:
OK, so if already indexed with SOLR it should be OK (elastic and solr uses the same search core that is lucene)

marwoodandrew:
Made some changes to the client so that images can be fetched into superdesk. One issue is that these items do not have an associated ingest_provider so do not display a source when in archive, the same applies to images that have been uploaded.

petrjasek:
imo it would be good if there is an api to list configured search providers so we don't have to modify client if we add a new one. but it would also require some generic fetch api that would take provider as param.

marwoodandrew:
One problem we have is that there is no ingest_provider associated with the stuff that gets fetched (also a problem for uploaded images) so once they are in the system you have know idea how they got there. 

It's been suggested that we should have an ingest provider entry for each search source, perhaps with a particular type (search?), we could list these when showing the available search providers in the search api. haven't got my head around it all yet!

marwoodandrew:
Have added a search type ingest provider

marwoodandrew:
It will now list the search type ingest providers in the global search in the UI, I should now be possible to add additional search sources, you just need to write an end point. That implements the search and the fetch services.

client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -1,15 +1,29 @@
...
-                <input type="checkbox" id="ingest-collection" ng-model="repo.ingest" ng-checked="repo.ingest" tabindex="4">
-            </label>
+        <label> 
+            <input type="radio" value="local" ng-model="repo.search"> 
+            In Superdesk
+        </label>
+        <div ng-show="repo.search == 'local'">
+            <div class="field">
+                <label for="ingest-collection">
+                    {{ 'Ingest' | translate}}
```
amagdas: `{{:: }}`

amagdas:
:+1:

akintolga:
:+1:

PR #247, opened 2015-03-24
fix(desks): fixed desks service to handle 200+ desks/stages.

eozan:
SD-1882

### DIFF
--- a/client/app/scripts/superdesk-desks/desks.js
+++ b/client/app/scripts/superdesk-desks/desks.js
@@ -275,6 +275,21 @@
 
                 var userDesks, userDesksPromise;
 
+                var _fetchCombined = function(endpoint, page, items) {
+                    page = page || 1;
+                    items = items || [];
+                    return api(endpoint)
+                    .query({max_results: 200, page: page})
+                    .then(function(result) {
+                        items = items.concat(result._items);
+                        if (result._links.next) {
+                            page++;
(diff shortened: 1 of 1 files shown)


amagdas:
:+1:

mugurrus:
:+1:

client/app/scripts/superdesk-desks/desks.js
```diff
@@ -275,6 +275,21 @@
 
                 var userDesks, userDesksPromise;
 
+                var _fetchCombined = function(endpoint, page, items) {
```
mugurrus: This retrieves the whole collection. Shouldn't there be a limit though?

sivakuna-aap:
:+1: 
Probably a better name for `_fetchCombined()`
It would be better if we can externalize the `max_results`. In future the value can be decreased or increased.

client/app/scripts/superdesk-desks/desks.js
eozan: Currently, desks service fetches all desks/stages/members and compiles lookup tables to use app-wide, and supposedly this should happen only once.

Ideally it needs to change, and fetch data only when needed and cache for future use. But until then, this needed to be done to be able to display stages.

mdhaman:
:+1:

ioanpocol:
:+1:

PR #248, opened 2015-03-24
Enable media deletion

amagdas:


server/superdesk/commands/clean_images.py
```diff
@@ -19,27 +19,27 @@ class CleanImages(superdesk.Command):
...
     Probably running db.repairDatabase() is needed in Mongo to shring the DB size.
     """
+
     def run(self):
         try:
             print('Starting image cleaning.')
             used_images = set()
 
-            archive_items = superdesk.get_resource_service('archive').get(None, {'type': 'picture'})
+            archive_items = superdesk.get_resource_service('archive').get_from_mongo(None, {'type': 'picture'})
```
mugurrus: Implementing an API tightly coupled to a specific database is not a good idea.

ioanpocol: Actually is to get from storage server and not from search server. Probably we just have to rename it.

ioanpocol:
:+1:

server/superdesk/commands/clean_images.py
amagdas: @mugurrus it is a good idea if you want to work specifically with the said database.

amagdas: @ioanpocol actually I think that the name is good as we are using specific mongo syntax when using it, I prefer to know what kind of API I can use.

ioanpocol: @amagdas  I don't see any specific/proprietary mongo syntax. Probably if we switch to other storage server we still can be used the same syntax. What is important here is that by this function you force to get data from  storage server and not from search server. 
But it is a minor issue so not needed to insist more on it.

PR #249, opened 2015-03-24
SD-1641 As an administrator, I would like to maintain a global dictionary

mugurrus:


petrjasek:
:+1:

server/apps/dictionaries/resource.py
```diff
@@ -0,0 +1,69 @@
...
+    'name': {
+        'type': 'string',
+        'unique': True,
+        'required': True
+    },
+    'language_id': {
+        'type': 'string',
+        'required': True
+    },
+    'language_name': {
```
ioanpocol: Why to add here language_name?

ioanpocol:
:+1:

server/apps/dictionaries/resource.py
mugurrus: Currently we don't have a languages collection containing language metadata; maybe it would have been a good idea to add it now...

PR #250, opened 2015-03-24
fix(dashboard): On desks combobox change refresh the page only once

ioanpocol:


client/app/scripts/superdesk-dashboard/module.js
```diff
@@ -38,26 +33,19 @@ define([
...
-                                    scope.select(scope.userDesks[0], false);
-                            });
-                        } else {
-                            scope.select(_.find(scope.userDesks, {_id: currentDeskId}), false);
-                        }
+                desks.initialize()
+                .then(function() {
+                    desks.fetchCurrentUserDesks().then(function (userDesks) {
+                    	scope.userDesks = userDesks._items;
+                    	if (!desks.activeDeskId && scope.userDesks) {
```
petrjasek: `scope.userDesks` will be `true` even for empty list, it would be better to check `length`

client/app/scripts/superdesk-dashboard/module.js
```diff
@@ -38,26 +33,19 @@ define([
...
+                    		scope.select(scope.userDesks[0], false);
+                    	} else {
+                    		if (desks.getCurrentDesk() !== scope.selected) {
```
petrjasek: here it might be more obvious to test for `!= null`, `scope.selected` is not yet defined anyhow right?

petrjasek:
:+1:

amagdas:
:+1:

PR #251, opened 2015-03-24
fix(workspace): prevent preferences update errors

petrjasek:
there were multiple calls to update preferences at the same time,
both using same etag, so the later had to fail. that was happening
on desk/stage change which are 2 settings. now those 2 should be
updated together, and also there is no update if the new value
matches the old one.

SD-1867

### DIFF
--- a/server/apps/preferences.py
+++ b/server/apps/preferences.py
@@ -159,7 +159,8 @@ class PreferencesService(BaseService):
 
             existing = existing_session_preferences.get(session_id, {})
             existing.update(session_prefs)
-            updates[_session_preferences_key] = existing
+            existing_session_preferences[session_id] = existing
+            updates[_session_preferences_key] = existing_session_preferences
 
     def update_user_prefs(self, updates, existing_user_preferences):
         user_prefs = updates.get(_user_preferences_key)
--- a/server/apps/preferences.py
+++ b/server/apps/preferences.py
@@ -154,13 +154,12 @@ class PreferencesService(BaseService):
     def update_session_prefs(self, updates, existing_session_preferences, session_id):
         session_prefs = updates.get(_session_preferences_key)
         if session_prefs is not None:
-            for k in ((k for k, v in session_prefs.items() if k not in superdesk.default_session_preferences)):
+            for k in (k for k, v in session_prefs.items() if k not in superdesk.default_session_preferences):
                 raise ValidationError('Invalid preference: %s' % k)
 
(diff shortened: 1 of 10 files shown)


petrjasek:
seems like there will be conflict with #250

server/apps/preferences.py
```diff
@@ -154,13 +154,12 @@ def on_update(self, updates, original):
     def update_session_prefs(self, updates, existing_session_preferences, session_id):
         session_prefs = updates.get(_session_preferences_key)
         if session_prefs is not None:
-            for k in ((k for k, v in session_prefs.items() if k not in superdesk.default_session_preferences)):
+            for k in (k for k, v in session_prefs.items() if k not in superdesk.default_session_preferences):
                 raise ValidationError('Invalid preference: %s' % k)
 
             existing = existing_session_preferences.get(session_id, {})
             existing.update(session_prefs)
-            existing_session_preferences[session_id] = existing
```
petrjasek: any idea why this was here? it prevented local updates to session preferences which got only modified using response but it was in form `{'session_preferences': {'session_id': {'key': 'va'}}}` which was not handled by prefs service
@amagdas @mugurrus @ioanpocol @akintolga @marwoodandrew @sivakuna-aap @mdhaman

petrjasek:
will rebase this once #250 is merged

server/apps/preferences.py
amagdas: This is intended behavior, the session preferences are stored exactly as you said which is ok, when fetching the preferences for a specific session we return only the prefs for that session.
I think this PR is OK except the backend changes.

amagdas:
:+1: to the frontend changes, I don't think that backend changes are needed.

server/apps/preferences.py
petrjasek: thing is that [in saveLocally](https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk/services/preferencesService.js#L33) it doesn't know about this structure so those settings changes are not available until you reload the page

petrjasek:
reverted that server part, but still think it's wrong. don't see why client would ever want to get preferences for other sessions. seems like it's leaking some implementation details @amagdas

amagdas:
You are totally right, if this is the case, my bad. I used `after_` events to return only the matched session but it looks like it doesn't work as supposed.
I will merge this PR and fix the server part today.
:+1:

PR #255, opened 2015-03-25
fix(highlights): Fix mark for highlights

ioanpocol:
There are cases when the document has the list of highlights but the
value is null


mugurrus:
:+1:

PR #261, opened 2015-03-25
fix(generate): ignore content for non-text items

petrjasek:
in case there is no `body_html` in item it will not output `None`


amagdas:
:+1:

ioanpocol:
:+1:

PR #269, opened 2015-03-26
fix(preferences): Fixed Update session preference settings behave test

ioanpocol:


petrjasek:
:+1:

mugurrus:
:+1:

ioanpocol:
After I updated requirements this is reproduced also local. So it is not needed any change.

PR #270, opened 2015-03-26
fix(desks): Fix 'Users' tab styling on master desk

nidzix:
SD-1929


client/app/scripts/superdesk-desks/views/desk_users.html
```diff
@@ -1,6 +1,7 @@
 <div class="desk-header">
-	<div class="desk-name pull-left">{{ desk.name }} 
+	<div class="desk-name pull-left">{{ desk.name }} {{desk._id}}
```
ioanpocol: Why is need to display desk id?

nidzix: oh, my fault, was checking smth and forgot to delete

ioanpocol:
:+1:

PR #272, opened 2015-03-26
chore(server): set superdesk/eve version via  commit

petrjasek:


actionless:
LGTM

server/requirements.txt
```diff
@@ -40,6 +40,6 @@ python3-ldap==0.9.5.3
 behave==1.2.5
 wooper==0.4.2
 
-git+git://github.com/superdesk/eve@sync-upstream#egg=Eve==0.6-dev
+git+git://github.com/superdesk/eve@e49cca#egg=Eve==0.6-dev
```
jey07ro: why is enought just e49cca when the full hash is e49cca9e48b4d33fe362305d8463314113642a6f?

server/requirements.txt
```diff
@@ -40,6 +40,6 @@ python3-ldap==0.9.5.3
 behave==1.2.5
 wooper==0.4.2
 
-git+git://github.com/superdesk/eve@sync-upstream#egg=Eve==0.6-dev
```
nistormihai: how is enough just e49cca when the full hash is e49cca9e48b4d33fe362305d8463314113642a6f?

plamut: GIT is smart enough to find a specific commit based only on the first few characters of the hash (provided that the combination is unique, of course).

ioanpocol: it's a kind of magic :)

nistormihai: smart is always better, we need to make smart things also :)
Thanks for help Peter.

plamut: YW :)

PR #273, opened 2015-03-26
fix(desks): check if user current desk exists

petrjasek:
we get user default desk via preferences, but this desk might not
exist anymore when using it, so it will check if that desk is there
when fetching user desks and if not it will set 1st desk as current.


ioanpocol:
:+1:

sivakuna-aap:
:+1:

PR #277, opened 2015-03-26
fix(desks): avoid updating preferences when fetching desks

petrjasek:
there is no need for this, and also it was a side effect doing updates
in fetch current user desks.


ioanpocol:
:+1:

PR #281, opened 2015-03-26
chore(master desks): Implement e2e tests for master desks

ioanpocol:


client/spec/helpers/pages.js
```diff
@@ -216,3 +217,30 @@ function IngestDashboard() {
         return dashboard.element(by.css('.ingested-count'));
     };
 }
+
+function MasterDesks() {
```
mugurrus: This should be defined in the desks module; we should put e2e tests in the corresponding modules.

mugurrus:
:+1:

petrjasek:
:+1:

client/spec/master_desk_spec.js
```diff
@@ -0,0 +1,116 @@
+
+'use strict';
+var openUrl = require('./helpers/utils').open;
+
+function MasterDesks() {
```
actionless: shouldn't it be moved to pages.js?

ioanpocol: Per Mugur feedback no (see above his comment).
I think we have to add on pages things that potentially are used in multiple tests, like workspace manipulation. Otherwise in time we will have a too big pages file like in the case of steps.

actionless: i see, i think he means pages/master_desk.js
@mugurrus right?

ioanpocol:
Looks like there are a lot of errors on travis because of github connectivity issues

PR #284, opened 2015-03-27
feat(aggregate): UI: add config modal for aggregate widget

nidzix:


client/app/scripts/superdesk/services/modalService.js
```diff
@@ -43,5 +43,45 @@ define(['angular'], function(angular) {
                 return delay.promise;
             };
 
+        }])
+        .directive('sdModal', ['$document', function($document) {
+
+            return {
+                replace: true,
```
petrjasek: replace is deprecated

client/app/scripts/superdesk/services/modalService.js
```diff
@@ -43,5 +43,45 @@ define(['angular'], function(angular) {
...
+                    '</div>'].join(''),
+                transclude: true,
+                link: function(scope, element, attrs) {
+
+                    var par = element.parent();
+
+                    attrs.$observe('open', function(model) {
+                        scope.isModalOpen = attrs.open;
+                        if (model === 'true') {
+                            element.appendTo($document.find('body'));
```
petrjasek: is this called more than once?

client/app/scripts/superdesk/services/modalService.js
plamut: Just wanted to say the same. :)

client/app/scripts/superdesk/services/modalService.js
```diff
@@ -43,5 +43,45 @@ define(['angular'], function(angular) {
...
+                    '<div class="modal-dialog" ng-if="isModalOpen === \'true\'"><div class="modal-content" ng-transclude></div></div>',
+                    '</div>'].join(''),
+                transclude: true,
+                link: function(scope, element, attrs) {
+
+                    var par = element.parent();
+
+                    attrs.$observe('open', function(model) {
+                        scope.isModalOpen = attrs.open;
+                        if (model === 'true') {
```
petrjasek: imo it could be boolean

nidzix: using `$parse`?

client/app/scripts/superdesk-desks/views/aggregate-widget.html
```diff
@@ -2,12 +2,25 @@
     <div class="search-container">
         <div class="search-box">
             <input type="text" placeholder="{{ 'Search'|translate }}" ng-model="query">
```
plamut: Fixed strings like this one should use the one-time binding syntax (::) to reduce the number of $watches.

client/app/scripts/superdesk/services/modalService.js
nidzix: it's called every time you run a modal, but it's also paired with close modal function, when you return content to original place

client/app/scripts/superdesk-desks/views/aggregate-widget.html
```diff
@@ -2,12 +2,25 @@
...
 
         <div class="filter-list">
-            
+
         </div>
-        
+
         <div class="content-list">
             <div class="desk"  ng-repeat="desk in desks._items">
                 <div class="stage" ng-repeat="stage in deskStages[desk._id]" ng-show="total" ng-init="open = true">
```
plamut: Does the list of desk stages get updated (recreated)? If so, it is advisable to add `track by` clause to ng-repeat to potentially save some unnecessary DOM operations (i.e. destroying and re-creating DOM elements corresponding to the same item in the list).

client/app/scripts/superdesk-desks/views/aggregate-widget.html
```diff
@@ -35,4 +48,35 @@
...
+    <div sd-modal data-open="{{modalActive}}" class="aggregate-widget-config">
+        <div class="modal-header">
+            <button class="close" ng-click="closeModal()"><i class="icon-close-small"></i></button>
+            <h3 translate>Configure content</h3>
+        </div>
+
+        <div class="modal-body">
+            <div class="content">
+                <div class="legend" translate>Select desks to view</div>
+                <div class="desk" ng-repeat="desk in desks._items">
```
plamut: Same here, adding `track by` might be beneficial (depending on how the `desks._items` list gets changed, of course).

client/app/scripts/superdesk/services/modalService.js
petrjasek: i mean you can do `if (model) {`

nidzix: but when it becomes `'false'` it will return true ?

petrjasek: yep, so why is this string and not just `true/false` ?

nidzix: because attrs.open returns model not string

petrjasek: i see

petrjasek:
:+1:

ioanpocol:
:+1:

client/app/scripts/superdesk/services/modalService.js
```diff
@@ -43,5 +43,53 @@ define(['angular'], function(angular) {
...
+            return {
+                template: [
+                    '<div class="modal fade hide">',
+                    '<div class="modal-dialog" ng-if="isModalOpen"><div class="modal-content" ng-transclude></div></div>',
+                    '</div>'].join(''),
+                transclude: true,
+                link: function(scope, element, attrs) {
+
+                    var content, _initialized = false;
+                    var trigger = attrs.open;
```
petrjasek: I think you can use isolated scope here, it's a bit confusing

nidzix: Is it worth isolating popup scope because of this?

plamut: If nothing else, it prevents accidental bugs caused by property shadowing due to scopes' prototypical inheritance.

nidzix: Yes, but this is reusable component, so if we have isolated scope, popup directive isn't extendable to use specific functions in ceratin cases?  *(e.g. when you need to pass save() function to it)*

plamut: Well, using isolated scopes is actually a _recommended_ for reusable components (for the exact reason I mentioned above).

By "extendable" you probably mean "configurable"? General-prupose components are easily configurable, it's done using the extra attributes on the DOM element a directive is used on. For example, specifying a _save()_ handler can be done like this:

<element data-onsave="&onSaveCallbackHandler()" ....>

petrjasek: this should be fine because it does transclude so this limitation will be only for this modals template.

PR #285, opened 2015-03-27
e2e test for multi content view

petrjasek:
based on #281 and includes some refactorings where helpers page objects are put into separate files

### DIFF
--- a/client/app/scripts/superdesk-archive/controllers/list.js
+++ b/client/app/scripts/superdesk-archive/controllers/list.js
@@ -5,13 +5,14 @@ define([
     'use strict';
 
     ArchiveListController.$inject = [
-        '$scope', '$injector', '$location',
+        '$scope', '$injector', '$location', '$timeout',
         'superdesk', 'session', 'api', 'desks', 'ContentCtrl', 'StagesCtrl'
     ];
-    function ArchiveListController($scope, $injector, $location, superdesk, session, api, desks, ContentCtrl, StagesCtrl) {
+    function ArchiveListController($scope, $injector, $location, $timeout, superdesk, session, api, desks, ContentCtrl, StagesCtrl) {
 
-        var resource;
-        var self = this;
+        var resource,
+            timeout,
+            self = this;
 
         $injector.invoke(BaseListController, this, {$scope: $scope});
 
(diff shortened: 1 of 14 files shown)


ioanpocol:
OK, I will close then #281

ioanpocol:
:+1:

client/app/scripts/superdesk-archive/controllers/list.js
```diff
@@ -5,13 +5,14 @@ define([
...
-    function ArchiveListController($scope, $injector, $location, superdesk, session, api, desks, ContentCtrl, StagesCtrl) {
+    function ArchiveListController($scope, $injector, $location, $timeout, superdesk, session, api, desks, ContentCtrl, StagesCtrl) {
 
-        var resource;
-        var self = this;
+        var resource,
+            timeout,
+            self = this;
 
         $injector.invoke(BaseListController, this, {$scope: $scope});
```
plamut: Let's use `self` instead, if we already have an alias for `this`. :)

client/app/scripts/superdesk-desks/views/desk_users.html
```diff
@@ -1,6 +1,6 @@
 <div class="desk-header">
-	<div class="desk-name pull-left">{{ desk.name }}
-		<span class="view" ng-if="isMemberOf(desk)" ng-click="openDeskView(desk)" title="{{ 'Open desk content' | translate}}"><i class="icon-external"></i></span>
+	<div class="desk-name pull-left">{{ desk.name }} 
+		<span class="view" ng-if="isMemberOf(desk)" ng-click="openDeskView(desk, 'task')" title="{{ 'Open desk content' | translate}}"><i class="icon-external"></i></span>
```
plamut: Missing one-time binding for the fixed translation string.

nidzix:
:+1:

client/app/scripts/superdesk-archive/controllers/list.js
petrjasek: I don't like python in js. better would be to avoid the alias imo

plamut: Well, sometimes it's needed because `this` can have different meanings depending on the context, thus storing a reference to the original `this` is fine (it doesn't have to be called "self", though).

But if the reference is not really needed in this controller, then it's best to simply remove the variable?

petrjasek: it's used later, but could be binded there

PR #286, opened 2015-03-27
fix(package): Fix package item styling

nidzix:
SD-1944


petrjasek:
:+1:

ioanpocol:
:+1:

plamut:
@nidzix Just a tip - it's better to work on new features in a topic branch than directly in your master branch. In case that a pull request gets rejected, you are left with a master branch which is inconsistent with the main repository's master, causing troubles later (conflicts etc.). ;)

PR #287, opened 2015-03-27
SD 1782: Provide a way to configure field name aliases for RSS ingest feeds.

plamut:
Provide a way to configure field name aliases for RSS ingest feeds.


server/superdesk/io/rss.py
```diff
@@ -104,23 +127,32 @@ def _fetch_data(self, config, provider):
...
         :return: created content item
         :rtype: dict
         """
-        item = dict()
-        item['guid'] = item['uri'] = data.get('guid')
-        item['type'] = 'text'
-        if data.get('updated_parsed'):
-            item['versioncreated'] = utcfromtimestamp(timegm(data.get('updated_parsed')))
-        if data.get('published_parsed') or data.get('updated_parsed'):
-            item['firstcreated'] = utcfromtimestamp(timegm(data.get('published_parsed', data.get('updated_parsed'))))
```
plamut: @amagdas This behavior has slightly changed, the `firstcreated` field is now only generated from the `published_parsed` field in data, but not from `updated_parsed` - is that OK now that we can define field name aliases?

plamut:
*Note to QA:* there is a bug with saving the provider configuration object to database, it is not possible to remove an already existing field alias even if correct data is sent in the PATCH request. Please keep this in mind when testing the feature.

(the bug itself has been reported in a separate issue)

amagdas:
Unfortunately I have to :-1:  this PR, the downsides  of doing this are:
- much harder debugging for Ingest errors as the fields now are going to be provider dependent
- easy to introduce "Fat fingers" mistakes while configuring the ingest, after that we will have a bunch of data to fix
- the ingest channel configuration becomes tedious work as in order to get a working channel a bunch of work needs to be done

I would much rather provide some sane defaults and then allow for this aliasing if it's really needed.

ioanpocol:
@amagdas "I would much rather provide some sane defaults and then allow for this aliasing if it's really needed. "
Yes, this was the idea and in this way it has been implemented.

ioanpocol:
:+1:

plamut:
@amagdas As for the complexitiy, it's true, aliases make debuggingmore difficult. OTOH is there any other sensible way to deal with _generic_ data sources? After some discussions we decided that aliases would be a good way to tackle this field naming problem.

As @ioanpocol has said, we do use default field names for data sources and aliases only need to be defined when a data source names a particular field differently (and certainly not for all the fields, that would indeed be a hassle).

petrjasek:
:+1:

amagdas:
Then please take care when merging this that all the current ingest sources work as expected on master and data instances. Probably we need to stop all RSS ingest feeds before merging, then do the needed configuration and then start the feed again.

ioanpocol:
@amagdas Please confirm that your vote has been changed or not

amagdas:
:+1: But please take care of properly handling the already configured rss providers on the running instances, if not celery will spit a lot of errors again.

plamut:
Who has a good overveiw of all current RSS providers configured across all the instances we use?
(I don't, unfortunately)

plamut:
Anyone? :)

amagdas:
I will take care of merging this after I fix SD-1945

plamut:
OK, sounds like a plan. Thanks!

mugurrus:
:+1:

amagdas:
Closing this, created #309 as a replacement.

PR #289, opened 2015-03-30
feat(authoring): Refactoring send-to for fetch-as

eozan:


amagdas:
Maybe we should fix the tests for this branch, delaying to do so will get harder and harder to fix them.

eozan:
I'm having problems with running e2e test server, and can't fix them now. It would be good if we have instructions on setting the environment properly.

PR #291, opened 2015-03-30
fix(ingest) - always generated an _id, guid is going to point to the ID in the provider system

amagdas:
SD-1946 Fetching operation fails on recently ingested items with Post 404 not found code

This is a big change in how the system works, summary of changes:
- all the items will have an `_id` generated that is going to be different from `guid`, this is because `guid` is not always suited to be used inside URL's
- we won't be able to use `_id` and `guid` interchangeably which is a good thing IMO, we shouldn't rely on this at all, I've updated the obvious situations, hopefully I fixed all of them

@petrjasek @superdesk/developers @sivakuna-aap @akintolga @marwoodandrew @mdhaman

### DIFF
--- a/server/apps/duplication/archive_fetch.py
+++ b/server/apps/duplication/archive_fetch.py
@@ -41,7 +41,7 @@ class FetchResource(Resource):
         }
     }
 
-    url = 'ingest/<{0}:guid>/fetch'.format(item_url)
+    url = 'ingest/<{0}:id>/fetch'.format(item_url)
 
     resource_methods = ['POST']
     item_methods = []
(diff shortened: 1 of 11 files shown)


server/apps/duplication/archive_fetch.py
```diff
@@ -52,23 +52,24 @@ class FetchResource(Resource):
...
 
-    def __fetch(self, docs, guid=None, **kwargs):
-        guid_of_fetched_items = []
+    def __fetch(self, docs, id=None, **kwargs):
+        id_of_fetched_items = []
 
         for doc in docs:
-            guid_of_item_to_be_fetched = doc['guid'] if guid is None else guid
+            id_of_item_to_be_fetched = doc.get('_id') if id is None else id
+            print('Id: ', id_of_item_to_be_fetched)
```
mugurrus: remove print

server/apps/duplication/archive_fetch.py
```diff
@@ -52,23 +52,24 @@ class FetchResource(Resource):
...
+        return self.__fetch(docs, id=request.view_args.get('id'), **kwargs)
 
-    def __fetch(self, docs, guid=None, **kwargs):
-        guid_of_fetched_items = []
+    def __fetch(self, docs, id=None, **kwargs):
+        id_of_fetched_items = []
 
         for doc in docs:
-            guid_of_item_to_be_fetched = doc['guid'] if guid is None else guid
+            id_of_item_to_be_fetched = doc.get('_id') if id is None else id
```
mugurrus: When id is given, why do you do the ingest for all docs in the list?

sivakuna-aap: Agreed, but that's the syntax of service handler in `BaseService` for `POST` request

petrjasek:
:+1: guid as _id didn't work anyhow. but I'm surprised there are no client changes..

amagdas:
@petrjasek Will retest the client, most probably we use the self href from the entity.

server/apps/duplication/archive_fetch.py
amagdas: @mugurrus I think the idea is to be able to apply multiple fetch actions on the same item, most probably to apply different macros for routing/publishing.

amagdas:
@petrjasek Retested, the client side looks ok.

amagdas:
@sivakuna-aap @akintolga @marwoodandrew @mdhaman 
Added the fix for fetching the content from celery, review please.

mdhaman:
:+1:

PR #293, opened 2015-03-30
auth.js: use regex instead of comparing location path and route path

vied12:
When `$location.path` contains variables (id or parameters),
requiresLogin() can't return the `auth` status.
So here I am using the regex provided by `$route` to test if a route match the `$location.path`.
Then we retrieve the `auth` status or `false` if not found.


petrjasek:
:+1:

ioanpocol:
:+1:

amagdas:
:+1:

PR #298, opened 2015-03-31
Refactoring and standardizing modals

nidzix:


ioanpocol:
:+1:

client/app/docs/module.js
```diff
@@ -32,6 +43,23 @@
         return {
             restrict: 'C',
             link: function postLink(scope, element, attrs) {
+
+                //remove leading whitespaces
```
petrjasek: you can use `String.trimLeft` for this

nidzix: This is all about proper indentation of code, where every nested level should be indented 4chars more than previous one. So the idea is to pick indentation from first line, and trim left all other lines with that indentation.

petrjasek: I see, my bad

petrjasek:
:+1:

amagdas:
:+1:

PR #302, opened 2015-04-01
fix(docs): Fix images path in docs

nidzix:


ioanpocol:
:+1:

client/tasks/options/copy.js
```diff
@@ -13,17 +13,28 @@ module.exports = {
             ]
         }]
     },
-    docs: {
+    docsViews: {
```
petrjasek: no need to split, `files` is list

nidzix: :+1:

petrjasek:
:+1:

PR #303, opened 2015-04-01
chore(highlights): Fix e2e tests

ioanpocol:


client/app/scripts/superdesk-archive/views/list.html
```diff
@@ -2,7 +2,7 @@
 	<div class="button-stack right-stack">   
 	    <div class="navbtn dropdown strict highlights-dropdown" 
             dropdown sd-package-highlights-dropdown 
-            ng-if="type === 'archive' && selected.desk">
```
ioanpocol: @petrjasek  At line 21 is still used selected.desk and it looks like is not set correctly when current desk selection is changed. I was not able to fix it.

petrjasek:
seems like there are some js formatting issues

client/app/scripts/superdesk-archive/views/list.html
petrjasek: I think you can fix it if you add `$scope.selected = active;` to [controller](https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-archive/controllers/list.js#L107-L114)

client/spec/helpers/highlights.js
```diff
@@ -0,0 +1,117 @@
+
+'use strict';
+
+var openUrl = require('./helpers/utils').open;
```
petrjasek: here it should be `require('./utils').open` it's in helpers folder

petrjasek:
:+1:

PR #304, opened 2015-04-01
fix(docs): Include vendor libs

nidzix:
Looks like some angular bootstrap functionality is broken because of missing libs

Appended another fix for images


amagdas:
:+1:

ioanpocol:
:+1:

PR #305, opened 2015-04-01
feat(highlights): When an item is already marked for a highlight, disabl...

ioanpocol:
...e highlight on makr dropdown


amagdas:
:+1:

mugurrus:
:+1:

PR #308, opened 2015-04-02
[SD-1871] - Macros couldn't get loaded dynamically in celery

akintolga:


sivakuna-aap:
:+1:

server/settings.py
```diff
@@ -111,7 +111,11 @@ def env(variable, fallback_value=None):
     'spike:gc': {
         'task': 'apps.archive.content_purge',
         'schedule': crontab(minute=30)
-    }
+    },
+    'macros:print': {
+        'task': 'apps.macros.print_macros',
+        'schedule': timedelta(seconds=30),
```
amagdas: I think the delta should be bigger, ~10 minutes, as this is not going to change really often.

amagdas:
:+1:

ioanpocol:
:+1:

mugurrus:
:+1:

server/settings.py
petrjasek: this part should be removed, it was there only for testing

PR #309, opened 2015-04-02
Provide a way to configure field name aliases for RSS ingest feeds.

amagdas:
Replaces #287


mugurrus:
:+1:

server/superdesk/utils.py
```diff
@@ -75,6 +75,17 @@ def is_hashed(input_str):
     return input_str.startswith('$2a$')
 
 
+def merge_dicts(dict_args):
```
plamut: It would be slightly more pythonic if this is changed to `*dict_args` - this way the calling code won't have to explicitly pack the dictionariy arguments into a list or a tuple.

amagdas: In this case, we actually have  a list, we can improve it when the time comes.

plamut: That's true for our specific case, I was just thinking about a general use case. It's a minor remark anyway.

plamut:
All in all it seems good. When those two redundant forEach loops are removed, we can merge this.

ioanpocol:
:+1:

PR #310, opened 2015-04-02
data initialization - add test and create some default indexes

amagdas:


petrjasek:
:+1:
would be nice to define indexed in the schema, like elastic mapping

amagdas:
I agree, this would be a nice feature to add to Eve.

mugurrus:
:+1:

PR #313, opened 2015-04-02
feat(highlights): Add the option to see the marked highlights on authori...

ioanpocol:
...ng screen

Add the option to see the marked highlights on authoring screen
When an item is marked for highlights his _update, _etag will not
change.   The highlights is not used on _etag calculation anymore.


petrjasek:
:+1: but could use some behave test for this new endpoint

nidzix:
:+1:

PR #314, opened 2015-04-02
fix(client): make indentation consistent in js

petrjasek:
it's set to 4 spaces now
@superdesk/developers @sivakuna-aap @marwoodandrew @mdhaman @akintolga 
make sure your editors are set to expand tabs..


plamut:
:-1: The linter is already doing it's job well, causing Travis to fail due to inconsistent indentation in some of the files. :)

petrjasek:
ups :smile: can we get rid of it? :smiley:
i will have to update my npm

plamut:
What do you mean? Is it because these not-a-multiple-of-four indentation errors are caused by some automated task?
If not, it seems trivial to fix the problematic files.

petrjasek:
no I was using it when fixing locally, but didn't get those errors. after doing npm update I can see it

plamut:
It's good now. :+1:

PR #327, opened 2015-04-06
SD-1974: Create icon for disable user action, user logged status and ses...

fritzSF:
- Added info if the user is logged or not on the User management screen and the Users view on Master desk.


nidzix:
:+1:

ioanpocol:
:+1:

PR #328, opened 2015-04-07
(fix) Hot Fixes

sivakuna-aap:
1. Email Validation fails for valid email id's, for example: user@aap.com.au
2. When there is an error reported by API while saving user the client says the user is saved although it's not saved
3. Javascript Error reported on console for "field_aliases" while adding/modifying Ingest Provider


amagdas:
:+1: 
We should add some tests for this so it won't regress.

PR #335, opened 2015-04-07
(fix) Hot Fixes

sivakuna-aap:
1. Session_Preferences were not deleted from users collection on logout
2. Clear distinction between logged-in and not logged-in users

### DIFF
--- a/client/app/scripts/superdesk-users/users.js
+++ b/client/app/scripts/superdesk-users/users.js
@@ -7,8 +7,10 @@
     UsersService.$inject = ['api', '$q', 'notify'];
     function UsersService(api, $q, notify) {
 
-        this.usernamePattern = /^[A-Za-z0-9_.'-]+$/;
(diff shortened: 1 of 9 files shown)


amagdas:
Behave test please.

sivakuna-aap:
On the way :smile:

mugurrus:
:+1:

client/app/scripts/superdesk-users/users.js
```diff
@@ -1148,6 +1148,10 @@
                         }
                     });
 
+                    scope.isLoggedIn = function(user) {
```
mugurrus: This could be moved to the users service since it's used at least 2 times and may be used multiple times in the future.

sivakuna-aap: Refactored `isLoggedIn()` to `UsersService`

amagdas:
:+1:  :heart: :blue_heart:

petrjasek:
:+1:

PR #339, opened 2015-04-08
Show picture and description in authoring screen

amagdas:
SD-1914

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -21,7 +21,8 @@
         dateline: '',
         language: null,
         unique_name: '',
-        keywords: []
+        keywords: [],
+        description: null
     };
 
     /**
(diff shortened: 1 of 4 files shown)


marwoodandrew:
:+1:

eozan:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -662,6 +663,10 @@
                     superdesk.intent('view', 'content');
                 };
 
+                $scope.isMediaType = function isMediaType(item) {
+                    return item && _.contains(['audio', 'video', 'picture'], item.type);
```
petrjasek: this won't change during item editing, so it could be like ```$scope.isMediaType = _.contains(.., item.type);```
there is no need to do this iteration in every digest. or those could be just one time watches in template

amagdas: You are right, will fix.

PR #346, opened 2015-04-08
fix(reuters): only fetch items that are not expired

petrjasek:
instead of fetching items up to 7 days old only fetch items
that will pass expiry check from reuters api.


ioanpocol:
:+1:

amagdas:
:+1:

PR #347, opened 2015-04-08
feat(authoring): implement multi content view config

petrjasek:
use can select which desks/stages should be there


nidzix:
:+1:

amagdas:
:+1:

PR #348, opened 2015-04-08
fix(fetch): use item._id as family_id in ingest

petrjasek:
content family_id equals to ingest item _id when its fetched


amagdas:
:+1:

ioanpocol:
:+1:

PR #349, opened 2015-04-09
[SD-1328] - Automatically close ingest channel after critical error

akintolga:
List of changes for this task:
- A new end point "ingest_errors" to get the list of errors per provider
- Change in provider end point to save the selected critcal errors for that provider
- UI change to have a "critical errors" tab to select the errors that supposed to close the channel
- provider_service catches all SuperdeskIngestErrors and closes the channel if the error code is in the list of critical errors

### DIFF
--- a/server/features/ingest_errors.feature
+++ b/server/features/ingest_errors.feature
@@ -1,4 +1,4 @@
-Feature: Macros
+Feature: Ingest Errors
 
     @auth
     Scenario: Get list of all errors
--- a/server/features/ingest_errors.feature
+++ b/server/features/ingest_errors.feature
@@ -1,0 +1,25 @@
+Feature: Macros
+
+    @auth
+    Scenario: Get list of all errors
+        When we get "/ingest_errors"
+        Then we get list with 1 items
+            """
+            {"_items":[{"all_errors":{"1001":"Message could not be parsed","1002":"Ingest file could not be parsed","1004":"NewsML1 input could not be processed","1006":"NITF input could not be processed","1008":"ZCZC input could not be processed","1009":"IPTC7901 input could not be processed","2004":"Ingest error","4000":"Unknown API ingest error","4001":"API ingest connection has timed out.","4002":"API ingest has too many redirects","4003":"API ingest has request error","4004":"API ingest Unicode Encode Error","4005":"API ingest xml parse error","4006":"API service not found(404) error","4007":"API authorization error","5000":"FTP ingest error","5001":"FTP parser could not be found","6000":"Email authentication failure","6002":"Email ingest error"}}]}
+            """
+
+    @auth
+    Scenario: Get list of all errors for reuters
+        When we get "/ingest_errors?source_type=reuters"
+        Then we get list with 1 items
+            """
+            {"_items":[{"all_errors":{"1001":"Message could not be parsed","1002":"Ingest file could not be parsed"},"source_errors":{"4000":"Unknown API ingest error","4001":"API ingest connection has timed out.","4002":"API ingest has too many redirects","4003":"API ingest has request error","4004":"API ingest Unicode Encode Error","4005":"API ingest xml parse error"}}]}
+            """
+
+    @auth
+    Scenario: Get list of all errors for email
+        When we get "/ingest_errors?source_type=email"
+        Then we get list with 1 items
+            """
+            {"_items":[{"all_errors":{"1001":"Message could not be parsed","1002":"Ingest file could not be parsed"},"source_errors":{"6000":"Email authentication failure","6002":"Email ingest error"}}]}
+            """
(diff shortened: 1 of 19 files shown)


server/superdesk/io/ingest_errors.py
```diff
@@ -0,0 +1,59 @@
...
+    def get(self, req, lookup):
+        """Return all ingest errors."""
+        try:
+            source_type = getattr(req, 'args', {}).get('source_type')
+            if source_type:
+                return ListCursor([self.get_errors_by_source_type(source_type)])
+            else:
+                return ListCursor([self.get_all_errors()])
+        except Exception as ex:
+            print(ex)
```
sivakuna-aap: Print can go away

server/features/ingest_provider.feature
```diff
@@ -121,6 +121,29 @@ Feature: Ingest Provider
...
+        When we patch "/ingest_providers/#ingest_providers._id#"
+        """
+        {"critical_errors":{"6000":true, "6001":true}}
+        """
+        Then we get updated response
+        """
+        {"critical_errors":{"6000":true, "6001":true}}
+        """
+
+
```
sivakuna-aap: Can we have behave scenario for closing the channel on critical error?

server/features/ingest_errors.feature
```diff
@@ -0,0 +1,25 @@
+Feature: Macros
```
sivakuna-aap: Feature name should be `Ingest Errors` instead of `Macros`

server/superdesk/io/ingest_errors.py
```diff
@@ -0,0 +1,59 @@
...
+from superdesk.utils import ListCursor
+
+logger = logging.getLogger(__name__)
+
+
+class IngestErrorsService(superdesk.Service):
+    def get(self, req, lookup):
+        """Return all ingest errors."""
+        try:
+            source_type = getattr(req, 'args', {}).get('source_type')
```
sivakuna-aap: This case is better handled by eve itself with the help of `additional_lookup`.

server/superdesk/io/ingest_errors.py
```diff
@@ -0,0 +1,59 @@
...
+    def _get_all_errors(self):
+        all_errors = {}
+        for k, v in provider_errors.items():
+            all_errors.update(v)
+
+        return all_errors
+
+
+class IngestErrorsResource(superdesk.Resource):
+    resource_methods = ['GET', 'POST']
```
sivakuna-aap: Since we don't create ingest_errors in Mongo, we can remove `POST` and `privileges`

amagdas:
It needs https://github.com/superdesk/superdesk/pull/350 to be merged first so the build is fixed.

server/features/ingest_provider.feature
akintolga: I created a unit test for closing the channel when a critical error is thrown. It is on update_ingest_tests.py. I think it covers the closing functionality. Let me know if you see anything missing.

server/superdesk/errors.py
```diff
@@ -142,18 +142,23 @@ def __init__(self, code, exception, provider=None):
...
         self.provider_name = provider.get('name', 'Unknown provider') if provider else 'Unknown provider'
 
-        if provider.get('notifications', {}).get('on_error', True):
-            exception_msg = str(exception)[-200:]
-            update_notifiers('error',
-                             'Error [%s] on ingest provider {{name}}: %s' % (code, exception_msg),
-                             resource='ingest_providers' if provider else None,
-                             name=self.provider_name,
-                             provider_id=provider.get('_id', ''))
+        if exception:
```
amagdas: `if not exception: return` would be a bit more cleaner IMO or just use one `if exception:`

server/superdesk/io/aap.py
```diff
@@ -24,6 +24,9 @@
 
 logger = logging.getLogger(__name__)
 PROVIDER = 'aap'
+errors = [ParserError.nitfParserError(None, None).get_error_description(),
+          ProviderError.ingestError(None, None).get_error_description(),
+          ParserError.parseFileError(None, None, None, None).get_error_description()]
```
amagdas: This looks ugly, maybe we can rid of them or use named parameters so we know to what we are passing `None`

server/superdesk/errors.py
akintolga: Absolutely.

server/features/ingest_provider.feature
amagdas: We still need to test the triggering of `close_provider` so I think a new test is needed.

amagdas:
:+1:

server/features/ingest_provider.feature
akintolga: That's true. I added another test in update_ingest_tests: test_ingest_provider_calls_close_provider
This checks if the close_provider function is called when an error is thrown from the update.

ioanpocol:
:+1:

PR #352, opened 2015-04-09
feat(user sessions): Selete user sessions

ioanpocol:
e2e tests will be in the next commit/PR.

### DIFF
--- a/client/app/scripts/superdesk-users/users.js
+++ b/client/app/scripts/superdesk-users/users.js
@@ -80,6 +80,18 @@
             return user && _.size(user.session_preferences) > 0;
         };
 
+        /**
+         * Close user's sessions
+         */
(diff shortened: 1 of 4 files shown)


sivakuna-aap:
:+1:

amagdas:
I think it will be safer if we do this by disabling the user.
When disabling the user, we remove all the session and the user is no longer able to login/create a new session until the admin finishes with the role changes.

sivakuna-aap:
If the intention behind this PR is the one mentioned by @amagdas then I do agree with what he says

amagdas:
So until we clarify this: :-1:

ioanpocol:
Per Andrew Lasky the disable user cover the close session functionality so I close this PR.

PR #355, opened 2015-04-09
[SD-1089]- There seems to be an unnecessary internal information on imag...

sjunaid:
...e metadata displayed in preview


petrjasek:
yep guid is not internal, it's external (global) id

PR #357, opened 2015-04-09
feat(highlights): add unmark to highlights tooltip

petrjasek:
it has some workarounds for tooltip, once there is new angular bootstrap it can use `tooltip-template` directive


amagdas:
:+1:

PR #358, opened 2015-04-09
Fix loading of all users, only the first 200 users were loaded

amagdas:
SD-2017  Add user to desk does not take

### DIFF
--- a/client/app/scripts/superdesk-desks/desks.js
+++ b/client/app/scripts/superdesk-desks/desks.js
@@ -331,6 +331,7 @@
                 var _fetchAll = function(endpoint, page, items) {
                     page = page || 1;
                     items = items || [];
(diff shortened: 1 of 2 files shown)


sivakuna-aap:
Is it a good idea to load all users into browser cache? Can't we use auto complete instead of loading at once?

amagdas:
We have  a bunch of places where we depends on this behaviour, I agree that we should refactor this.
I will create a technical task to propose a fix.

sivakuna-aap:
:+1:

amagdas:
@superdesk/developers @akintolga @marwoodandrew

eozan:
I agree that it's not a good idea to do this as a permanent solution.
:+1:

PR #361, opened 2015-04-10
feat(highlights): Add the option to set the groups for highlights

ioanpocol:


client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -79,6 +79,21 @@
                 highlight: highlight._id
             };
 
+            if (highlight.groups && highlight.groups.length > 0) {
+                var groups = [{
```
amagdas: This part looks to be a copy of the packagesService, any reason why we can't reuse that?

client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -323,6 +350,51 @@
...
+            if (group !== '') {
+                $scope.editingGroup = {'name': group, 'id': $scope.configEdit.groups.indexOf(group) + 1};
+            } else {
+                $scope.editingGroup = {'name': '', 'id': 0};
+            }
+        };
+
+        $scope.removeGroup = function(group) {
+            $scope.configEdit.groups.splice($scope.configEdit.groups.indexOf(group), 1);
+            console.log('after remove: ', $scope.configEdit.groups);
```
amagdas: probably not needed.

client/app/scripts/superdesk-packaging/packaging.js
```diff
@@ -217,7 +217,17 @@
 
         $scope.$watch('item', function(item) {
             $scope.highlight = item.highlight;
-            init = true;
+            if ($scope.highligh) {
```
amagdas: `$scope.highlight` instead of `$scope.highligh`

amagdas:
:+1:

petrjasek:
:+1:

PR #365, opened 2015-04-13
Subscribers end point

akintolga:
[SD-2048] - Subscribers end point created
[SD-2047] - Output channels are changed to have list of subscribers

### DIFF
--- a/server/features/subscribers.feature
+++ b/server/features/subscribers.feature
@@ -73,7 +73,7 @@ Feature: Subscribers
       "name":"Channel 3"
     }
     """
-    When we post to "/output_channels"
+    When we post to "/output_channels" with success
     """
     [
       {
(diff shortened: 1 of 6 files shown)


marwoodandrew:
:+1:

server/features/subscribers.feature
```diff
@@ -0,0 +1,90 @@
...
+      "name":"Channel 3"
+    }
+    """
+    Then we get latest
+    """
+    {
+      "name":"Channel 3"
+    }
+    """
+    When we post to "/output_channels"
```
amagdas: We should check that the post works as expected. We can do it by calling:
`When we post to "/output_channels" with success`

mdhaman:
:+1:  :100: 
I had to comment on this ....

amagdas:
:+1: 
@mdhaman stay away from the computer :smile:

PR #372, opened 2015-04-14
feat(dictionary): let user modify existing dictionary

petrjasek:
user can add/remove words from dictionary.

### DIFF
--- a/client/app/scripts/superdesk-dictionaries/dictionaries.js
+++ b/client/app/scripts/superdesk-dictionaries/dictionaries.js
@@ -24,7 +24,7 @@
             api.find('dictionaries', dictionary._id).then(success, error);
         };
 
-        this.upload = function create(dictionary, data, file, success, error) {
+        this.upload = function create(dictionary, data, file, success, error, progress) {
             var hasId = _.has(dictionary, '_id') && dictionary._id !== null;
             var method = hasId ? 'PATCH' : 'POST';
             var headers = hasId ? {'If-Match': dictionary._etag} : {};
@@ -53,7 +53,7 @@
                     data: sendData,
                     file: file,
                     headers: headers
-                }).then(success, error);
+                }).then(success, error, progress);
             }, error);
         };
 
@@ -124,6 +124,7 @@
             $scope.closeDictionary();
             $scope.fetchDictionaries();
             notify.success(gettext('Dictionary saved succesfully'));
+            $scope.progress = null;
             return result;
         }
 
@@ -134,6 +135,7 @@
             } else {
                 notify.error(gettext('Error. Dictionary not saved.'));
             }
+            $scope.progress = null;
         }
 
         // listen for the file selected event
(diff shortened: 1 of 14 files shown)


amagdas:
:+1:

client/app/scripts/superdesk-dictionaries/dictionaries.js
```diff
@@ -102,74 +114,93 @@
...
         };
 
         $scope.cancel = function() {
-            reset();
+            $scope.closeDictionary();
+        };
+
+        $scope.addWord = function(word) {
+            $scope.dictionary.content = $scope.dictionary.content || [];
+            $scope.dictionary.content.push(word);
```
mugurrus: It's faster to use the addword endpoint and to send only the new word instead of sending the whole dictionary content, which can have 100k words.

client/app/scripts/superdesk-dictionaries/dictionaries.js
```diff
@@ -102,74 +114,93 @@
...
-                    notify.success(gettext('Word added succesfully: ') + updates.word);
-                }, onError);
+        $scope.removeWord = function(wordModel, search) {
+            $scope.dictionary.content = $scope.dictionary.content || [];
+            $scope.dictionary.content.splice(wordModel.index, 1);
+            $scope.filterWords(search);
         };
 
-        reset();
+        function sortWords(a, b) {
```
mugurrus: There's no need to sort words, they are sorted on the backend anyway.

server/apps/dictionaries/service.py
```diff
@@ -21,34 +21,51 @@
 logger = logging.getLogger(__name__)
 
 
+def words(text):
```
mugurrus: This only works for English

server/apps/dictionaries/service.py
```diff
@@ -21,34 +21,51 @@
...
+
+def merge(doc, words):
+    doc['content'] = list(set(doc.get('content', []) + words))
+
+
+def read_from_file(doc):
+    content = doc.pop(DICTIONARY_FILE)
+    if 'text/' not in content.mimetype:
+        raise SuperdeskApiError.badRequestError('A text dictionary file is required')
+    return words(read(content))
```
mugurrus: This would return an invalid dictionary for non English languages. Even for English it wouldn't work for words like don't, can't etc.

mugurrus: Dictionaries have a word on each line and it's safer to read one line, strip it and consider it a word

server/apps/dictionaries/service.py
petrjasek: right by design

client/app/scripts/superdesk-dictionaries/dictionaries.js
petrjasek: yep but it handles adding and removing words together. and you can add many words and save it only once. so having extra endpoint seems like premature optimisation atm

server/apps/dictionaries/service.py
petrjasek: it's hard to say what would work for english when there is no spellchecker.. so it's not important atm

ioanpocol:
:+1:

amagdas:
:+1:

PR #373, opened 2015-04-14
fix(users): Prevent etag change on user login and user logout

ioanpocol:


amagdas:
What's the actual problem we are trying to solve?

petrjasek:
@ioanpocol it can use some test imo
@amagdas etag is changed on logout so if someone wants to edit a user, asks him to logout, and clicks save - it will fail because of etag change.

amagdas:
Ok, thx for the info.

server/apps/preferences.py
```diff
@@ -155,7 +148,6 @@ def on_fetched_item(self, doc):
         self.enhance_document_with_user_privileges(doc)
 
     def on_update(self, updates, original):
-        # Beware, dragons ahead
```
amagdas: This was such a nice comment, it's a pity to remove it :)

amagdas:
:+1:

mugurrus:
:+1:

PR #375, opened 2015-04-15
Fix for ingest race condition

amagdas:


petrjasek:
:+1: 
there is also [python](https://github.com/SPSCommerce/redlock-py) [locks]( http://redis.io/topics/distlock) which might be simple to use and should deal with redis failures. but not sure about python 3+

amagdas:
Yup, I saw it just that I'm reluctant to introduce locking, now I just detect the race condition and exit.

mugurrus:
:+1:

PR #376, opened 2015-04-15
feat(publish): Added subscribers to output channels.

eozan:
SD-2049

### DIFF
--- a/client/app/scripts/superdesk-publish/publish.js
+++ b/client/app/scripts/superdesk-publish/publish.js
@@ -176,7 +176,7 @@
                 $scope.remove = function(subscriber) {
                     modal.confirm(gettext('Are you sure you want to delete subscriber?'))
                     .then(function() {
-                        api.subscribers.remove(subscriber);
+                        return api.subscribers.remove(subscriber);
                     })
                     .then(function(result) {
                         _.remove($scope.subscribers, subscriber);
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-publish/publish.js
```diff
@@ -176,17 +176,18 @@
...
-                            _.remove($scope.subscribers, subscriber);
-                        }, function(response) {
-                            if (angular.isDefined(response.data._message)) {
-                                notify.error(gettext('Error: ' + response.data._message));
-                            } else {
-                                notify.error(gettext('There is an error. Subscriber cannot be deleted.'));
-                            }
-                        }).then(fetchSubscribers);
-                    });
+                        api.subscribers.remove(subscriber);
```
plamut: Missing `return`? We must return the `api`'s promise so that the next `.then()` in the chain only gets executed once the API call has completed.

client/app/scripts/superdesk-publish/publish.js
```diff
@@ -244,18 +278,20 @@
...
-                            if (angular.isDefined(response.data._message)) {
-                                notify.error(gettext('Error: ' + response.data._message));
-                            } else {
-                                notify.error(gettext('There is an error. Output Channel cannot be deleted.'));
-                            }
-                        }).then(fetchOutputChannels);
-                    });
+                    modal.confirm(gettext('Are you sure you want to delete output channel?'))
+                    .then(function() {
+                        api.output_channels.remove(outputChannel);
```
plamut: Missing `return` here, too?
(for the same reason as in the previous comment)

amagdas:
:+1:

ioanpocol:
:+1:

PR #381, opened 2015-04-15
Add tests skeleton for the public API.

plamut:
The base for all public API-related tests.

### DIFF
--- a/server/publicapi/__init__.py
+++ b/server/publicapi/__init__.py
@@ -22,7 +22,9 @@ from eve import Eve
 from publicapi.datalayer import ApiDataLayer
 
 
+_script_directory = os.path.dirname(os.path.realpath(__file__))
+
 app = Eve(
-    settings=os.path.join(__name__, 'settings.py'),
+    settings=os.path.join(_script_directory, 'settings.py'),
     data=ApiDataLayer,
 )
--- a/server/publicapi/errors.py
+++ b/server/publicapi/errors.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014, 2015 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+"""
+A module that contains exception types for the Superdesk public API.
+"""
+
+
+from superdesk.errors import SuperdeskError
+
+
+class PublicApiError(SuperdeskError):
+    """Base class for all Superdesk public API errors."""
+
+    _codes = {
+        10000: "Unknown API error.",
+    }
+    """A mapping of error codes to error messages."""
+
+    def __init__(self, error_code=10000):
+        super().__init__(error_code)
+
+
+class UnknownParameterError(PublicApiError):
+    """Used when request contains an unknown parameter."""
+
+    PublicApiError._codes[10001] = "Unknown parameter."
+
+    def __init__(self):
+        super().__init__(10001)
+
+
+class BadParameterValueError(PublicApiError):
+    """Used when request contains a parameter with an invalid value."""
+
+    PublicApiError._codes[10002] = "Bad parameter value."
+
+    def __init__(self):
+        super().__init__(10002)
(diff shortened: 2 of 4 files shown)


server/publicapi/tests/datalayer_test.py
```diff
@@ -0,0 +1,33 @@
...
+    """Base class for `publicapi.datalayer.ApiDataLayer` class tests."""
+
+    def setUp(self):
+        try:
+            from publicapi.datalayer import ApiDataLayer
+        except ImportError:
+            # a missing class should result in a test failure, not in an error
+            self.fail("Could not import class under test (ApiDataLayer).")
+        else:
+            # XXX: do we need to pass the app object here? or is None fine?
```
amagdas: Why is this needed?

ioanpocol:
In order to have setup the test infrastructure I think that it will help if you will add an endpoint even if the implementation is very simple. After you have an endpoint you can write a simple unittest and a simple behave test. 
By using settings we should be able to have the new endpoint in the current app instance or to have a dedicated instance for it.

mugurrus:
I agree with Nelu. We don't need this template, unit and behave tests will be in the endpoint directory.

server/publicapi/tests/datalayer_test.py
plamut: `eve.io.base.DataLayer` (the grandparent of the class under test) expects an (optional) Flask application instance as a [parameter to `__init__`](https://github.com/nicolaiarocci/eve/blob/develop/eve/io/base.py#L94). It's a pattern to extend Flask.

I am not yet sure whether this is needed for the test purposes or not, or is passing `None` (or not passing anything) just fine, hence the comment.

plamut:
What do you mean by adding an endpoint? The endpoint(s) already exist and the tests for the exposed application object will then test that the expected endpoints indeed exist, behave as expected, etc. After all, the whole point of Eve is to add new endpoints quickly declaratively (i.e. via configuration).

Testing that configuration correctly results in the actual endpoints in the API application is out of the scope IMO, since this is already covered by tests of the Eve framework itself.

Please clarify, maybe I don't understand you correctly. Thanks!

ioanpocol:
Here is the summary of our meeting:
1. There will be a new application with its own database
2. The code will be still integrated on server folder
3. The definitions of the new endpoints will be using the same infrastructure/base classes as endpoint from main(server) application
4. On content publishing a celery task will move the data on the new database
5. As much as possible the generation of documents on different delivery formats will be performed at insert/update time and not when the document is requested by an external client
6. First it will be implemented a single/simple endpoint and then validate that the test infrastructure is working

amagdas:
Hmm, a bunch of this is already being done on the destinations branch, we should look into reusing what's in there already.
@akintolga @petrjasek @plamut please talk about this...

plamut:
Just a thought - how likely it is that the API will be required to support write operations in the future? Fetching items from the new API-specific database is easy, but writing to it and synchronizing with the "main" database... well, you get the point.

Probably not that likely, since the data enters the system via ingest, right? Thoughts?

ioanpocol:
@amagdas Looks like the new application will be like a subscriber. But a discussion will be useful to clarify how the things will work together.

mugurrus:
@amagdas Looks like destinations implements functionality for defining destinations where to publish content, subscribers, routing etc. Which would certainly be useful for populating the published items database. But the public content API would still be needed.

plamut:
To recap the quick discussion on HipChat:

* We will use the existing functionality in `destinations` branch for populating the API database. However, a NINJS formatter will have to be implemented, since that is the format API will use.
* On top of the API database, API endpoints will be implemented as Superdesk Services containing business logic (e.g. request parameters validation).

akintolga:
Just a quick heads-up for the implementation in destinations branch:
- This branch will have the code structure to push the data to our subscribers
- Once an article is published all deliverables (formatted article per transmission channel) will be queued in publish_queue
- Celery tasks will then poll the queue to perform the transmission.

So in the light of the above summary I agree with @plamut s last comments. I think for the API we can use the same publish queue where the articles will be converted to the format of the output channel: in this case it will be API.

plamut:
Thanks for the heads-up, @akintolga, it seems that we are on the right track then. Also thanks to @amagdas for mentioning the `destinations` branch, it will save us from reinventing some of the stuff. :+1:

PR #388, opened 2015-04-16
[SD-2035] - POC Infinite scroll (front end) in a separate branch (initia...

sjunaid:
...l version)

### DIFF
--- a/client/app/scripts/superdesk-archive/controllers/list.js
+++ b/client/app/scripts/superdesk-archive/controllers/list.js
@@ -34,6 +34,7 @@ define([
             $scope.stages.select(null);
         };
         $scope.fetching = false;
+        $scope.page = 1;
 
         $scope.stageSelect = function(stage) {
             if ($scope.spike) {
@@ -54,7 +55,6 @@ define([
             .then(function(items) {
                 $scope.loading = false;
                 $scope.items = items;
-                $scope.page = 1;
             }, function() {
                 $scope.loading = false;
             });
@@ -70,10 +70,12 @@ define([
                 $scope.loading = true;
                 resource.query(criteria)
                 .then(function(items) {
-                    $scope.loading = false;
                     $scope.items._items = $scope.items._items.concat(items._items);
                     $scope.fetching = false;
                 }, function() {
+                    //
+                })
+                ['finally'](function() {
                     $scope.loading = false;
                 });
             }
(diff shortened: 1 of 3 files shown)


client/app/scripts/superdesk-archive/controllers/list.js
```diff
@@ -44,31 +45,49 @@ define([
...
+                query.from = ($scope.page - 1) * query.size;
+                var criteria = {source: query};
+                $scope.loading = true;
+                resource.query(criteria)
+                .then(function(items) {
+                    $scope.loading = false;
+                    $scope.items._items = $scope.items._items.concat(items._items);
+                    $scope.fetching = false;
+                }, function() {
+                    $scope.loading = false;
```
plamut: `$scope.loading = false` repeats both in success and error handler - better to have that in `.finally()` to avoid repetition.

client/app/scripts/superdesk-search/views/search-results.html
```diff
@@ -16,24 +16,20 @@
...
-                        <div sd-media-box></div>
-                    </li>
-                </ul>
+        <div class="content" sd-scrolled data-next-action="fetchNext" data-previous-action="fetchPrevious">
+            <div sd-updown data-items="items._items" data-select="preview(item)">
+                <div class="loading-indicator" ng-show="loading" translate>loading</div>
+                <div sd-search-tags></div>
+                <div>
+                    <ul class="{{view}}-view list-view">
+                        <li ng-repeat="item in items._items" class="list-item-view" ng-class="{active: item === selected.preview}">
```
plamut: Should we add `track by item._id` here? In what way does the `items._items` list change, can it gets re-populated with potentially same elements?

eozan:
:+1:

client/app/scripts/superdesk-archive/controllers/list.js
sjunaid: done

client/app/scripts/superdesk-search/views/search-results.html
sjunaid: Hi, its just a POC at the moment to see the scrolling loading the data. we don't know the specifics of how the list will be populated at this stage, so more specific scenarios we will handle later on.

akintolga:
:+1:

client/app/scripts/superdesk-search/views/search-results.html
plamut: Oh, OK then. I just mentioned it for the reference if things become slow at some point. `Track by` could help then.

sjunaid: Thanks Peter, much appreciated.

PR #391, opened 2015-04-16
feat spell check

petrjasek:
early stage of spell checking ui. feedback welcome

### DIFF
--- a/client/app/scripts/superdesk/editor/editor.js
+++ b/client/app/scripts/superdesk/editor/editor.js
@@ -11,9 +11,40 @@
 
 'use strict';
 
-var TEXT_TYPE = 3;
+var TEXT_TYPE = 3,
+    MARKER_CLASS = 'sdmark';
+
+function click(target) {
+    target.dispatchEvent(new MouseEvent('click'));
+}
 
 function EditorService() {
+
+    this.storeSelection = function storeSelection(node) {
+        var selection = document.getSelection();
+        if (!selection.anchorNode) {
+            return;
+        }
+
+        var span = document.createElement('span'),
+            range = document.createRange();
+        range.setStart(selection.anchorNode, selection.anchorOffset);
+        span.classList.add(MARKER_CLASS);
+        range.insertNode(span);
+    };
+
+    this.resetSelection = function resetSelection(node, data) {
+        var marks = node.getElementsByClassName(MARKER_CLASS),
+            selection = document.getSelection(),
+            range = document.createRange();
+        selection.removeAllRanges();
+        while (marks.length) {
+            var mark = marks.item(0);
+            range.setStartBefore(mark);
+            selection.addRange(range);
+            mark.parentNode.removeChild(mark);
+        }
+    };
 
     this.clear = function() {
         this.elem = null;
@@ -203,7 +234,7 @@ angular.module('superdesk.editor', [])
 
     .service('editor', EditorService)
 
-    .directive('sdTextEditor', ['editor', function (editor) {
+    .directive('sdTextEditor', ['editor', 'spellcheck', '$timeout', function (editor, spellcheck, $timeout) {
 
         var config = {
             buttons: ['bold', 'italic', 'underline', 'quote', 'anchor'],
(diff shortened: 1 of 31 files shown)


client/app/scripts/superdesk/editor/editor.js
```diff
@@ -13,6 +13,10 @@
 
 var TEXT_TYPE = 3;
 
+function escapeRegExp(string) {
+    return string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
plamut: Is this something you found somewhere or did you write it by yourself? Escaping can be tricky, thus it is probably best to use an existing well-tested expression.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -199,11 +203,125 @@ function FindReplaceCommand(rootNode) {
     }
 }
 
+function SpellcheckService($q) {
+
+    var selection = document.getSelection();
+
+    this.check = function check(text) {
+        var words = _.uniq(_.filter(text.split(/[\W0-9]+/)));
```
plamut: This will not always work correctly, because `\W` does not recognize non-ascii letters. In the AES project we solved this problem by using [XRegExp](http://xregexp.com/). We used the following expression to search for words in a text:

```javascript
var unicodeWords = new XRegExp('(\\p{Letter}|\\d)+', 'g');
```
Hope this helps!

petrjasek: it's not searching for words, but for whitespace/numbers

client/app/scripts/superdesk/editor/editor.js
petrjasek: it's from mdn

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -199,11 +203,125 @@ function FindReplaceCommand(rootNode) {
...
+        return errors;
+    };
+
+    function cursorReset(node) {
+        selection.collapse(node, 0);
+    }
+
+    function cursorMove(times) {
+        for (var i = 0; i < times; i++) {
+            selection.modify('move', 'right', 'character');
```
plamut: `selection.modify()` is [not standard](https://developer.mozilla.org/en-US/docs/Web/API/Selection/modify). We must be careful and test this across different browsers.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -271,28 +389,51 @@ angular.module('superdesk.editor', [])
             templateUrl: 'scripts/superdesk/editor/views/editor.html',
             link: function(scope, elem, attrs, ngModel) {
 
-                var editorElem;
+                var editorElem, spelling = false, timeout;
```
plamut: This can be confusing for Python developers, it reads as if you were doing this:

```javascript
var editorElem = false;
var spelling = timeout;
```

I suggest having each variable declaration in its own line for better clarity.

client/app/scripts/superdesk/editor/editor.js
plamut: OK, good then.

client/app/scripts/superdesk/editor/editor.js
plamut: But `/[\W0-9]+/` will also match letters like `é`, `ö`, `ñ`, splitting the word `Schön` into `Sch` and `n` - surely this is not desired or ...?

petrjasek: right. I will update it

plamut: I also suggest renaming the "words" variable, since it doesn't actually contain a list of words but a list of non-word sequences. On the first glance I thought you were extracting words from the text. ;)

petrjasek: it is list of words, it does split the text on non-word characters

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -199,11 +205,171 @@ function FindReplaceCommand(rootNode) {
...
+                return dict;
+            });
+        }
+
+        return dictPromise;
+    }
+
+    this.check = function check(text) {
+        return getDict().then(function() {
+            var words = _.uniq(_.filter(text.split(/[\W0-9]+/)));
```
plamut: Adding a marker comment (splitting non-ascii words) - the last one got outdated due to the obsolete diff.

client/app/scripts/superdesk/editor/editor.js
amagdas: But this is javascript, right, we should be able to read javascript if doing code reviews...

plamut: In a vacuum this is true, of course, but it might be different when the head is tired and you are switching contexts often.
(a bit of a guess, I admit, but it did happen to me when I was glancing over the code in a hurry and parsed the statement in the way described :) ... but in any case, it's an obsolete diff anyway)

server/apps/dictionaries/service.py
```diff
@@ -10,34 +10,57 @@
...
+
+def train(features):
+    model = collections.defaultdict(lambda: 1)
+    for f in features:
+        model[f] += 1
+    return model
+
+
 def words(text):
     return re.findall('[a-z]+', text.lower())
```
plamut: Better to use `\w` instead of `[a-z]` IMO, because we deal with unicode text and we want to recognize the words containing non-ascii letters, too.

petrjasek: this would keep numbers there. probably it would be better to split on `[\W\d]`

plamut: It depends on how do we define words, but yeah, splitting on `[\W\d]` seems like a good choice. After all, natural languages don't contain words like `variable123`, thus we can omit the digits.

Another option would be to simply match all letter-only sequences, e.g. `[^\d\W_]`?

amagdas:
:+1:

plamut:
:+1:

amagdas:
:+1:

PR #394, opened 2015-04-17
public content API endpoint and behave infrastructure

mugurrus:


ioanpocol:
:+1: 
Some changes are needed on travis to create a new build job for runing the behave tests for content api endpoints.

plamut:
Thanks for mentioning. It's not urgent though, we are still creating a proof of concept.

plamut:
Re: `publicapi/tests.py` - the `tests` module can quickly become lengthy, I suggest we convert to `tests` package (i.e. put it into its own directory).

Thoughts?

mugurrus:
The file publicapi/tests.py is for behave tests setup, not for unit tests. I could rename it if you think it's confusing.

plamut:
That would be a good idea, yes, so that it's immediately clear what belongs to unit and what to behavioral tests, even for newcomers. Also, try to find a name that doesn't get discovered by nosetests by default (avoids the need for `nose` extra configuration).
Thanks!

petrjasek:
I think behave would load file `environment.py` for setup

plamut:
If this PR is fine in general, I suggest we merge it despite not being ideal yet, because we need this proof-of-concept code to continue working on top of it.

PR #398, opened 2015-04-19
Improvements on publishing

akintolga:
Changes for this PR:

- Archive items are formatted per output channel. The same formatted item will be used for many transmissions because output channels usually contains many destinations (transmission). So created a separate end point for formatted items.
- Selector codes are now resolved as per output channel and added to the queue. 
- Created a simple publish queue page to monitor the status of the transmissions.
- Published items will be moved to designated publish stage on the same desk if there's any
- Created an Email transmission

### DIFF
--- a/client/app/scripts/superdesk-publish/publish.js
+++ b/client/app/scripts/superdesk-publish/publish.js
@@ -129,6 +129,44 @@
         };
     }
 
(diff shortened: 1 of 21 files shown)


amagdas:
:+1:

petrjasek:
:+1:

PR #407, opened 2015-04-21
Show only online users

ioanpocol:


client/app/scripts/superdesk-users/users.js
```diff
@@ -252,13 +253,26 @@
                 };
 
             if (params.q) {
-                criteria.where = JSON.stringify({
+                var query = {
```
akintolga: There is a UserListService in the same file on line 90. Maybe we should use that service here. I have another PR (https://github.com/superdesk/superdesk/pull/404) where I supposed to do the same actually. What do you think?

amagdas:
:+1:

client/app/scripts/superdesk-users/users.js
ioanpocol: I will merge your PR and then I will make the refactoring.

petrjasek:
:+1:

akintolga:
:+1:

ioanpocol:
@akintolga I didn't refactored yet the users to use the cached service. I think that the effort to do this is not so small as I thought initially, so I will add a new PR for it.

akintolga:
Fair enough. I am also more than happy to look at the re-factoring that bit in the next few days as I supposed to do that in the first place.

ioanpocol:
OK

PR #410, opened 2015-04-21
fix(test): Added missing publish module to test config.

eozan:


amagdas:
The tests are still failing.

PR #411, opened 2015-04-21
fix(publish) - A better error message is used

akintolga:


marwoodandrew:
:+1:

amagdas:
:+1:

server/apps/publish/archive_publish.py
```diff
@@ -79,8 +79,10 @@ def update(self, id, updates, original):
             user = get_user()
             push_notification('item:publish', item=str(item.get('_id')), user=str(user))
             return item
-        except KeyError:
-            raise SuperdeskApiError.badRequestError(message="A non-existent content id is requested to publish")
+        except KeyError as e:
+            raise SuperdeskApiError.badRequestError(
+                message="Key is missing on article to be published: {}"
+                .format(str(e)))
```
plamut: Probably good to suppress the exception context here ( raise ... **`from None`**) to avoid too verbose stack traces? After all we are only interested in the last error (which is already descriptive enough) and not in the original KeyError.

akintolga: I observed this key problem in destinations branch few times, so some of the articles are missing a key but I don't know which one. That's why I am interested in the original KeyError as I need to see what key is creating problem. 

Please let me know if there's a better way of doing that. Cheers.

plamut: If you need to know the key that caused the problem, including that piece of information in the error message is perfectly fine, even desired.

I was referring more to the actual tracebacks that get printed to the console - I was thinking that if we already include all the relevant info in the raised error, then we probably don't really need the whole traceback of the original KeyError anymore and can suppress it?

It might not be problematic here, but I've seen cases when a single error flooded the console with _extremely_ long tracebacks of all the exception in the chain, even though printing only the last one would already be enough (and more friendly on the reader :smile: ) - hence the proposal.

Anyway, good night and happy coding down there! ;)

PR #413, opened 2015-04-21
fix(publish) make the published items move to the published stage

marwoodandrew:


server/features/content_publish.feature
```diff
@@ -35,7 +35,8 @@ Feature: Content Publishing
       """
       And we publish "#archive._id#"
       Then we get OK response
-      And we get existing resource
+      When we get "/archive/#archive._id#"
```
amagdas: If this is needed, it means the returned doc is not having all the updates and the client side needs to refetch the item.

marwoodandrew: True, will fix

amagdas:
:+1:

petrjasek:
:+1:

akintolga:
:+1:

PR #416, opened 2015-04-21
[SD-1670] New archive item is deleted if it is closed before it is saved

akintolga:
- When a new content item is created, there's an empty record in the archive collection. If that content item is closed by user without making any change (that is still version 1) then the item sits there with no information in it.
- Now if the user closes the newly created item without making any change, system will delete the empty record.
- These type of items will also be deleted on session recycle and log out events as well.
- This PR also addresses the problem in image upload: When user selects an image but then presses cancel button we are left with the empty item. Now when an image is loaded we don't save it until user presses the save button.
- This also solves the issue [SD-2140]


amagdas:
:+1:

mugurrus:
:+1:

petrjasek:
:+1:

ioanpocol:
:+1:

PR #418, opened 2015-04-21
feat(publish): Added new subscriber type "both"

eozan:


plamut:
:+1:
I assume the empty value of `subscriber.subscriber_type` is interpreted as "any of the two types", right?

eozan:
that interpretation will be done on server, but yes, I think that's the idea.

eozan:
(Failing tests are already fixed in https://github.com/superdesk/superdesk/pull/410)

plamut:
OK, good, just wanted to make sure.

amagdas:
:+1:

akintolga:
:+1:

PR #419, opened 2015-04-21
feat(authoring): simple search in aggregate view

petrjasek:
implement search field in aggregate widget

### DIFF
--- a/client/app/scripts/superdesk-desks/desks.js
+++ b/client/app/scripts/superdesk-desks/desks.js
@@ -73,29 +73,42 @@
                 allowed: '=',
                 showEmpty: '=?',
                 selected: '=?',
-                action: '&'
+                action: '&',
+                filter: '='
             },
             link: function(scope, elem) {
-
-                var query = search.query({});
-                query.filter({term: {'task.stage': scope.stage}});
-                query.size(10);
-                var criteria = {source: query.getCriteria()};
-
-                scope.loading = true;
-
-                api('archive').query(criteria).then(function(items) {
-                    scope.loading = false;
-                    scope.items = items._items;
-                    scope.total = items._meta.total;
-                }, function() {
-                    scope.loading = false;
-                });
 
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-desks/desks.js
```diff
@@ -73,29 +73,43 @@
...
+                    var criteria = {source: query.getCriteria()};
+
+                    scope.loading = true;
+                    scope.items = scope.total = null;
+                    api('archive').query(criteria).then(function(items) {
+                        scope.loading = false;
+                        scope.items = items._items;
+                        scope.total = items._meta.total;
+                    }, function() {
+                        scope.loading = false;
```
plamut: This `scope.loading` repeats in both the success and the error handler, better to move it into `finally()` to avoid code repetition.

client/app/scripts/superdesk-desks/views/aggregate-widget.html
```diff
@@ -1,7 +1,10 @@
 <div class="content-search" ng-class="{extended: agg.selected}" ng-controller="AggregateWidgetCtrl as agg">
     <div class="search-container">
         <div class="search-box">
-            <input type="text" placeholder="{{ :: 'Search'|translate }}" ng-model="query">
+            <input type="text" placeholder="{{ :: 'Search' | translate }}"
+                ng-model="query"
+                ng-model-options="{debounce: 500}"
```
plamut: Glad to see that `ng-model-options` are becoming more common in our code. :wink:

petrjasek: yep, we should probably define what debounce we should use for all fields so it's consistent

plamut: That would make sense indeed.

plamut:
:+1:

amagdas:
:+1:

PR #420, opened 2015-04-21
fix(tooltips): Added custom tooltips on more buttons

pavlovicnemanja92:
SD-2065

### DIFF
--- a/client/app/scripts/superdesk-archive/views/list.html
+++ b/client/app/scripts/superdesk-archive/views/list.html
@@ -1,11 +1,12 @@
 <div class="subnav">
 	<div class="button-stack right-stack">   
 	    <div class="navbtn dropdown strict highlights-dropdown" 
+            tooltip="{{ 'Mark item' | translate}}" tooltip-placement="left"
             dropdown sd-package-highlights-dropdown 
             ng-if="type === 'archive' && selected.desk">
         </div>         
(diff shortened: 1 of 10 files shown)


client/app/scripts/superdesk-archive/views/list.html
```diff
@@ -1,11 +1,12 @@
 <div class="subnav">
 	<div class="button-stack right-stack">   
 	    <div class="navbtn dropdown strict highlights-dropdown" 
+            tooltip="{{ 'Mark item' | translate}}" tooltip-placement="left"
```
petrjasek: would be good to use one time bindings for translate eg `{{ :: 'Mark item' | translate }}`

client/app/scripts/superdesk-search/views/item-sortbar.html
```diff
@@ -9,7 +9,7 @@
 	        </li>
 	    </ul>
 	</div>
-	<button ng-click="toggleDir()" class="pull-left direction" title="{{active.dir}}">
+	<button ng-click="toggleDir()" class="pull-left direction" tooltip="{{ :: active.dir}}" tooltip-placement="bottom">
```
plamut: Just to be sure - the value of the `active.dir` expression does not change in the lifetime of the view?

PR #429, opened 2015-04-22
Travis public api

actionless:


plamut:
:+1:
(but let's just wait for Travis to finish before merging, just in case)

mugurrus:
Created and merged PR 441 from this one

PR #430, opened 2015-04-22
feat(content): sync multi content view

petrjasek:
in case an item is send to a different stage, both previous and new stage lists will refresh

requires https://github.com/superdesk/eve/pull/19


amagdas:
:+1:

eozan:
:+1:

PR #433, opened 2015-04-23
fux(users) - Show the same error message on notification and on edit scr...

ioanpocol:
...een when is no info about what fields are with error

Conflicts:
	client/app/scripts/superdesk-users/users.js


petrjasek:
:+1:

mugurrus:
:+1:

PR #436, opened 2015-04-23
feat(multi-content): refresh stages when item is sent

petrjasek:
it will reload content for both previous and next stage

includes required eve update


ioanpocol:
:+1:

mugurrus:
:+1:

PR #439, opened 2015-04-23
fix(multi-content): fix item preview in multi content widget

petrjasek:
it was broken after using `controller as`


ioanpocol:
:+1:

mugurrus:
:+1:

PR #440, opened 2015-04-23
fix(online users): Styled checkbox for online users

pavlovicnemanja92:
SD-2155

### DIFF
--- a/client/app/scripts/superdesk-desks/views/main.html
+++ b/client/app/scripts/superdesk-desks/views/main.html
@@ -3,12 +3,12 @@
     	<ul>
 	    	<li ng-repeat="v in views" ng-class="{active: v === view}">
 	    		<a href="" id="{{v}}" ng-click="setView(v)" translate>{{v}}</a>
-	    		<div ng-if="v === view && view === 'users'">
-	    			<input type="checkbox" ng-model="online_users" name="online_users" id="online_users" ng-change="changeOnlineUsers(online_users)">
-	    			<label translate>online users</label>
-	    		</div> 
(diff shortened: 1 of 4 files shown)


ioanpocol:
:+1:

petrjasek:
:+1:

client/app/scripts/superdesk-desks/views/main.html
```diff
@@ -3,12 +3,12 @@
     	<ul>
 	    	<li ng-repeat="v in views" ng-class="{active: v === view}">
 	    		<a href="" id="{{v}}" ng-click="setView(v)" translate>{{v}}</a>
-	    		<div ng-if="v === view && view === 'users'">
-	    			<input type="checkbox" ng-model="online_users" name="online_users" id="online_users" ng-change="changeOnlineUsers(online_users)">
-	    			<label translate>online users</label>
-	    		</div> 
+                        <div class="online-users" ng-if="v === view && view === 'users'">
```
plamut: Can't we just say `v === 'users'`? :wink:

ioanpocol: No

plamut: Why not? The identity operator is transitive...

ioanpocol: Yes, you can do ng-if="v === view && v === 'users'" if you wish.
The point is that v and view should have value 'users' so just v==='users' is not enough.

plamut: Oh, OK, then it's fine of course, I thought it was just about the `v`. Thanks for clarifying!

PR #442, opened 2015-04-23
fix(preferences): hide preferences without label from ui

petrjasek:
and make sure we use defined label/category not what gets saved into
db so this change will work also for existing users


ioanpocol:
:+1:

amagdas:
:+1:

PR #443, opened 2015-04-23
feat(master desks): Add the option to edit the user from master desk, us...

ioanpocol:
...ers tab


petrjasek:
:+1:
maybe you could use `user` as model for modal and avoid this `modalActive`

plamut:
:+1: 
(I agree that we can probably get rid of the `modalActive` flag)

ioanpocol:
Deleted modalActive and use user instead

PR #445, opened 2015-04-24
Add /packages endpoint to public API

plamut:
Please have a closer look at the `package` schema definition and verify it, thanks!

### DIFF
--- a/server/publicapi/packages/resource.py
+++ b/server/publicapi/packages/resource.py
@@ -1,0 +1,49 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.resource import Resource
+
+
+class PackagesResource(Resource):
+    """A class defining and configuring the /packages API endpoint."""
+
+    # Example of an ID of an object in database (whitout quotes):
+    #
+    #   "tag:example.com,0000:newsml_BRE9A605"
+    #
+    item_url = 'regex("[\w,.:-]+")'
+
+    schema = {
+        'guid': {'type': 'string'},
+        'type': {'type': 'string'},
+        'unique_id': {'type': 'integer'},
+        'unique_name': {'type': 'string'},
+        'pubstatus': {'type': 'string'},
+        'firstcreated': {'type': 'datetime'},
+        'versioncreated': {'type': 'datetime'},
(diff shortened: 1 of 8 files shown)


petrjasek:
:+1: schema looks good

server/publicapi/packages/service.py
```diff
@@ -0,0 +1,22 @@
...
+import logging
+from superdesk.services import BaseService
+
+
+logger = logging.getLogger(__name__)
+
+
+class PackagesService(BaseService):
+    """
+    Data layer providing the CRUD functionality for the `package`
```
ioanpocol: "Data Layer" is not an accurate description.

ioanpocol:
Since here is not a lot of functionality I think that more relevant are a few behave tests than unittests.

ioanpocol:
:+1:

server/publicapi/packages/resource.py
```diff
@@ -0,0 +1,49 @@
...
+        'guid': {'type': 'string'},
+        'type': {'type': 'string'},
+        'unique_id': {'type': 'integer'},
+        'unique_name': {'type': 'string'},
+        'pubstatus': {'type': 'string'},
+        'firstcreated': {'type': 'datetime'},
+        'versioncreated': {'type': 'datetime'},
+        'original_creator': {'type': 'string'},
+        'version_creator': {'type': 'string'},
+        'groups': {'type': 'list'},
```
ioanpocol: Probably it will help to detail in schema how the groups will look like. Is groups identical with what we have on regular package resource?

server/publicapi/packages/service.py
plamut: Yeah, I was contemplating what term to use here. Suggestions welcome, because a misleading comment is even worse than no comment at all.

plamut:
@ioanpocol Behave tests will also be added once we have the whole stack (including the DB) in place. Though IMO they should not be a replacement for unit tests, but rather something that complements them (i.e. testing the application from the user's perspective as opposed to the developer's perspective what unit tests do).

server/publicapi/packages/resource.py
plamut: I believe yes, but I can't be 100% sure. I basically extracted the schema for `packages` from the behave tests and from the example output of the API (http://docs.superdeskcontentapi.apiary.io/#reference/content-packages/package/fetch-a-single-package-item)

plamut:
I've updated the docstrings to more accurately describe the purpose of the *Service classes.

Regarding the packages schema - after a discussion with @petrjasek we concluded that it's not necessary to specify it in a more detailed way. The schema is only used when doing the insert/update operations, but we only allow fetching a resource, thus it mostly serves the documentation purposes for the code readers.

Merging.

PR #447, opened 2015-04-24
fix(dictionaries): validate dictionary file type

petrjasek:
and set form as invalid in case it's not plain text


ioanpocol:
:+1:

mugurrus:
:+1:

PR #448, opened 2015-04-24
fix(style issue): Fixed styled issue on user tab in master desk.

pavlovicnemanja92:
SD-2169

### DIFF
--- a/client/app/scripts/superdesk-users/views/edit-form.html
+++ b/client/app/scripts/superdesk-users/views/edit-form.html
@@ -25,10 +25,7 @@
               <span sd-switch ng-model="$parent._active" ng-click="toggleStatus(_active)"></span>
             </div>
         </div>
-        <div class="error-holder">
-          <p class="alert alert-error" ng-show="error && !error.message" translate>Please fix the issues below.</p>
-          <p class="alert alert-error" ng-show="error && error.message" translate>{{error.message}}</p>
-        </div>
+
         
         <div class="details-info form-flat" ng-class="{shifted : error}">
             <div class="title">{{ 'General' | translate }} <span class="required-info" translate>* mandatory fields</span></div>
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-users/views/edit-form.html
```diff
@@ -25,10 +25,7 @@
               <span sd-switch ng-model="$parent._active" ng-click="toggleStatus(_active)"></span>
             </div>
         </div>
-        <div class="error-holder">
-          <p class="alert alert-error" ng-show="error && !error.message" translate>Please fix the issues below.</p>
-          <p class="alert alert-error" ng-show="error && error.message" translate>{{error.message}}</p>
```
petrjasek: imo this should stay somewhere

pavlovicnemanja92: I talk with Fritz about that, and we agreed that error message on a top of the window is enough. I can see with him to put this error somewhere too if it is important.

petrjasek: and it's with the same `error.message` ? then it is ok

ioanpocol: Not always is the same error and the problem is that the notification disappear in 5s or so.
Probably in this case because is the same error we can hide one of them.
The line with Please fix issue below should not be deleted, it will be showed on input validation errors.
The second line can be deleted.

petrjasek: imo it should be the other way ;) line with `please fix issues below` is not needed because you can see the errors below, but the other message is important

ioanpocol: The other is shown all the time as notification, so to delete both line seams to be ok.

ioanpocol:
:+1:

petrjasek:
:+1:

PR #451, opened 2015-04-24
Register the public API "packages" app so that the endpoint becomes avai...

plamut:
Fix: make the `/packages` endpoint publicly available by adding the `publicapi.packages` app to the list of installed applications.


mugurrus:
:+1:

ioanpocol:
:+1:

PR #459, opened 2015-04-27
fix(preferences): fix fetching preferences

petrjasek:
there was an error for preference which were a list of items


sivakuna-aap:
:+1:

PR #460, opened 2015-04-28
[SD-2193](fix) when a stage is no longer set as default publish publishe...

marwoodandrew:
...d items will no longer move there


sivakuna-aap:
:+1:

amagdas:
We need a test for this so we are sure no regressions will occur.

akintolga:
:+1:

amagdas:
:+1:

PR #462, opened 2015-04-28
(fix) NITF Formatter fails to format when subject doesn't have 'qcode'

sivakuna-aap:
Used get_resource_service() instead of MoveService()


amagdas:
flake8 issue

amagdas:
:+1:

akintolga:
:+1:

PR #463, opened 2015-04-28
fix(icons): Change icon on widget for opening to edit mode

pavlovicnemanja92:
There is no taks in JIRA for this.


ioanpocol:
:+1:

PR #465, opened 2015-04-28
fix(users): Allow the user to change the status of other logged user

ioanpocol:


petrjasek:
:+1:

server/apps/users/services.py
```diff
@@ -96,19 +96,22 @@ def __is_invalid_operation(self, user, updates, method):
         Checks if the requested 'PATCH' or 'DELETE' operation is Invalid.
         Operation is invalid if one of the below is True:
             1. Check if the user is updating his/her own status.
-            2. Check if the user is changing the status of other logged-in users.
-            3. A user without 'User Management' privilege is changing role/user_type/privileges
+            2. Check if the user is changing the changing role/user_type/privileges of other logged-in users.
```
mugurrus: This should write like: "Check if the user is changing the role/..."

mugurrus:
:+1:

PR #468, opened 2015-04-28
fix(archive): remove empty revisions

petrjasek:
when creating an item set its revision to 0
and filter those from content list and versions list


akintolga:
:+1:

sivakuna-aap:
:+1:

ioanpocol:
:+1:

PR #471, opened 2015-04-28
Content api - prepopulation functionality

mugurrus:


server/publicapi/packages/service.py
```diff
@@ -9,13 +9,13 @@
...
 import logging
-from superdesk.services import BaseService
+from publicapi.items.service import ItemsService
 
 
 logger = logging.getLogger(__name__)
 
 
-class PackagesService(BaseService):
+class PackagesService(ItemsService):
```
plamut: Why is the packages service now a subclass of the items service? Wouldn't it be better and more correct if both inherited from the same base class and not from each other? (if the purpose is to share some common functionality between the two)

mugurrus: Currently the whole item functionality is needed in packages so it makes sense for now. If and when item and package functionality diverge creating a base class for them will be trivial. Currently as Petr says I think it's premature optimization :)

plamut:
There are some (docstring) comments missing at a couple of places and the unit tests do not cover all the changes, but nothing critical (I'll add those in a separate PR).

The only question we need to resolve is the hierarchical relation between the `ItemsService` and `PackagesService`.

server/publicapi/packages/service.py
plamut: Hm... I don't like it that much but OK. My main concern is that at some point somebody will forget to decouple the relation if the functionality diverges (it's always like that :smile: ).

I think it's best if we add a comment reminding people of that, it elegantly addresses this.

petrjasek: this inheritance makes sense to me, package is a specific item type

plamut: @petrjasek Isn't a package more of a _container_ of items? (at least by my understanding)
It's like saying that a box is a specific type of shoe, because it can contain different types of shoes and other boxes.

Please correct me if this analogy is not correct.

petrjasek: imo items are metadata + content, and while the content can be different, the metadata part is similar. so it's more about those boxes than shoes.

mugurrus:
Writing comments on obvious classes/methods is not helpful, only makes the file bigger. I find it harder to navigate around comments that don't help me.

mugurrus:
There's no issue with the hierarchical relation, as I said the base for items and packages is really the items class. If we had a functionality needed in items but not in packages then it would make sense to create a separate base.

ioanpocol:
:+1:

server/publicapi/packages/service.py
mugurrus: "Hm... I don't like it that much but OK. My main concern is that at some point somebody will forget to decouple the relation if the functionality diverges (it's always like that :smile: )."

That's called refactoring :) It's about detecting when code structure does not match the functionality :)

plamut:
Re: comments - we need to take into account that a lot of people (hopefully) will be contributing to the project. What might seem obvious to people involved in the project for several months/years might not be obvious for those who are just getting started and/or only want to submit a small bug fix/improvement.

Since we lack good documentation, code comments are the way to go. It only takes a couple of minutes to write them, but they potentially save a lot more time for someone just glancing over the code.

I suggest we merge this to keep the things going and address other things separately.

:+1:

server/publicapi/packages/service.py
plamut: Oh OK, viewing the items/packages this way (i.e. as a combination of metadata and content) actually makes the hierarchy to feel more "natural".

@mugurrus Can we be sure that every potential contributor out there will have a good enough oversight of the project to be able to detect when refactoring is needed? :wink: (at least until we have good documentation in place).

PR #473, opened 2015-04-29
Displaying Published Items 

akintolga:


server/apps/publish/published_item.py
```diff
@@ -0,0 +1,64 @@
...
+            item['_id']=item['item_id']
+        return items
+
+    def get_other_published_items(self, _id):
+        query = {'query': {'filtered': {'filter': {'term': {'item_id': _id}}}}}
+        request = ParsedRequest()
+        request.args = {'source': json.dumps(query)}
+        return super().get(req=request, lookup=None)
+
+    def update_other_published_items(self, _id, state):
```
akintolga: @petrjasek I don't understand why my patches fail with "OriginalChanged" everytime here? I know that I do some field manipulation on line:45 when I post a new item and I know that is the reason. But I have to do that because I need to be able to post different versions of the same item.

petrjasek: any idea if etag in elastic matches with etag in mongo for those items?

akintolga: I checked them. They are the same. It is the etag calculated on post function of the BaseService.

server/apps/publish/published_item.py
```diff
@@ -0,0 +1,64 @@
...
+        items = self.get_other_published_items(_id)
+        for item in items:
+            # resolve_document_etag([item], self.datasource)
```
akintolga: If I recalculate the etag of the item here it is different, I'm not sure if it is expected.

server/apps/publish/published_item.py
```diff
@@ -0,0 +1,64 @@
...
+            super().system_update(item['_id'], {'last_publish_action': state}, item)
```
petrjasek: is the `item['_id']` here an `ObjectId` instance or a string?

mugurrus:
:+1:

amagdas:
Rebased and pushed upstream.

PR #476, opened 2015-04-29
fix(spellcheck): allow words with numbers in dictionary

petrjasek:
and spellcheck such words

### DIFF
--- a/client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
+++ b/client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
@@ -61,10 +61,8 @@ function rgb(color) {
  * Replace given dom elem with its contents
  */
 function replaceSpan(elem) {
-    var range = document.createRange();
-    range.selectNode(elem);
-    var fragment = range.createContextualFragment(elem.innerHTML);
-    elem.parentNode.replaceChild(fragment, elem);
+    var text = document.createTextNode(elem.textContent);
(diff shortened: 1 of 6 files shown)


ioanpocol:
:+1:

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -123,16 +121,21 @@ function SpellcheckService($q, api, dictionaries, editor) {
...
+
         return getDict().then(function() {
-            var words = text.match(/[a-zA-Z\u00C0-\u1FFF\u2C00-\uD7FF]+/g),
+            var words = text.match(/[0-9a-zA-Z\u00C0-\u1FFF\u2C00-\uD7FF]+/g),
                 errors = [];
             angular.forEach(words, function(word) {
-                var lowerWord = word.toLowerCase();
-                if (!dict[lowerWord]) {
-                    errors.push(word);
+                if (isNaN(word)) {
```
plamut: Just checking - do we indeed want to interpret strings like `1e3` and `0x5` as numbers and not report them as errors?

amagdas:
:+1:

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
petrjasek: yep exactly

PR #478, opened 2015-04-29
added steps to install newest docker

girgen79:
for all versions of OS

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Use [docker-compose](http://fig.sh "") and the config from the `docker` folder o
 ```sh
 $ sudo apt-get install docker.io python2 python-virtualenv
 ```
-
+and install [the newest docker](https://docs.docker.com/installation/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ Use [docker-compose](http://fig.sh "") and the config from the `docker` folder o
 ```sh
 $ sudo apt-get install docker.io python2 python-virtualenv
 ```
-
+and install [the newest docker](https://docs.docker.com/installation/).
```
petrjasek: so is there a need to install docker.io in previous step?

leksa: I have passed this problem before. 
Repo docker.io from apt-get give us the old docker. It will give problem when the docker up, running client and server in different version.
http://askubuntu.com/questions/472412/how-do-i-upgrade-docker

amagdas:
:+1:

PR #479, opened 2015-04-29
fix(multi-edit): Fix preview item

ioanpocol:


petrjasek:
there was an e2e test for this, we should put it back

ioanpocol:
added back e2e test

petrjasek:
:+1:

PR #480, opened 2015-04-29
feat(dropdown): HTML & CSS structure of new nested dropdown style

pavlovicnemanja92:
This is relative to SD-2101 task in JIRA. This is complete HTML and CSS structure, and I provide a little  jQuery code that will need to be implement into angular js file relative to dropdown menus.

Also, I create this with dummy data, and as we speak on today's scrum, someone need to create dynamic structure for this.

Currently, I put all the HTML and jQuery code in metadata-terms.html, and it shows on every sdTypeaHead component in right side bar. We will need this only on subject field.


amagdas:
:+1:

client/app/scripts/superdesk-authoring/metadata/views/metadata-terms.html
```diff
@@ -1,12 +1,32 @@
 <div class="term-editor data visible">
-	<div sd-typeahead items="terms" term="selectedTerm" search="searchTerms(term)" select="selectTerm(item)" data-disabled="disabled">
-		<ul>
-			<li typeahead-item="t" ng-repeat="t in terms">{{ t.name }}</li>
-		</ul>
-	</div>
+    <div sd-typeahead items="terms" term="selectedTerm" search="searchTerms(term)" select="selectTerm(item)" data-disabled="disabled">
+        <ul>
+            <li typeahead-item="t" ng-repeat="t in terms ">{{ t.name }}</li>
```
plamut: Can we use one-time binding here? The name of each term does not change?

client/app/scripts/superdesk-authoring/metadata/views/metadata-terms.html
```diff
@@ -1,12 +1,32 @@
...
+            <button ng-if="!tree[term.qcode]" ng-click="selectTerm(term)">{{ :: term.name }}</button>
+        </li>
+    </ul>
 </div>
 <div class="terms" ng-if="item[field].length">
-	<ul>
-		<li ng-repeat="t in item[field]" ng-click="removeTerm(t)">{{ t.name }}</li>
-	</ul>
+    <ul>
+        <li ng-repeat="t in item[field]" ng-click="removeTerm(t)">{{ t.name}}</li>
```
plamut: Similar question to above (`track by` and `::`)

ioanpocol:
:+1:

pavlovicnemanja92:
:+1:

PR #483, opened 2015-04-29
Content api

plamut:
Merging updates in master into content-api (a long running topic branch).

Edit: sorry for so many commits, I was a bit in a hurry (should've probably squashed all these "master" commits into a single one).

### DIFF
--- a/server/publicapi/__init__.py
+++ b/server/publicapi/__init__.py
@@ -18,6 +18,7 @@ be more specific, an `Eve framework <http://python-eve.org/>`_ application).
 from eve import Eve
 from eve.io.mongo.mongo import MongoJSONEncoder
 from eve.render import send_response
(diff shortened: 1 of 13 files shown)


amagdas:
I think a rebase is needed instead of merge...

plamut:
The rebase was kind of messy (that's why I wanted to avoid it first after a discussion with @petrjasek ), but in the end I got it through.

Should I simply do a force push now? I've checked with @mugurrus and he's not working on this API branch at the moment, so it should be OK.

amagdas:
Use the force Luke :)

plamut:
Force applied. :smile:

plamut:
Also, I suggest we start rebasing this branch on a more regular basis - but *not* without prior communication with anyone potentially involved, of course.

PR #484, opened 2015-04-29
(feat) [SD-1990] Publishing channel sequence numbering

sivakuna-aap:


amagdas:
:+1:

akintolga:
Siva, do we need to save the sequence number to the queue as well? What do you think?

sivakuna-aap:
@akintolga is there any requirement which requires the sequence number in queue?

akintolga:
I don't think so. But I'd say it would be a good check the see what numbers are assigned. A sanity check.

sivakuna-aap:
Ok. I will add make the changes and update the pr

sivakuna-aap:
Adding published sequence number to both publish_queue and formatted_item

akintolga:
:+1:

amagdas:
:+1:

PR #494, opened 2015-05-01
(fix) [SD-2213] Adding a duplicate subscriber gives unlcear error

sivakuna-aap:


amagdas:
:+1: 
It would be nice to extract this into a helper function so we can reuse it in other places.

akintolga:
:+1:

PR #499, opened 2015-05-03
Add a file with sample NINJS data

plamut:
The file contains data with more variety (item types, authors, creation dates, etc.) for easier testing and development of the public API.

Resolves SD-2198

P.S.: The items' and packages schema differs a bit from the current format we use, because it closely follows the NINJS specification.
P.P.S.: The packages in demo data contain full items, not just the references to them. We might need to change them to only contain links to the contained items (i.e. the `uri` field of each object in the package's `associations` list).

### DIFF
--- a/server/publicapi/demo_data/data.json
+++ b/server/publicapi/demo_data/data.json
@@ -1,0 +1,483 @@
+{
+    "_id": "tag:demodata.org,0001:ninjs_XYZ123",
+    "uri": "http://new.demodata.org/archive/20150316pic1234",
+    "type": "picture",
+    "version": 3,
+    "versioncreated": "2015-03-16T08:12:00Z",
+    "usageterms": "You can use this photo in any way you want.",
+    "byline": "Andy Catlover",
+    "headline": "Amazing! A very cute fluffy cat!",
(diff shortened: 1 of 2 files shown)


amagdas:
:+1:

server/publicapi/demo_data/data.json
```diff
@@ -0,0 +1,483 @@
+{
+    "_id": "tag:demodata.org,0001:ninjs_XYZ123",
+    "uri": "http://new.demodata.org/archive/20150316pic1234",
```
mugurrus: URI should not be stored, it's overwritten anyway

mugurrus:
:+1:

server/publicapi/demo_data/data.json
plamut: Good catch, I'll update.

(we'll probably need to set this field on the fly from the `self` link that Eve generates)

plamut: BTW, what about the `href` attribute of the images' renditions objects? That should probably stay, right, because it might link to a resource outside the API database, e.g. to something in a CDN.

mugurrus: yep

PR #500, opened 2015-05-04
(feat) [SD-1989] Content can be queued to a publishing channel that is closed

sivakuna-aap:
Supposed to use `updates` branch. But there were some issues and big pending PR, that's the reason used `master` branch for this feature.

### DIFF
--- a/server/superdesk/publish/publish_content.py
+++ b/server/superdesk/publish/publish_content.py
@@ -63,8 +63,16 @@ def publish(subscriber, destination):
 def transmit_items(queue_items, subscriber, destination):
     failed_items = []
 
+    output_channels = superdesk.get_resource_service('output_channels').get(req=None, lookup={})
+    output_channels = {str(output_channel['_id']): output_channel for output_channel in output_channels}
+
     for queue_item in queue_items:
+        # Check if output channel is active
+        if(output_channels.get(str(queue_item['output_channel_id']), {})).get('is_active', False):
+            continue
+
         try:
+
             # update the status of the item to in-progress
             queue_update = {'state': 'in-progress', 'transmit_started_at': utcnow()}
             superdesk.get_resource_service('publish_queue').patch(queue_item.get('_id'), queue_update)
--- a/server/superdesk/publish/publish_content.py
+++ b/server/superdesk/publish/publish_content.py
@@ -23,12 +23,16 @@ class PublishContent(superdesk.Command):
     """Runs deliveries"""
 
     def run(self, provider_type=None):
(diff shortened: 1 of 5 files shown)


akintolga:
There's one thing missing from the celery tasks:
https://github.com/superdesk/superdesk/blob/master/server/superdesk/publish/publish_content.py#L27

Here I was checking if the subscriber is active, but with this task we also need to check if the output channel is active or not. What do you think?

amagdas:
This looks a bit fishy to me, I don't think we should show an error message to the user in Authoring screen if one of the output channels is closed. Maybe a warning...

mugurrus:
@amagdas When one tries to publish to a closed channel, shouldn't that be considered an error?

amagdas:
Actually no, the item is queued for publishing on that channel anyway, giving the fact that usually there are multiple channels involved when publishing an article I don't think we should say it failed when in fact it didn't and the journalist I don't think he cares about the channel level, the sysadmin or a similar role would do.

sivakuna-aap:
@akintolga thanks for pointing it out. I have added.

sivakuna-aap:
@amagdas that's how the promise has been resolved and because of that am doing like this. Since there are talks about refactoring the authoring code I didn't change much. 

Eve throws all exceptions as errors. If we need to change the alert type to warning then we need to check for specifics. 

what do u think?

amagdas:
Checking for the specific error code should be OK.

server/superdesk/publish/publish_content.py
```diff
@@ -63,8 +63,16 @@ def publish(subscriber, destination):
 def transmit_items(queue_items, subscriber, destination):
     failed_items = []
 
+    output_channels = superdesk.get_resource_service('output_channels').get(req=None, lookup={})
```
akintolga: Just a minor thing: Since we know the output_channel_id at this point we can use the lookup to get that one only: 
```
lookup = {'_id': 'output_channel_id'}
```

Or we can do get all of them in previous function and feed the output channel to each thread starts with .apply_sync. What do you think?

sivakuna-aap: Reason for getting all output_channels is to avoid multiple calls to Mongo. We can get only active channels if the number of output_channels are more.

If we need gracefulness then yes, I do think it's better to fetch channels in the previous function and start the thread.

akintolga: I see. Happy.

sivakuna-aap: I still see some room for improvement. I'll refactor

akintolga:
:+1:

ioanpocol:
:+1:

amagdas:
:+1:

PR #502, opened 2015-05-04
[SD-1992] Command to export the AAP text archive into the superdesk text archive

marwoodandrew:


sivakuna-aap:
Better to have all modules related to `text_archive` in one place

server/apps/import_text_archive/app_initialize.py
```diff
@@ -0,0 +1,255 @@
...
+            return e
+        return None
+
+    def _get_head_value(self, doc, field):
+        el = doc.find('dcdossier/document/head/' + field)
+        if el is not None:
+            return el.text
+        return None
+
+    def _process_iptc_codes(self, item):
```
mdhaman: Looks like duplicate code from update_ingest.py.

marwoodandrew:
The import command will only be required to build the initial archive, the end point will live on!

akintolga:
:+1:

server/apps/archive/__init__.py
```diff
@@ -31,6 +31,7 @@
 from superdesk.celery_app import celery
 from .saved_searches import SavedSearchesService, SavedSearchesResource, \
     SavedSearchItemsResource, SavedSearchItemsService
+from .text_archive import TextArchiveResource, TextArchiveService
```
amagdas: We should move this into it's own module.

server/apps/import_text_archive/app_initialize.py
```diff
@@ -0,0 +1,250 @@
...
+            if not original:
+                res.post([item])
+            else:
+                res.patch(original['_id'], item)
+
+            if self._limit:
+                self._limit -= 1
+            # print(item)
+
+    def run(self, start_id, user, password, url, query, limit):
```
amagdas: Can we move the `run` function at the top of the class, after the options declaration? Will make things easier to read.

server/apps/import_text_archive/app_initialize.py
```diff
@@ -0,0 +1,250 @@
...
+        self._api_login()
+
+        x = self._get_bunch(self._id)
+        while x:
+            self._process_bunch(x)
+            x = self._get_bunch(self._id)
+            if self._limit is not None and self._limit <= 0:
+                break
+
+        print('finished text archive import')
```
amagdas: Does it make sense to kick `ImportFromMongo` command if this went fine?

server/apps/archive/__init__.py
marwoodandrew: OK

server/apps/import_text_archive/app_initialize.py
marwoodandrew: OK

server/apps/import_text_archive/app_initialize.py
marwoodandrew: No need, this populates both mongo and elastic, might be a problem in it's self! So far the performance is limited by DC/REST not elastic. I guess this may change as the dataset gets big.

amagdas: Sounds good :+1:

amagdas:
:+1:

PR #505, opened 2015-05-04
fix(auth): reload route when it's a redirect

petrjasek:
after session expiry it would stop without an active route so after
login it would continue nowhere.


plamut:
:+1: 
Would be nice to add the tests for this, too. :wink:

vied12:
LGTM ! :+1:

amagdas:
:+1:

petrjasek:
@plamut it wasn't you, I've updated it :smile:

PR #507, opened 2015-05-04
Use route.href to build the link of the `main link`

vied12:
Because using the label to generate the link is not always relevant

### DIFF
--- a/client/app/scripts/superdesk/analytics/analytics.js
+++ b/client/app/scripts/superdesk/analytics/analytics.js
@@ -23,7 +23,9 @@ define([
 
         .run(['$rootScope', 'analytics', function($rootScope, analytics) {
             $rootScope.$on('$routeChangeSuccess', function(ev, route) {
-                analytics.track(route.$$route);
+                if (angular.isDefined(route)) {
+                    analytics.track(route.$$route);
+                }
             });
         }]);
 });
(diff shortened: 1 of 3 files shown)


amagdas:
:+1:

client/app/scripts/superdesk/analytics/analytics.js
```diff
@@ -23,7 +23,9 @@ define([
 
         .run(['$rootScope', 'analytics', function($rootScope, analytics) {
             $rootScope.$on('$routeChangeSuccess', function(ev, route) {
-                analytics.track(route.$$route);
+                if (angular.isDefined(route)) {
+                    analytics.track(route.$$route);
```
plamut: I'm not sure this is the best way, attributes prefixed by `$$` are considered internal to Angular and should not be directly. They might, for instance, change without notice in future framework versions.

Is there any other way to achieve the same functionality?

vied12:
@plamut Updated. It seems to work exactly in the same way without `$$route`.

plamut:
Do we have a test for that? But other than that, good thing that we got rid of `$$route`.

:+1:

vied12:
Yes there is a test, I also updated it

plamut:
Great, feel free to merge it then.

vied12:
I don't have the right. Please do it

PR #513, opened 2015-05-05
fix(dropdown icon): Fixed position of nested dropdown icon

pavlovicnemanja92:
There is no task in JIRA for this. This is just a small fix


plamut:
:+1:

PR #517, opened 2015-05-05
feat(dashboards): add multi dashboards support

petrjasek:
there is new resource `/dashboards` that can handle different types
of dashboards, it's user and desk dashboards atm

SD-2208


amagdas:
A rebase is needed...

petrjasek:
will merge this to the branch, it changes how desk dropdown works which breaks the e2e tests, but it's part of prototype so not fixing it atm

PR #537, opened 2015-05-07
fix(multiedit): fix image size in multiedit

petrjasek:
there was firefox specific issue with img max-width: 100%, added a workaround


plamut:
:+1:

amagdas:
:+1:

PR #539, opened 2015-05-07
Add an option to provide a detailed description to all custom exceptions

plamut:
This is needed in cases when the static pre-defined exception error messages by themselves are not descriptive enough.

Resolves [SD-2262](https://dev.sourcefabric.org/browse/SD-2262).


mugurrus:
:+1:

petrjasek:
:+1:

PR #540, opened 2015-05-07
fix(authoring): fix backspace/delete behaviour

petrjasek:
there were some changes going on keydown, not it only happens on input


amagdas:
:+1:

sivakuna-aap:
:+1:

PR #541, opened 2015-05-08
Fix (package item close): On item close, return to package

Aca-jov:
SD-1877

I'm not sure if there is better way to check if previousRoute is packaging


plamut:
Not sure either, but we must get rid of relying on the `$$route`attribute, it's internal to Angular and can change without notice.

mugurrus:
:+1:

plamut:
Hey, wait, there was only one thumb up, this should have been merged just yet. :wink:

PR #542, opened 2015-05-08
public API: return HTTP 422 if request contains unexpected parameters

plamut:
Resolves [SD-2242](https://dev.sourcefabric.org/browse/SD-2242).

NOTE: the behave tests will be updated under a [separate ticket](https://dev.sourcefabric.org/browse/SD-2171) that we've opened.

### DIFF
--- a/server/publicapi/items/service.py
+++ b/server/publicapi/items/service.py
@@ -8,11 +8,13 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-from flask import current_app as app
 import logging
-from urllib.parse import urljoin, quote
 
+from eve.utils import ParsedRequest
+from flask import current_app as app
+from publicapi.errors import UnexpectedParameterError
 from superdesk.services import BaseService
+from urllib.parse import urljoin, quote
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


amagdas:
:+1:

ioanpocol:
:+1:

server/publicapi/items/service.py
```diff
@@ -26,6 +28,60 @@ class ItemsService(BaseService):
...
+        :type req: `eve.utils.ParsedRequest`
+        :param dict lookup: requested item lookup, contains its ID
+
+        :return: requested item (if found)
+        :rtype: dict or None
+        """
+        request_params = req.args or {}
+        params = list(request_params.keys())
+
+        if len(params) > 0:
```
mugurrus: I think it would be better to split parameter checking in a different method, find_one and get should contain only business logic. Also, why not name the query parameter "query"?

mugurrus:
:+1:  Aside from my comment everything else seems OK

server/publicapi/items/service.py
plamut: I agree with factoring the parameter checking out into a separate method, it's already planned and will be done when we start adding semantic checking as well (i.e. checking the values and the format of the parameters received).

Re: "query" - I don't know, ask @hlmnrmr, that's what it's called in the API specification. :smile: I would support changing it to a better (read: more explicit) name.

PR #549, opened 2015-05-11
Initial support for takes packaging

amagdas:
Please ignore all the logging, it's going to be removed before creating the PR to master.


mdhaman:
:+1:

akintolga:
:+1:

server/apps/archive/archive_composite.py
```diff
@@ -61,6 +63,53 @@ def get_item_ref(item):
 
 class PackageService():
```
sivakuna-aap: This should extend from `BaseService`

amagdas: Nop, it's not a resource service, more like a helper service. You can't get it through `get_resource_service`

PR #558, opened 2015-05-12
UI changes for send and continue.

mdhaman:


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -293,6 +293,25 @@
             });
             item._locked = true;
         };
+
+        /**
+        * Link an item for takes.
+        * @param {Object} item : Target Item
+        * @param {string} link_id: Optional. If not provider it returns the new Linked item.
+        * @param {string} desk: Optional. Desk for newly create item.
```
plamut: Tip: Parameters can also be [marked as optional](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values) by wrapping them into angular brackets, and then the docs generator takes care of the `optional`notion.

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1070,6 +1077,25 @@
...
+                };
+
+                scope.sendAndContinue = function() {
+                    var deskId = scope.selectedDesk._id;
+                    var stageId = scope.selectedStage._id || scope.selectedDesk.incoming_stage;
+                    var activeDeskId = desks.activeDeskId;
+                    scope.item.more_coming = true;
+                    return sendAuthoring(deskId, stageId, scope.selectedMacro, true)
+                        .then(function() {
+                            authoring.linkItem(scope.item, null, activeDeskId).then(function (item) {
```
plamut: IMO, flattening the promise chain is a bit more readable than having nested `.then()` blocks.

server/apps/archive/archive_link.py
```diff
@@ -24,8 +25,8 @@ class ArchiveLinkResource(Resource):
     resource_title = endpoint_name
 
     schema = {
-        'link_id': Resource.rel('archive', False, required=False),
-        'desk': Resource.rel('desks', required=False)
+        'link_id': Resource.rel('archive', False, type='string'),
+        'desk': Resource.rel('desks', False)
```
plamut: It might be a good idea to retain the `required` argument keyword so that's it's immediately clear what that `False` argument stands for.

akintolga:
:+1:

amagdas:
:+1:

PR #559, opened 2015-05-12
test(coverage): report test coverage to coveralls

petrjasek:
it does client and server unittests together and reports the coverage to coveralls


plamut:
This looks good in general, I like it! Though It seems to me that it is to eager on a couple of things? For example here: https://landscape.io/diff/150790

We are not using any old-style classes, right? And the complaints about unused variables are false positives IMO, those functions are registered as error handlers, but the linter does not recognize that.

I think we should ajdust the linter settings a bit before merging (there could be other false positives as well, but didn't manage to check it yet).

plamut:
I see that you disabled some python2-specific warnings, that's good. :+1:

petrjasek:
yep, also until we tweak the settings I've disabled those comments

plamut:
Yes, the only relevant info for now should be just the test coverage anyway. The rest is just a bonus and should not make the builds appear erroneous, especially not before its behaviour is in sync with `flake8` we currently use for code quality checks.

amagdas:
:+1:

PR #561, opened 2015-05-12
Fix (word count): Fixed word and character count on creating preform…

Aca-jov:
…ated item

SD-1907

There was also problem with character count, not included in this issue, which is fixed now.

Also there is one more problem: On creating preformatted item if you type something in body and press enter, body loses focus. If you press enter and continue typing it works good, but if only press enter, focus is lost. I tried to find problem and found in:
..\app\scripts\bower_components\medium-editor\dist\js\medium-editor.js
function handleKeyup: `if (event.which === Util.keyCode.ENTER && !Util.isListItem(node)) {`

This is where on enter it loses focus, but don't know how to handle since it is in bower and probably used in other places, so if someone can take a look and solve/create issue if needed..

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -25,8 +25,7 @@
         description: null,
         destination_groups: null,
         sign_off: null,
-        publish_schedule: null,
-        pubstatus: null
+        publish_schedule: null
     };
 
     /**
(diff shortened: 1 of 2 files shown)


plamut:
If it is in the `bower_components` directory, it means that this is an external component that we pull into our project. If you believe there is a bug, you should probably report a bug to the developers of the [medium-editor](https://github.com/daviferreira/medium-editor).

In any case we should not attempt to fix this directly, it needs to be fixed in the component itself. If we have the time and resources, we can even try to fix it ourselves, the maintainers of the component will surely appreciate it.

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -849,7 +842,8 @@
     }
 
     var cleanHtml = function(data) {
-        return data.replace(/<\/?[^>]+>/gi, '').replace('&nbsp;', ' ');
+        return data.replace(/<br>/g, '&nbsp;').replace(/<\/?[^>]+><\/?[^>]+>/gi, ' ')
```
plamut: What about e.g. `<br/>` or `<br />`(note the space)? Could these also appear in data?

ioanpocol:
:+1:

mugurrus:
:+1:

PR #564, opened 2015-05-13
removing the html tags for headline before save.

mdhaman:


amagdas:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -40,6 +40,16 @@
         return angular.extend(dest, _.pick(src, _.keys(CONTENT_FIELDS_DEFAULTS)));
     }
 
+    function stripHtml(item) {
```
petrjasek: fyi it would be also possible via dom apis, like:
```
var elem = document.createElement('div');
elem.innerHTML = item.headline;
return elem.textContent;
```

mdhaman: yes. will change that.

PR #566, opened 2015-05-13
[SD-2166] Validate mandatory fields before publishing

marwoodandrew:
The validators will need to be loaded after or as part of the deployment


amagdas:
:+1:  :heart: :blue_heart: :yellow_heart:

mdhaman:
:+1:

PR #571, opened 2015-05-13
[SD-2166] add validator for corrections and fix subjects

marwoodandrew:


amagdas:
:+1:

server/apps/prepopulate/data_initialization/validators.json
```diff
@@ -41,10 +41,16 @@
         },
         "subject" : {
             "type" : "list",
-            "required" : true
+            "required" : true,
+            "minlength" : 1
```
plamut: Doesn't the `required` flag already imply a length of at least 1 character? Is this perhaps some validator quirk and should thus be addressed there instead of circumventing it in config?

Just want to make sure, thanks.

akintolga: I think it is still needed. With required empty strings are validated. 
https://cerberus.readthedocs.org/en/latest/#required

plamut: OK, if this is the way the validator is designed, then it's fine of course and my comment can be ignored (thanks for the link BTW).

:+1:

akintolga: No worries Peter, thanks for pointing out. Cheers.

plamut: Better safe than sorry. ;)

marwoodandrew: Yep, with just the required : true it allows empty lists, need to minlength to ensure that there is at least one entry

PR #575, opened 2015-05-13
(mp) SD-2219 publish queue page improvements

sivakuna-aap:


client/app/scripts/superdesk-publish/publish.js
```diff
@@ -145,12 +145,21 @@
...
         $scope.publish_queue = null;
 
+        $scope.selectedFilterChannel = null;
+        $scope.selectedFilterSubscriber = null;
+
+        $scope.multiSelectCount = 0;
+        $scope.selectedQueueItems = [];
+
+        $scope.showResendBtn = false;
+        $scope.showCacnelBtn = false;
```
amagdas: Spelling error, `cancel`

amagdas:
:+1:

client/app/scripts/superdesk-publish/publish.js
```diff
@@ -205,8 +229,21 @@
...
             newItem.unique_name = item.unique_name;
             newItem.destination = item.destination;
+            return newItem;
+        };
+
+        $scope.scheduleToSend = function(item) {
+            var queueItems;
+
+            if (angular.isDefined(item)) {
+                queueItems = $scope.buildNewSchedule(item);
```
akintolga: Here maybe we don't need to build a new object. We can pass the same item after setting its _id=null.

client/app/scripts/superdesk-publish/publish.js
```diff
@@ -223,20 +260,91 @@
...
+
+            if (angular.isDefined(item)) {
+                api.publish_queue.save(item, _updates).then(
+                    function(response) {
+                    },
+                    function(response) {
+                        if (angular.isDefined(response.data._issues)) {
+                            if (angular.isDefined(response.data._issues['validator exception'])) {
+                                notify.error(gettext('Error: ' + response.data._issues['validator exception']));
+                            }
```
mugurrus: Shouldn't it display a notification when there are other issues than 'validator exception'?

mugurrus:
:+1:

client/app/scripts/superdesk-publish/publish.js
sivakuna-aap: True. But those are raised by cerberus based on the validators declared in schema definition. Since `publish_queue` doesn't have validators declared at schema level, I didn't raise notifications.

client/app/scripts/superdesk-publish/publish.js
sivakuna-aap: There are some properties which become invalid in case of post like error_message,...in future there can be some others, that's the reason am building a new object.

client/app/scripts/superdesk-publish/publish.js
```diff
@@ -167,35 +176,50 @@
...
+
+                    $scope.showResendBtn = false;
+                    $scope.showCacnelBtn = false;
                 });
             });
         };
 
-        $scope.scheduleToSend = function(item) {
+        $scope.buildNewSchedule = function (item) {
             var newItem = {};
```
mdhaman: maybe we do _.omit(angular.extend({}, item), 'published_seq_num') instead of buildNewSchedule

sivakuna-aap: There are other properties other than `published_seq_num` like error_message,...in future there can be some others, that's the reason am building a new object.

akintolga:
:+1:

PR #577, opened 2015-05-14
[SD-1988] Closing output channels on error

akintolga:


client/app/scripts/superdesk-publish/views/output-channels.html
```diff
@@ -29,73 +29,93 @@ <h3 translate ng-hide="outputChannel._id" translate>Add New Output Channel</h3>
...
+                    </tab>
+                    
+
+                    <tab heading="Critical Errors" class="TabContent">
+                        <div class="content privileges-settings">
+                            <div class="horizontal">
+                                <div class="roles-list">
+                                    <div ng-repeat="(key, val) in all_errors | orderBy:name" class="field">
+                                        <label>
+                                            {{key}}: {{val}}
```
plamut: Are these two values fixed (i.e. remain unchanged)? If so, we should use one-time bindings for them.

client/app/scripts/superdesk-publish/views/subscribers.html
```diff
@@ -29,37 +29,56 @@ <h3 translate ng-hide="subscriber._id" translate>Add New Subscriber</h3>
...
+                        </div>
+                    </tab>
+
+                    <tab heading="Critical Errors" class="TabContent">
+                        <div class="content privileges-settings">
+                            <div class="horizontal">
+                                <div class="roles-list">
+                                    <div ng-repeat="(key, val) in all_errors | orderBy:name" class="field">
+                                        <label>
+                                            {{key}}: {{val}}
```
plamut: Same comments as above (i.e. possible to use one-time bindings?).

client/app/scripts/superdesk-publish/views/output-channels.html
akintolga: You are quite right Peter. I'm fixing those. Cheers.

sivakuna-aap:
:+1:

plamut:
:+1:

PR #578, opened 2015-05-15
fix(prepopulate) allow updates for some entities and don't crash if t…

marwoodandrew:
…he data is already there


amagdas:
:+1:

sivakuna-aap:
:+1:

PR #579, opened 2015-05-15
Added Published item expiry and removed published stage.

mdhaman:


sivakuna-aap:
why are removing published stage?

amagdas:
@sivakuna-aap It's not needed anymore as we don't show it, we show instead the Published collection.

amagdas:
:+1:

server/apps/publish/archive_publish.py
```diff
@@ -142,12 +139,8 @@ def update(self, id, updates, original):
             raise SuperdeskApiError.internalError(message="Failed to publish the item: {}"
                                                   .format(str(e)))
 
-    def publish(self, doc, updates, target_output_channels=None):
-        any_channel_closed = self.queue_transmission(doc=doc, target_output_channels=target_output_channels)
-        task = self.__send_to_publish_stage(doc)
-        if task:
-            updates['task'] = task
-        return any_channel_closed
+    def publish(self, doc, target_output_channels=None):
```
sivakuna-aap: Since this is just invoking `queue_transmission` can we remove this method and directly invoke `queue_transmission` above?

sivakuna-aap:
thank you @amagdas

server/apps/publish/archive_publish.py
mdhaman: will do that.

mdhaman:
Thanks @amagdas.

akintolga:
:+1:

PR #582, opened 2015-05-15
Content API: Allow filtering items/packages by creation date

plamut:
Also did refactored parameter parsing out into helper methods to keep the `ItemsService.get()` method clean and readable.

Resolves [SD-2252](https://dev.sourcefabric.org/browse/SD-2252)

### DIFF
--- a/server/publicapi/features/prepopulate.feature
+++ b/server/publicapi/features/prepopulate.feature
@@ -7,7 +7,7 @@ Feature: Prepopulate public content
     	"""
         Then we get response code 201
 
-        When we get "/items"
+        When we get "/items?start_date=1970-01-01"
         Then we get list with 1 items
         """
         {
@@ -20,7 +20,7 @@ Feature: Prepopulate public content
         }
         """
 
-        When we get "/packages"
+        When we get "/packages?start_date=1970-01-01"
         Then we get list with 1 items
         """
         {
--- a/server/publicapi/items/service.py
+++ b/server/publicapi/items/service.py
@@ -11,6 +11,7 @@
 import json
 import logging
 
+from datetime import datetime
 from eve.utils import ParsedRequest
 from flask import current_app as app
 from publicapi.errors import BadParameterValueError, UnexpectedParameterError
@@ -29,40 +30,6 @@ class ItemsService(BaseService):
 
     Serves mainly as a proxy to the data layer.
     """
-
-    def _check_request_params(self, request, whitelist, allow_filtering=True):
-        """Check if the request contains only valid parameters.
-
-        :param req: object representing the HTTP request
-        :type req: `eve.utils.ParsedRequest`
-        :param whitelist: iterable containing the names of allowed parameters.
-        :param bool allow_filtering: whether or not the filtering parameter is
-            allowed (True by default). Used for disallowing it when retrieving
-            a single object.
-
-        :raises BadParameterValueError: if the request contains more than one
-            value for any of the parameters
-        :raises UnexpectedParameterError: if the request contains a parameter
-            that is not whitelisted
-        """
-        request_params = (request.args or MultiDict())
-
-        if not allow_filtering and 'q' in request_params.keys():
-            desc = (
-                "Filtering is not supported when retrieving a single "
-                "object (the \"q\" parameter)"
-            )
-            raise UnexpectedParameterError(desc=desc)
-
-        for param_name in request_params.keys():
-            if param_name not in whitelist:
-                raise UnexpectedParameterError(
-                    desc="Unexpected parameter ({})".format(param_name)
-                )
-
-            if len(request_params.getlist(param_name)) > 1:
-                desc = "Multiple values received for parameter ({})"
-                raise BadParameterValueError(desc=desc.format(param_name))
 
     def find_one(self, req, **lookup):
         """Retrieve a specific item.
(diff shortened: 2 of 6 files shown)


mugurrus:
It's good to consolidate commits

plamut:
Do you mean here in this case or as a general rule?

IMO there shouldn't be just a single giant commit that contains everything, it's better to have more granular commits, each containing a rounded logical piece of functionality. I see two main advantages with this approach:
* The commits themselves more clearly explain, what pieces of (sub)functionality form the actual "big" feature, as they provide a better insight into developer's thought process.
* In case a bug is discovered, it is much easier to locate the exact piece of code that caused it (e.g. by using `git bisect`). That would be more difficult if a bug happened to appear in a large monolithic commit.

Let me know your thoughts on this (others welcome to comment, too, of course).

mugurrus:
I'm not suggesting a single monolithic commit but filling the history with many small commits "dilutes" the feature.

plamut:
Well, some features are (speaking in general here) more complex and by their very nature require more commits by default, assuming we do not want the commits to be too big and thus less readable. It's a compromise between having fewer commits and their readability.

Do you have any concrete suggestions on where to draw the line in cases like these? If history is a concern, we could perhaps change the format of the commit messages. Putting the issue number _before_ the commit message would already make it immediately clear which commits together form a particular feature.

mugurrus:
:+1:

server/publicapi/items/service.py
```diff
@@ -67,19 +62,169 @@ def get(self, req, lookup):
...
+            raise BadParameterValueError(
+                desc="Invalid parameter value (end_date)") from None
+
+        # disallow dates in the future
+        future_err_msg = (
+            "{} date ({}) must not be set in the future "
+            "(current server date: {})")
+        today = date.today()
+
+        if (start_date is not None) and (start_date > today):
```
ioanpocol: For the case when items for current day is requested, you have to add here start_date >= today.

server/publicapi/items/service.py
```diff
@@ -67,19 +62,169 @@ def get(self, req, lookup):
...
+            raise BadParameterValueError(
+                desc=future_err_msg.format(
+                    'Start', start_date.isoformat(), today.isoformat()
+                )
+            )
+
+        if (end_date is not None) and (end_date > today):
```
ioanpocol: For the case when items for current day is requested, you have to add here end_date <= today.

server/publicapi/features/prepopulate.feature
```diff
@@ -7,7 +7,7 @@ Feature: Prepopulate public content
     	"""
```
ioanpocol: More behave tests are needed.

server/publicapi/items/service.py
```diff
@@ -67,19 +62,169 @@ def get(self, req, lookup):
...
+            end_date = self._parse_iso_date(request_params.get('end_date'))
+        except ValueError:
+            raise BadParameterValueError(
+                desc="Invalid parameter value (end_date)") from None
+
+        # disallow dates in the future
+        future_err_msg = (
+            "{} date ({}) must not be set in the future "
+            "(current server date: {})")
+        today = date.today()
```
ioanpocol: So filter parameters should be on server local zone.

server/publicapi/items/service.py
```diff
@@ -67,19 +62,169 @@ def get(self, req, lookup):
...
+        :type end_date: datetime.date or None
+
+        :return: MongoDB date range filter (as a dictionary)
+        """
+        if (start_date is None) and (end_date is None):
+            return {}  # nothing to set for the date range filter
+
+        date_filter = {'versioncreated': {}}
+
+        if start_date is not None:
```
ioanpocol: If start_date and end_date are equal, the query will not return any response, since version created is a datetime.

server/publicapi/features/prepopulate.feature
```diff
@@ -7,7 +7,7 @@ Feature: Prepopulate public content
...
         Then we get response code 201
 
-        When we get "/items"
+        When we get "/items?start_date=1970-01-01"
```
ioanpocol: How can perform a client a pooling for new items added on current day?
If we use this day granularity function we will get at every request all items from the current day.
We have to use a datetime or better a change counter.

server/publicapi/items/service.py
plamut: I assume yes, but it's true that this was never specified exactly. I'm thinking it would probably be better to use the current UTC time, what do you think?

server/publicapi/items/service.py
```diff
@@ -67,19 +62,169 @@ def get(self, req, lookup):
...
+        :param date_str:
+        :type date_str: str or None
+
+        :return: resulting date object or None if None is given
+        :rtype: datetime.date
+        """
+        if date_str is None:
+            return None
+        else:
+            return datetime.strptime(date_str, '%Y-%m-%d').date()
```
ioanpocol: If the client sends also hours and minutes for date, the hours are reset but the user don't receive any error.

server/publicapi/items/service.py
ioanpocol: Yes, UTC time will eliminate some problems.

server/publicapi/items/service.py
plamut: Actually it does - if the time part is sent as well, `strptime()` raises ValueError. This is BTW intented, clients should only be allowed to specify dates (as per specification).

server/publicapi/features/prepopulate.feature
plamut: I agree. The thing is realized that we need to improve our behave testing data set if we want to write good tests, and felt it is kind of beyond the scope of this ticket. Basically the `populate` would need to be changed so that it initializes database with data containing more variation in it.

server/publicapi/items/service.py
plamut: Oh OK, good catch, forgot about that detail (I initally had datetimes but converted them to dates after discussing it with Holman).
I'll change the way filter is created if the start and end dates are equal.

server/publicapi/items/service.py
ioanpocol: Yes, strptime() raises error if excess data after parsing

server/publicapi/features/prepopulate.feature
plamut: As per the current specification, this is not supported, but I agree it would make sense. The reason we opted for dates only was that many APIs from the major news companies behave the same.

@hlmnrmr what do you think about it? When we were discussing this, we didn't know the exact reason for that, we just followed their support for dates only.

server/publicapi/items/service.py
plamut: Hm, what exactly do you mean here? `end_date <= today` is a valid value - or am I missing something?

ioanpocol: When you request items from current day is it true the following relation: 
 start_date = today = end_date ?

plamut:
Thanks for a lot of useful feedback, @ioanpocol!

To sum up, here are the changes that should be made:
* fix the edge case when start_date == end_date (querying today's items)
* when determining today's date, make sure to use UTC timezone (and not server's local timezone), because internally all datetimes are stored in UTC.
* update the `_parse_iso_date` method's docstring to explain in what cases a `ValueError` is raised
* behave tests ... this should probably be addressed in a separate ticket, because it requires some refactoring of the current behave setup for public API?

server/publicapi/items/service.py
plamut: Yes, that's the way querying today's items is intended to work.

It doesn't work yet though, as you pointed out, I'll add an additional check to cover this edge case. This will probably resolve this concern here too, right? (and the related one a few lines above)

server/publicapi/items/service.py
hlmnrmr: Right, system generated dates should be represented as UTC

ioanpocol:
:+1:

plamut:
Updated, please review.

plamut:
Yes, this is indeed the requirement (described [here](https://dev.sourcefabric.org/browse/SD-2252)). If `start_date` is not given, it should default to "`end_date` minus one day", returning the items created in the last 24-48 hours (both date range limits are inclusive BTW).

The `_get_date_range()` helper method deals only with parsing and validating the date range parameters (and setting their default values if not given). The logic dealing with translating the given date range to the actual DB filter is fully contained in the `_create_date_range_filter()` method.

Re: behave tests - unit tests would easily catch this kind of error, too. :wink: 
(I suggest adding behave tests under a separate ticket, because refactoring of the whole database pre-populating is required for that)

plamut:
As per the updated specification, the default value for `start_date` is now changed to be simply the `end_date`.

ioanpocol:
:+1:

PR #583, opened 2015-05-16
(feat) [SD-2250] Assign a default Source to user created content Items

sivakuna-aap:


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -91,6 +91,11 @@
             this.stop(item);
             timeouts[item._id] = $timeout(function() {
                 var diff = extendItem({_id: item._id}, item);
+
+                if (diff.publish_schedule === null) {
```
amagdas: Isn't this affecting scheduled publishing?

server/settings.py
```diff
@@ -269,3 +269,6 @@ def env(variable, fallback_value=None):
 
 # Defines the maximum value of Publish Sequence Number after which the value will start from 1
 MAX_VALUE_OF_PUBLISH_SEQUENCE = 9999
+
+# Defines default value for Source to be set for manually created articles
+DEFAULT_SOURCE_VALUE_FOR_MANUAL_ARTICLES = 'AAP'
```
amagdas: I think we should read this from system env so we can deploy for other envs without code changes.

amagdas:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
sivakuna-aap: No this won't affect. Without this autosave was throwing exceptions '`publish_schedule` is unknown field'. When I added in the schema it was throwing exception that null is not allowed in `publish_schedule`. That's the reason I had to add this check.

However, there is one more issue with published date for which I will raise JIRA.

amagdas: :+1:

akintolga:
:+1:

PR #585, opened 2015-05-18
feat(add takes) - refactor TakesPackageService

amagdas:
- mark the package as "takes_package" only when needed


server/apps/archive/takes_package_service.py
```diff
@@ -0,0 +1,119 @@
+# -*- coding: utf-8; -*-
```
sivakuna-aap: Since packages functionality is growing I think it's better to refactor to a different package.

amagdas: You mean to move the archive_composite.py and takes_package_service.py to another folder module?

sivakuna-aap: Yes, we can move other modules also if there are any related to package into one folder module.

amagdas: Sounds good, will do.

akintolga:
:+1:

mdhaman:
:+1:

PR #594, opened 2015-05-18
feat(content): Added creating content item from template.

eozan:
SD-2236

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1203,22 +1203,23 @@
         };
     }
 
-    ContentCreateDirective.$inject = ['templatesSettingsService'];
-    function ContentCreateDirective(templatesSettingsService) {
+    ContentCreateDirective.$inject = ['api'];
+    function ContentCreateDirective(api) {
+        var NUM_ITEMS = 5;
         return {
             templateUrl: 'scripts/superdesk-authoring/views/sd-content-create.html',
             link: function(scope) {
                 scope.contentTemplates = null;
 
-                var fetchTemplates = function() {
-                    templatesSettingsService.fetchContentTemplates()
-                    .then(function(contentTemplates) {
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1203,9 +1203,23 @@
...
-    function ContentCreateDirective() {
+    ContentCreateDirective.$inject = ['templatesSettingsService'];
+    function ContentCreateDirective(templatesSettingsService) {
         return {
-            templateUrl: 'scripts/superdesk-authoring/views/sd-content-create.html'
+            templateUrl: 'scripts/superdesk-authoring/views/sd-content-create.html',
+            link: function(scope) {
+                scope.contentTemplates = null;
+
+                var fetchTemplates = function() {
```
plamut: We can get rid of this private function here, as its only purpose is to be invoked once when the directive is linked. Just execute the function's body and that's it. :wink:

client/app/scripts/superdesk-workspace/content/content.js
```diff
@@ -4,6 +4,20 @@
 ContentCtrlFactory.$inject = ['api', 'superdesk'];
 function ContentCtrlFactory(api, superdesk) {
     return function ContentCtrl($scope) {
```
plamut: `$scope` is not used anywhere, we can remove it as a dependency.

plamut:
Apart from a few minor remarks it seems good. But please add unit tests for the new code before merging.

ioanpocol:
An e2e tests will help.

ioanpocol:
:+1:

plamut:
IMO e2e tests should not be viewed as a replacement for unit tests. e2e tests are generally slow to run and it's not feasible tu run them after every code change (i.e. when you are about to commit something).

eozan:
unit tests added.

plamut:
That's good!

:+1:

amagdas:
:+1:

PR #595, opened 2015-05-18
(imp) [SD-2318] Minor fixes for publish queue display:

sivakuna-aap:
1. Timestamps should be resolved/displayed to seconds.
2. Items are not being displayed in sequence number order

### DIFF
--- a/client/app/scripts/superdesk-publish/publish.js
+++ b/client/app/scripts/superdesk-publish/publish.js
@@ -184,7 +184,7 @@
         function fetchPublishQueue () {
             var criteria = criteria || {};
             criteria.max_results = 200;
-            criteria.sort = "[('published_seq_num',-1)]";
+            criteria.sort = '[(\'published_seq_num\',-1)]';
 
             if ($scope.selectedFilterSubscriber !== null && $scope.selectedFilterChannel !== null) {
                 criteria.where = JSON.stringify({
--- a/client/app/scripts/superdesk-publish/publish.js
+++ b/client/app/scripts/superdesk-publish/publish.js
@@ -184,7 +184,7 @@
         function fetchPublishQueue () {
             var criteria = criteria || {};
             criteria.max_results = 200;
-            criteria.sort = '[(\'_created\',-1)]';
+            criteria.sort = '[(\'published_seq_num\',-1)]';
 
             if ($scope.selectedFilterSubscriber !== null && $scope.selectedFilterChannel !== null) {
                 criteria.where = JSON.stringify({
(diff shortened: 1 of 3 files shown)


client/app/scripts/superdesk-publish/publish.js
```diff
@@ -184,7 +184,7 @@
         function fetchPublishQueue () {
             var criteria = criteria || {};
             criteria.max_results = 200;
-            criteria.sort = '[(\'_created\',-1)]';
+            criteria.sort = '[(\'published_seq_num\',-1)]';
```
plamut: Perhaps just use double quotes to avoid the need for escaping, makes it a bit easier on the eyes, but it's a minor thing.

client/app/scripts/superdesk-publish/views/publish-queue.html
```diff
@@ -62,9 +62,9 @@ <h3 class="page-nav-title" translate>Publish Queue</h3>
 				<td>{{subscriberLookup[queue_item.subscriber_id].name}}</td>
 				<td ng-if="queue_item.selector_codes != null && queue_item.selector_codes.length > 0">{{queue_item.destination.name}} {{ queue_item.selector_codes }}</td>
                 <td ng-if="queue_item.selector_codes == null || queue_item.selector_codes.length == 0">{{queue_item.destination.name}}</td>
-				<td>{{queue_item._created | dateTimeString}}</td>
-				<td>{{queue_item.publish_schedule | dateTimeString}}</td>
-				<td>{{queue_item.completed_at | dateTimeString}}</td>
+				<td>{{queue_item._created | dateTimeStringWithMillis}}</td>
+				<td>{{queue_item.publish_schedule | dateTimeStringWithMillis}}</td>
+				<td>{{queue_item.completed_at | dateTimeStringWithMillis}}</td>
```
plamut: Was that meant to be `dateTimeStringWithSecs`?

sivakuna-aap: @plamut you were right, thanks for pointing it out.

client/app/scripts/superdesk-publish/publish.js
sivakuna-aap: This isn't possible as jshint forces to use single quotes

akintolga:
:+1:

amagdas:
:+1:

client/app/scripts/superdesk-publish/publish.js
plamut: What about replacing the internal single quotes that are now escaped? Would that work or would some other component complain that it cannot parse the sort criterion due to the double quotes? Just an idea ...

plamut:
:+1: 
(my other remark is insignificant)

PR #600, opened 2015-05-19
feat(templates): add save as template action

petrjasek:


client/app/scripts/superdesk-authoring/views/authoring-topbar.html
```diff
@@ -42,6 +42,19 @@
             ng-show="_editable"
             translate>Publish</button>
 
+        <div class="navbtn dropdown pull-left" dropdown>
+            <button class="dropdown-toggle" dropdown-toggle><i class="icon-dots-vertical"></i></button>
+            <div class="dropdown-menu">
+                <ul>
+                    <li ng-controller="TemplateActionsController as template">
```
plamut: The name "template" for a controller could be a bit misleading, perhaps think of another name...but it's a minor thing.

client/app/scripts/superdesk-templates/templates.js
```diff
@@ -116,37 +126,114 @@
...
+        this.type = 'create';
+        this.name = item.slugline || null;
+        this.desk = desks.active.desk || null;
+
+        this.types = templates.types;
+        this.save = save;
+
+        activate();
+
+        function activate() {
```
plamut: Since this `activate()` function is invoked only once at controller instantiation and is not visible to the outside world, it might make sense to get rid of it and simply only retain its body?

plamut:
:+1: 

Apart for two minor remarks it looks good to me.

amagdas:
:+1:  Rebase needed.

client/app/scripts/superdesk-templates/templates.js
petrjasek: imo this makes things a bit more explicit

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
petrjasek: thing is I couldn't think of a better name :smile: any ideas?

client/app/scripts/superdesk-templates/templates.js
plamut: That's true, yes. The same could also be achieved with a comment, but it's more a matter of style than anything else.

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
plamut: `templateCtrl` - ugly and a bit lengthy, but probably still good enough? :D

petrjasek: but why is this important to know it's a controller and not something else? and what else could it be anyhow? ;)

plamut: Well, there's a lot of code elsewhere in the application that doesn't use `controllerAs` syntax and places everything into `$scope` instead. It's just about being more explicit in the template about where things come from, but as I said, it's not that big of a deal. :)

petrjasek: I will be working on this menu a bit more for next feature, so will make some changes there. thx

PR #601, opened 2015-05-19
Remove the symbolic link to bower.json. It works now on windows

vied12:
Because @darconny is still using windows :trollface:, we have to keep in mind that symblink doesn’t work on it.
Then I reversed the link and the file to have the real file at the root of the project. Bower won't fail any more on Windows for that.


plamut:
It seems that this broke the build, `cd client` command fails. :wink: Travis settings need to be updated as well.

plamut:
I think we can consider this build as passed, it's just the coverage that got stuck.

amagdas:
:+1:

plamut:
What's the status of this? Has the symbolic link been re-introduced?
(read something about that somewhere, I think...)

petrjasek:
:+1:

ioanpocol:
A rebase is needed

vied12:
Updated

plamut:
:+1:

PR #607, opened 2015-05-20
[SD-2327] - Urgency and body are added for the validation

akintolga:
- Body needs to be in "text" item validation which needs to be separated from other formats in later sprints
- There are improvements in error handling as well.


sivakuna-aap:
:+1:

PR #609, opened 2015-05-20
[SD-2327] - Urgency and body are added for the validation

akintolga:


amagdas:
:+1:

marwoodandrew:
Should we check for the allowed values in the urgency? 
Other than that :+1:

mdhaman:
we should check for allowed values.

PR #614, opened 2015-05-20
feat(spellcheck): let user turn spellcheck on and off

petrjasek:
there is a new submenu where you can toggle it on/off and in case
it's off you can run it from there too.


ioanpocol:
:+1:

amagdas:
:+1:

PR #615, opened 2015-05-20
Public API: return more descriptive msg when date range filtering is disallowed

plamut:
Resolves [SD-2330](https://dev.sourcefabric.org/browse/SD-2330).


ioanpocol:
:+1:

mugurrus:
:+1:

PR #616, opened 2015-05-20
feat(iptc search): Integrate structure on search panel

pavlovicnemanja92:
SD-2254

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/metadata.js
+++ b/client/app/scripts/superdesk-authoring/metadata/metadata.js
@@ -272,18 +272,18 @@ function MetadataService(api, $q, staticMetadata) {
             }
         },
         removeSubjectTerm: function(term) {
-            var o = {},
-                self = this,
+            var self = this,
+                tempItem = {},
                 subjectCodesArray = self.subjectScope.item[self.subjectScope.field],
-                temp = _.without(subjectCodesArray, term);
+                filteredArray = _.without(subjectCodesArray, term);
 
-            if (temp.length === subjectCodesArray.length) {
-                _.remove(temp, {name: term});
+            if (filteredArray.length === subjectCodesArray.length) {
+                _.remove(filteredArray, {name: term});
             }
 
-            o[self.subjectScope.field] = temp;
+            tempItem[self.subjectScope.field] = filteredArray;
 
-            _.extend(self.subjectScope.item, o);
+            _.extend(self.subjectScope.item, tempItem);
 
             self.subjectScope.change({item: self.subjectScope.item});
         },
--- a/client/app/scripts/superdesk-authoring/metadata/metadata.js
+++ b/client/app/scripts/superdesk-authoring/metadata/metadata.js
@@ -144,8 +144,9 @@ function MetadataWordsListEditingDirective() {
  * @param {String} unique - specify the name of the field, in list item which is unique (qcode, value...)
  *
  */
-MetadataListEditingDirective.$inject = [];
-function MetadataListEditingDirective() {
+var subjectScope;
+MetadataListEditingDirective.$inject = ['metadata'];
+function MetadataListEditingDirective(metadata) {
     return {
         scope: {
             item: '=',
@@ -158,6 +159,8 @@ function MetadataListEditingDirective() {
         },
         templateUrl: 'scripts/superdesk-authoring/metadata/views/metadata-terms.html',
         link: function(scope) {
+
+            subjectScope = scope;
 
             scope.$watch('list', function(items) {
                 if (!items || !items[0].hasOwnProperty('parent')) {
@@ -226,15 +229,7 @@ function MetadataListEditingDirective() {
             };
 
             scope.removeTerm = function(term) {
-                var temp = _.without(scope.item[scope.field], term);
-
-                //build object
-                var o = {};
-                o[scope.field] = temp;
-
-                _.extend(scope.item, o);
-
-                scope.change({item: scope.item});
+                metadata.removeSubjectTerm(term);
             };
         }
     };
@@ -274,6 +269,30 @@ function MetadataService(api, $q, staticMetadata) {
                 return api('archive').getById(_id).then(function(_item) {
                     return _item;
                 });
+            }
+        },
+        removeSubjectTerm: function(term) {
+            var o = {},
+                subjectCodesArray = subjectScope.item[subjectScope.field],
+                temp = _.without(subjectCodesArray, term);
+
+            if (temp.length === subjectCodesArray.length) {
+                temp = removeSubjectCode(subjectCodesArray, term);
+            }
+
+            o[subjectScope.field] = temp;
+
+            _.extend(subjectScope.item, o);
+
+            subjectScope.change({item: subjectScope.item});
+
+            function removeSubjectCode(subjectcodes, term) {
+                for (var i = 0, subjectCodesLenght = subjectcodes.length; i < subjectCodesLenght; i++) {
+                    if (subjectcodes[i].name === term) {
+                        subjectcodes.splice(i, 1);
+                        return subjectcodes;
+                    }
+                }
             }
         },
         initialize: function() {
--- a/client/app/scripts/superdesk-authoring/metadata/metadata.js
+++ b/client/app/scripts/superdesk-authoring/metadata/metadata.js
@@ -144,7 +144,6 @@ function MetadataWordsListEditingDirective() {
  * @param {String} unique - specify the name of the field, in list item which is unique (qcode, value...)
  *
  */
-var subjectScope;
 MetadataListEditingDirective.$inject = ['metadata'];
 function MetadataListEditingDirective(metadata) {
     return {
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -159,6 +160,8 @@ function MetadataListEditingDirective() {
         templateUrl: 'scripts/superdesk-authoring/metadata/views/metadata-terms.html',
         link: function(scope) {
 
+            subjectScope = scope;
```
plamut: `subjectScope` doesn't seem to be used anywhere in the directive. It's also polluting the global namespace.

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -276,6 +271,30 @@ function MetadataService(api, $q, staticMetadata) {
...
+                temp = removeSubjectCode(subjectCodesArray, term);
+            }
+
+            o[subjectScope.field] = temp;
+
+            _.extend(subjectScope.item, o);
+
+            subjectScope.change({item: subjectScope.item});
+
+            function removeSubjectCode(subjectcodes, term) {
```
plamut: This could probably be simplified by using LoDash's [remove()](https://lodash.com/docs#remove) function.

client/app/scripts/superdesk-search/search.js
```diff
@@ -1020,6 +1026,62 @@
...
+                            return filteredArray;
+                        }
+                        for (var i = 0, queryArrayLength = queryArray.length; i < queryArrayLength; i++) {
+                            var queryArrayElement = queryArray[i];
+                            if (queryArrayElement.indexOf('subject.name') !== -1) {
+                                var elementName = queryArrayElement.substring(
+                                        queryArrayElement.lastIndexOf('(') + 1,
+                                        queryArrayElement.lastIndexOf(')')
+                                        );
+                                for (var j = 0, subjectCodesLenght = scope.subjectcodes.length; j < subjectCodesLenght; j++) {
```
plamut: "lenght" is spelled "l-e-n-g-t-h" :wink: (seen it at a couple of places).

(don't need to change the PR just because of this, though, only consider changing it, if you need to apply some other changes as well)

client/app/scripts/superdesk-search/search.js
```diff
@@ -1020,6 +1026,62 @@
...
+                                                addItemSubjectName :
+                                                query + ' ' + addItemSubjectName);
+
+                                $location.search('q', q);
+                            } else {
+                                var params = $location.search();
+                                if (params.q) {
+                                    for (var j = 0; j < getSubjectCodes(currentTags).length; j++) {
+                                        if (item.subject.indexOf(getSubjectCodes(currentTags)[j]) === -1) {
+                                            var removeItemSubjectName = 'subject.name:(' + getSubjectCodes(currentTags)[j].name + ')';
```
plamut: `getSubjectCodes(currentTags)` is repeated multiple times, better to assign it to a variable at the beginning of the function, instead of repeatedly invoking the `getSubjectCodes()` function.

plamut:
You should also add tests and some JSDoc comments.

(it makes you popular with the girls, trust me)

client/app/scripts/superdesk-authoring/metadata/metadata.js
pavlovicnemanja92: We use subjectScope for defining object in variable, so we can use it on service.

client/app/scripts/superdesk-search/search.js
pavlovicnemanja92: This is a speed typing mistake :) I will change it :)

client/app/scripts/superdesk-authoring/metadata/metadata.js
plamut: Sorry, my bad, I overlooked it up there. Didn't expect it between the function declaration and its docstring.

petrjasek: this seems wrong, try to write a unittest for this service method..

client/app/scripts/superdesk-authoring/metadata/metadata.js
pavlovicnemanja92: Here we use _.without function... For second case, function removeSubjectCode first need to find object by name, because term variable from subject page is sent like string.

plamut: But LoDash allows exactly that, i.e. deleting objects from an array by the value of a property, like this:
```js
function removeSubjectCode(subjectcodes, term) {
    _.remove(subjectcodes, {name: term});
    return subjectcodes;
}
```
Much more readable and less error prone than a manual loop. :wink:

client/app/scripts/superdesk-search/search.js
```diff
@@ -1020,6 +1026,62 @@
                             }, 0, false);
                         });
                     });
+
+                    metadata.fetchSubjectcodes().then(function () {
+                        scope.subjectcodes = metadata.values.subjectcodes;
+                        tags.initSelectedFacets().then(function (currentTags) {
```
plamut: For better readability, a flat promise chain would be better than nesting `.then()` calls.

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -276,6 +271,27 @@ function MetadataService(api, $q, staticMetadata) {
...
             }
         },
+        removeSubjectTerm: function(term) {
+            var o = {},
+                self = this,
+                subjectCodesArray = self.subjectScope.item[self.subjectScope.field],
+                temp = _.without(subjectCodesArray, term);
+
+            if (temp.length === subjectCodesArray.length) {
+                temp = removeSubjectCode(subjectCodesArray, term);
```
plamut: Looking at this now, we could probably even get rid of the `removeSubjectCode()` function altogether and just replace it with the essential `_.remove(...)` line?
This will further increase the readability of the code and make it shorter.

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -276,6 +271,22 @@ function MetadataService(api, $q, staticMetadata) {
                 });
             }
         },
+        removeSubjectTerm: function(term) {
+            var o = {},
```
amagdas: Please use a more meaningful name for variables, single letter variable names are accepted only as indexes in `for` loops.

client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -47,18 +47,10 @@
...
-                    <input type="text" id="search-subject" ng-model="meta.subject.name" tabindex="11">
-                </div>
-            </div>
-            <div class="field">
-                <label for="search-slugline">Keyword</label>
-                <div class="control">
-                    <input type="text" id="search-slugline" ng-model="meta.slugline" tabindex="8">
-                </div>
+        <div class="field">
+            <label for="search-slugline">Keyword</label>
```
amagdas: Missing translate.

client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -78,7 +70,13 @@
             <input type="text" id="search-creator" ng-model="meta.original_creator" tabindex="12">
         </div>
 
-	</fieldset>
+        <div class="field">
+            <label for="search-subject">Subject</label>
```
amagdas: Translate

client/app/scripts/superdesk-search/search.js
```diff
@@ -1020,6 +1026,75 @@
...
+                     */
+                    function getSubjectCodes(currentTags) {
+                        var queryArray = currentTags.selectedParameters, filteredArray = [];
+                        if (!$location.search().q) {
+                            return filteredArray;
+                        }
+                        for (var i = 0, queryArrayLength = queryArray.length; i < queryArrayLength; i++) {
+                            var queryArrayElement = queryArray[i];
+                            if (queryArrayElement.indexOf('subject.name') !== -1) {
+                                var elementName = queryArrayElement.substring(
```
amagdas: This looks like duplicated logic, we should be able to extract it into the service class.

client/app/scripts/superdesk-authoring/metadata/metadata.js
pavlovicnemanja92: This is old variable, it was there :) But I will change it to something else ;)

plamut:
Looks much better now with all the updates. Just add the corresponding test(s), please (you know, for the popularity with the girls :wink: ), and that should be it.

ioanpocol:
:+1:

pavlovicnemanja92:
Can we merge this pull request and I will create new branch for e2e tests?

plamut:
Well, if it represents a blocker, then we probably can. AFAIK tests are not "officially" required on this project, they are only highly recommended (FWIW, this should be made more strict if you ask me :wink: ).

:+1:

PR #621, opened 2015-05-21
feat(content): Added template list modal.

eozan:
SD-2239

### DIFF
--- a/client/app/scripts/superdesk-authoring/views/sd-template-select.html
+++ b/client/app/scripts/superdesk-authoring/views/sd-template-select.html
@@ -1,0 +1,28 @@
+<div sd-modal data-model="open" class="output-channel-modal">
+    <div class="modal-header"> <a href="" class="close" ng-click="close()"><i class="icon-close-small"></i></a>
+        <h3 translate>Select template</h3>
+    </div>
+    <div class="modal-body">
+        <form>
+            <fieldset>
+                <div class="field">
+                    <label for="keyword" translate>Keyword</label>
+                    <input type="text" id="keyword" ng-model="options.keyword" ng-model-options="{ debounce: 300 }">
+                </div>
+                <div class="field">
+                    <label translate>Templates</label>
+                    <ul>
+                        <li ng-repeat="template in templates._items">
+                            <button class="btn btn-full" ng-click="select(template)">
+                                <span class="pull-left">{{ :: template.template_name }}</span>
+                            </button>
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-authoring/views/sd-template-select.html
```diff
@@ -0,0 +1,28 @@
...
+        <form>
+            <fieldset>
+                <div class="field">
+                    <label for="keyword" translate>Keyword</label>
+                    <input type="text" id="keyword" ng-model="options.keyword" ng-model-options="{ debounce: 300 }">
+                </div>
+                <div class="field">
+                    <label translate>Templates</label>
+                    <ul>
+                        <li ng-repeat="template in templates._items">
```
plamut: Highly recommended to add `track by` here, because the items list can change.

amagdas:
:+1:

client/app/scripts/superdesk-authoring/views/sd-template-select.html
amagdas: The number of templates in the list is going to be fairly small I think...

plamut: Well it's still a good practice. :)

plamut:
:+1: 

Approving, because the proposed `track by` is not essential here (but still recommended as a good practice).

client/app/scripts/superdesk-authoring/views/sd-template-select.html
petrjasek: imo here it would make sense because it will be filtering and re-rendering same templates multiple times

plamut: Doesn't it render only a button with the `template_name` here, and not the whole template?
(hence my other comment later that `track by` here is non-essential, only recommended)

petrjasek: it does. but my point is that here `templates._items` will change as you type into keyword input

plamut: Yes, we agree on this (that was BTW the exact reason for my initial comment).

petrjasek: yep, we agree ;)

eozan:
Added track by, merging.

PR #626, opened 2015-05-22
Server: dependencies defined in setup.py

vied12:
In order to allow this application to be used as a library, dependencies need to be defined with setuptool in `install_requires`.

Then I read `server/requirements.txt` and set the content (except urls) in `install_requires`.

Regarding urls, it would be nice at the end to use only packages registered on Pypi.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,10 +11,13 @@
 
 
 from setuptools import setup, find_packages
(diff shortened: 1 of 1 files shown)


plamut:
:+1:

amagdas:
:+1:

PR #627, opened 2015-05-23
Public API: Allow excluding content fields from the result ("projection")

plamut:
Resolves [SD-2253](https://dev.sourcefabric.org/browse/SD-2253).

### DIFF
--- a/server/publicapi/items/service.py
+++ b/server/publicapi/items/service.py
@@ -16,6 +16,7 @@ from datetime import datetime, timedelta
 from eve.utils import ParsedRequest
 from flask import current_app as app
 from publicapi.errors import BadParameterValueError, UnexpectedParameterError
+from publicapi.items import ItemsResource
 from superdesk.services import BaseService
 from superdesk.utc import utcnow
 from urllib.parse import urljoin, quote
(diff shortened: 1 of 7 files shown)


plamut:
[not directly related to the PR]
I noticed that for some reason the "user role view - edit user" e2e test failed, but after running the build again without changing anything, it passed. Links:
* [failed build](https://travis-ci.org/superdesk/superdesk/builds/63609928)
* [passed build](https://travis-ci.org/superdesk/superdesk/builds/63765578)

server/publicapi/items/service.py
```diff
@@ -228,6 +241,129 @@ def _create_date_range_filter(self, start_date, end_date):
 
         return date_filter
 
+    def _set_fields_filter(self, req):
```
petrjasek: just wondering why are all functions item service methods, seems like those could be defined outside and it would make it easier to reuse the code

petrjasek:
:+1:

server/publicapi/items/service.py
plamut: They are meant as helper methods to make the "public" methods more readable. An implementation detail basically, they are not designed to be general-purpose utility methods.

The reason is partially because the `PackagesService` already shares the code with the `ItemsService` via inheritance. On the other hand, if we had multiple independent services, then it would indeed be a good idea to put these helper methods (e.g. parameter parsing) into some common utility class.

amagdas:
:+1:

PR #628, opened 2015-05-25
[sd-2365]-(fix)-Destination group in Auth screen is enable even in re…

sjunaid:
…ad only mode plus having style issues.


amagdas:
:+1:

mdhaman:
:+1:

PR #637, opened 2015-05-26
fix(media icons): Icons appearing outside of items frame

pavlovicnemanja92:
SD-2343

### DIFF
--- a/client/app/scripts/superdesk-items-common/styles/media-archive.less
+++ b/client/app/scripts/superdesk-items-common/styles/media-archive.less
@@ -451,6 +451,7 @@
 			position: absolute;
 			height: 32px;
 			border-top: 1px solid rgba(0,0,0,0.1);
+                        font-size: 0;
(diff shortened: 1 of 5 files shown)


plamut:
An e2e test for the existence of the `copy` and the `duplicate` icons is [broken](https://travis-ci.org/superdesk/superdesk/jobs/64058629#L1154), it needs to be updated.

ioanpocol:
Running "jshint:all" (jshint) task

   app/scripts/superdesk-dashboard/workspace-tasks/tasks.js

     93 |    function fetch() {

                      ^ 'fetch' is defined but never used.

plamut:
Where does this come from, the `tasks.js` file was not modified in this particular PR?

ioanpocol:
There are a few PR with the same issue. Petr is working on it, so after it is fixed we can run the test again.

ioanpocol:
:+1:

PR #639, opened 2015-05-26
[SD-2382] - Analysis of mongodb and elastic consistency

akintolga:
With this PR:
- A new celery task run to compare the repos every hour
- Repos to be compared are: ingest, archive, published, text_archive
- A new page in settings "Data Consistency" to observe the figures
- Now we'll be able to see the items:
  - only in mongo
  - only in elastic
  - in both but different


marwoodandrew:
:+1:

mdhaman:
:+1:

server/superdesk/data_consistency/compare_repositories_tests.py
```diff
@@ -0,0 +1,128 @@
...
+import superdesk
+from eve.utils import ParsedRequest
+from superdesk.tests import TestCase
+from superdesk import get_resource_service
+from superdesk.data_consistency.compare_repositories import CompareRepositories
+from eve.methods.common import resolve_document_etag
+from time import sleep
+
+
+class RebuildIndexTestCase(TestCase):
```
sivakuna-aap: Probably the name should be `CompareRepositoriesTestCase`

server/superdesk/data_consistency/compare_repositories.py
```diff
@@ -0,0 +1,107 @@
...
+        cursor = service.get_from_mongo(None, {})
+        count = cursor.count()
+        no_of_buckets = len(range(0, count, self.default_page_size))
+        mongo_items = []
+        updated_mongo_items = []
+        for x in range(0, no_of_buckets):
+            skip = x * self.default_page_size
+            print('Page : {}, skip: {}'.format(x + 1, skip))
+            # don't get any new records since the elastic items are retrieved
+            cursor = service.get_from_mongo(None, {'_created': {'$lte': consistency_record['started_at']}})
```
amagdas: Adding a mongo projection would be nice so we retrieve only the `_id` and `_etag`.

amagdas:
:+1: besides the minor comment.

ioanpocol:
:+1:

server/superdesk/data_consistency/compare_repositories_tests.py
```diff
@@ -0,0 +1,128 @@
...
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+import superdesk
+from eve.utils import ParsedRequest
+from superdesk.tests import TestCase
+from superdesk import get_resource_service
+from superdesk.data_consistency.compare_repositories import CompareRepositories
```
plamut: It's a good strategy to import the class under test in a test class itself rather than at the test module level. The rationale behind it is that a failed import should not prevent the tests from being run, it should result in these tests' failures instead.

Just a suggestion.

server/superdesk/data_consistency/compare_repositories_tests.py
```diff
@@ -0,0 +1,128 @@
...
+        with self.app.app_context():
+            data = [{'headline': 'test {}'.format(i), 'slugline': 'rebuild {}'.format(i),
+                     'type': 'text' if (i % 2 == 0) else 'picture'} for i in range(1, 100)]
+            resolve_document_etag(data, 'archive')
+            superdesk.app.data._search_backend('archive').bulk_insert('archive', data)
+            get_resource_service('archive').post(data)
+            sleep(1)  # sleep so Elastic has time to refresh the indexes
+
+    def test_compare_repos(self):
+        with self.app.app_context():
```
plamut: Would it help to push the app context in the `setUp` method, and popping it in the `tearDown`? With that repeating the `with self.app.app_context():` line in every test case won't be needed anymore.

plamut:
Don't mind the broken build, it's the [same two tests](https://travis-ci.org/superdesk/superdesk/jobs/64230121#L1033) that often fail. Re-running the build helps like 99% of the time, those tests then automagically pass for no apparent reason.

server/superdesk/data_consistency/compare_repositories_tests.py
akintolga: I agree that the repetition of the with statement was not necessary in most of the test cases here. Especially on those where I test the 'process_results' case. So maybe I can revisit this towards the end of this sprint. Thanks for the comment.

plamut: Always glad to suggest possible code improvements, it will benefit us all in the long run. :wink:

Good night!

server/superdesk/data_consistency/compare_repositories_tests.py
akintolga: I think it is a good idea to come-up with a generic helper class like this where we can ask for ClassUnderTest to get loaded. I think we should do this across the board. Maybe we can create a technical ticket for this suggestion and one of us can pick it up. I'll probably merge this PR with its current shape now as we need to see this working. Thanks Peter.

plamut: Merging this PR is fine, yes.

Re: helper method/class - should be straightforward, we just add it to the `superdesk.tests.TestCase` and invoke it with the name of the class under test (as a string) that we want to import in a test module. Also thanks for taking care of opening a ticket.

PR #645, opened 2015-05-27
SD-2115 ninjs converter for public content database

mugurrus:


plamut:
Just wondering - what is the difference between the `server/apps/publicapi` and `server/publicapi` packages? It seems as if publicapi logic is split across two different packages, making it less obvious what the responsibilities of each of the packages are?

Thanks for clarifying this in advance.

mugurrus:
The apps/publicapi service is used for pushing data into the public content archive.

plamut:
Oh, OK, perhaps it's just the name of the new app ("publicapi") that causes a bit of confusion. It might make sense to think of another name that better describes the app, so that people won't mistakenly think that it implements public API's business logic.

mugurrus:
The best name I find now is public_content_publish which is a bit long; what do you think?

plamut:
Coming up with good names is often difficult, yet important. :)

What if we simply shorten it to `content_publish`, for example? The "publish" part by itself already implies in some way, that it's about "public" data, making the latter redundant.

mugurrus:
Public content is published also through email, ftp etc. This is specific to the public REST API.

plamut:
Hm... what about, let's say,  `api_data_publish`? Do you think that would describe it better?

server/apps/publish/archive_publish.py
```diff
@@ -188,7 +188,7 @@ def queue_transmission(self, doc, target_output_channels=None):
 
                 destination_groups = self.resolve_destination_groups(
                     doc.get('destination_groups'))
-                output_channels, selector_codes, format_types = \
+                output_channels, selector_codes, _format_types = \
```
akintolga: Hey Mugur, while I was looking at this change that I noticed an issue here, though it is nothing to do with your changes. We iterate through every output channel and format the story per output channel. We may have more than one output channel having the same format so we don't need to re-format again. I have created a technical task https://dev.sourcefabric.org/browse/SD-2420. So again thanks to your change in that field name.

plamut:
Sorry for accidentally commenting on the commit itself rather than on the pull request.

plamut:
Is this still WIP?

client/app/scripts/superdesk-publish/publish.js
```diff
@@ -29,6 +29,10 @@
         File: {
             label: 'File',
             templateUrl: 'scripts/superdesk-publish/views/file-config.html'
+        },
+        PublicArchive: {
+            label: 'Public Archive',
+            templateUrl: 'scripts/superdesk-publish/views/public-archive-config.html'
```
plamut: Is this needed here? The template itself is currently empty.

server/apps/publish/formatters/ninjs_formatter.py
```diff
@@ -0,0 +1,69 @@
...
+                ninjs['byline'] = self._get_byline(article)
+            except:
+                pass
+            for copy_property in self.direct_copy_properties:
+                if copy_property in article:
+                    ninjs[copy_property] = article[copy_property]
+            if article['type'] == 'composite':
+                article['associations'] = self._get_associations(article)
+
+            return pub_seq_num, json.dumps(ninjs, default=json_util.default)
```
plamut: There should also be _created and _updated fields here, Eve uses them to handle conditional HTTP requests (the `If-Modified-Since` header).

There is a [separate ticket](https://dev.sourcefabric.org/browse/SD-2429) for that, but we might as well resolve it here while we are adding this formatting functionality.

plamut:
As a general notice, docstrings would be helpful, especially to describe parameters and their types. Also unit tests should be added to improve the general test coverage.

ioanpocol:
:+1:

server/apps/publish/formatters/ninjs_formatter.py
mugurrus: done

petrjasek:
:+1:

amagdas:
:+1:

plamut:
For the reference: a bug was introduced in this pull request in a part of the code that is almost trivial ([generating the URIs](https://github.com/superdesk/superdesk/pull/645/files#diff-9b0ca67da673dc4f749b7fc515111f7eR35) for packages' `associations`). Unfortunately this issue caused me to lose quite some time today, even though it could have been easily prevented with a few simple tests that wouldn't take long to write (certainly less time that was later spent on debugging).

I know it's tempting to skip tests for seemingly trivial parts of the code, but as this example shows, bugs can appear there, too. Please do not underestimate the importance of tests and let's be more strict when merging new code, i.e. we should really make sure that the corresponding tests have been written as well.

Just my 2¢

PR #648, opened 2015-05-27
Fix (modal close): Fixed for all modal windows

Aca-jov:
SD-2294


client/app/docs/views/main.html
```diff
@@ -967,6 +967,13 @@ <h3 translate>Sample modal</h3>
                             scope.closeModal = function() {
                                 scope.isModalOpen = false;
                             };
+                            
+                            $(document).on('hidden.bs.modal', '.modal', function () {
+                                scope.modalActive = false;
+                                if (!scope.$$phase) {
```
plamut: Don't do that, it's considered an anti-pattern.  Variables names prefixed by `$$` are internal to Angular and their behaviour/existence can change without notice.

Check this link for an idea how to overcome the problem (the last section):
https://github.com/superdesk/superdesk-docs/blob/master/good-practices/angular-performance-tips.md

plamut: By the way, what is business logic doing here in the template? Since we are changing things here, I suggest we take this opportunity and refactor it out into an Angular component to avoid maintenance headaches in the future.

client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -137,6 +137,13 @@
                     scope.closeModal();
                 };
 
+                $(document).on('hidden.bs.modal', '.modal', function () {
+                    scope.closeModal();
+                    if (!scope.$$phase) {
```
plamut: Same comment as above (anti-pattern).

client/app/scripts/superdesk-desks/desks.js
```diff
@@ -392,7 +392,12 @@
             $scope.desk.edit = null;
         };
 
-        $(document).on('hidden.bs.modal', '.modal', function () { $scope.modalActive = false; });
+        $(document).on('hidden.bs.modal', '.modal', function () {
+            $scope.modalActive = false;
+            if (!$scope.$$phase) {
```
plamut: Same comment as above (anti-pattern).

client/app/scripts/superdesk-groups/groups.js
```diff
@@ -133,6 +133,13 @@
             $scope.group.edit = null;
         };
 
+        $(document).on('hidden.bs.modal', '.modal', function () {
+            $scope.modalActive = false;
+            if (!$scope.$$phase) {
```
plamut: Same comment as above (anti-pattern).

client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -276,6 +276,13 @@
             $scope.modalActive = false;
         };
 
+        $(document).on('hidden.bs.modal', '.modal', function () {
+            $scope.modalActive = false;
+            if (!$scope.$$phase) {
```
plamut: Same comment as above (anti-pattern).

client/app/styles/less/modals.less
```diff
@@ -62,7 +62,7 @@ body:not(:-moz-handler-blocked) .modal{
 		max-height: 100%;
 	}
 	.modal-content {
-		display: flex;
+		// display: flex;
```
plamut: If we don't need this anymore, we can as well remove it completely?

petrjasek:
this `$(document).on` handler should be defined only once, probably in the `superdesk.services.modal` in a `run` block. otherwise we would have to copy this every time we use a modal.. and it never unregisters those handlers, keeping those scopes around - so there is a memory leak.
to avoid this `$scope.$$phase` check you can use `$scope.$applyAsync()`

plamut:
:+1:
Should be fine now IMO (if the tests pass, of course).

amagdas:
:+1:

PR #651, opened 2015-05-28
fix(compare) - Elastic index value was misssing

akintolga:


amagdas:
:+1:

sivakuna-aap:
:+1: Its better to have some tests

PR #655, opened 2015-05-28
feat(content): use authoring for preview

petrjasek:


client/app/scripts/superdesk-archive/module.js
```diff
@@ -106,6 +106,79 @@ define([
         };
     }
 
+    ListLayoutController.$inject = ['preferences'];
+    function ListLayoutController(preferences) {
+        var MAX_VIEW = {max: true},
+            MIN_VIEW = {min: true},
+            ITEM_VIEW = {extended: true},
+            COMPACT_VIEW = {compact: true};
```
plamut: We might want to use [Object.freeze()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) for these kind of objects.

client/app/scripts/superdesk-archive/views/list.html
```diff
@@ -1,24 +1,68 @@
...
+                    <ul class="dropdown-menu pull-right" sd-content-create></ul>
+                </div>
+
+                <div class="navbtn dropdown" dropdown>
+                    <button class="dropdown-toggle" dropdown-toggle><i class="icon-dots-vertical"></i></button>
+                    <ul class="dropdown-menu" ng-if="layout.view.compact || layout.view.extended">
+                        <li>
+                            <button
+                                ng-disabled="layout.view.compact"
+                                ng-click="layout.compactView()">Compact mode</button>
```
plamut: Translate?

client/app/scripts/superdesk-archive/views/list.html
```diff
@@ -1,24 +1,68 @@
...
+                        </li>
+                        <li>
+                            <button
+                                ng-disabled="layout.view.extended"
+                                ng-click="layout.extendedView()">Extended mode</button>
```
plamut: Translate?

client/app/scripts/superdesk-search/search.js
```diff
@@ -629,32 +629,18 @@
...
             return {
                 require: '^sdSearchContainer',
                 templateUrl: asset.templateUrl('superdesk-search/views/search-results.html'),
                 link: function(scope, elem, attr, controller) {
 
-                    var GRID_VIEW = 'mgrid',
-                        LIST_VIEW = 'compact';
+                    var LIST_VIEW = 'compact';
 
                     var multiSelectable = (attr.multiSelectable === undefined) ? false : true;
```
plamut: This can be shortened to `attr.multiSelectable !== undefined`.

client/app/scripts/superdesk-search/views/search-results.html
```diff
@@ -17,46 +21,16 @@
...
-                <li ng-class="{active: tab === 'content'}">
-                    <button ng-click="tab = 'content'"><span translate>Content</span></button>
-                </li>
-                <li ng-class="{active: tab === 'metadata'}">
-                    <button ng-click="tab = 'metadata'"><span translate>Metadata</span></button>
-                </li>
-                <li ng-class="{active: tab === 'related'}">
-                    <button ng-click="tab = 'related'"><span translate>Related Items</span></button>
+            <ul class="{{view}}-view list-view">
+                <li ng-repeat="item in items._items" class="list-item-view" ng-class="{active: item === selected.preview || item._id === selected.fetch}">
```
plamut: Would it make sense to add `track by` here?

client/spec/authoring_spec.js
```diff
@@ -4,7 +4,7 @@
 var workspace = require('./helpers/workspace');
 
 describe('authoring', function() {
-    it('can open item stage', function() {
+    xit('can open item stage', function() { // obsolete?
```
plamut: This should be resolved before merging IMO. Either we keep the test or remove it completely.

client/spec/content_spec.js
```diff
@@ -42,7 +43,8 @@ describe('Content', function() {
         expect(element(by.id('search-input')).isDisplayed()).toBe(true);
     });
 
-    it('can toggle view with v', function() {
+    // todo(petr): should it toggle compact/extended/full view?
+    xit('can toggle view with v', function() { // obsolete
```
plamut: Same comment as above (resolving the necessity of the test).

client/spec/multicontent_spec.js
```diff
@@ -6,13 +6,14 @@ var workspace = require('./helpers/workspace'),
     content = require('./helpers/content'),
     multicontent = require('./helpers/multicontent');
 
-describe('multi content view', function() {
+// todo(petr): those features should go to list
+xdescribe('multi content view', function() {
```
plamut: Is "moving those features to list" (whatever that means) doable within this PR? If yes, we should do it in order to avoid having a disabled test lying around (these tend to get forgotten).

client/spec/package_spec.js
```diff
@@ -62,7 +62,8 @@ describe('Content', function() {
         expect(authoring.getGroupItems('MAIN').count()).toBe(3);
     });
 
-    it('can preview package in a package', function() {
+    // todo(petr): obsolete?
+    xit('can preview package in a package', function() {
```
plamut: Should be resolved before merging...

client/spec/multicontent_spec.js
petrjasek: it's not, that's why there is this todo

plamut: OK, if it needs more refactoring, then it's fine to leave the TODO here (just make sure that there is a ticket for that).

client/app/scripts/superdesk-search/views/search-results.html
petrjasek: it would, but there is no unique field to use

plamut: Would it be possible to perhaps join two fields to obtain a unique combination?
If not, well, then tough luck I guess...

petrjasek:
still not sure about those todos. i agree it should be resolved before merging this to master, but so far it's going to branch, is it ok? @plamut

petrjasek:
I mean is it ok to keep them?

plamut:
Meh, I forgot again that this is a PR to a topic branch. Yes, it's fine then, the TODOs will be resolved eventually before we merge the whole thing into master.

:+1:

amagdas:
:+1:

PR #657, opened 2015-05-28
feat(spellcheck): add keyboard shortcut for spellcheck

petrjasek:
it can be triggered via ctrl+shift+s

### DIFF
--- a/client/app/scripts/superdesk-authoring/views/authoring-topbar.html
+++ b/client/app/scripts/superdesk-authoring/views/authoring-topbar.html
@@ -111,7 +111,9 @@
                         <button type="button"
                             ng-click="spellcheckMenu.spellcheck()"
                             ng-disabled="spellcheckMenu.isAuto"
-                            translate>Check spelling</button>
+                            translate>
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-authoring/views/authoring-topbar.html
```diff
@@ -111,7 +111,9 @@
                         <button type="button"
                             ng-click="spellcheckMenu.spellcheck()"
                             ng-disabled="spellcheckMenu.isAuto"
-                            translate>Check spelling</button>
+                            translate>
+                                <span class="shortcut pull-right">Ctrl+Shift+S</span>
```
plamut: Should that be "translated" for Mac users somehow?

(meh, screw them :grinning: )

plamut:
:+1:

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
amagdas: Hey, I'm using a Mac :)

amagdas:
:+1:

PR #658, opened 2015-05-28
fix(current desk) - On refresh page keep selected the current desk on…

ioanpocol:
… workspace


mugurrus:
:+1:

plamut:
I think it's good, just fix the failing behave test and we're set.

:+1:

PR #660, opened 2015-05-29
feat(docker): add docker compose config for development

petrjasek:
it runs in dev mode, including livereload, via docker.

if you have docker-compose installed, you can test this via:
```
$ docker-compose build
$ docker-compose up
```
to create admin user, while having it up, run:
```
$ docker-compose run backend python3 manage.py users:create ...
```
client is on localhost:9000, server on localhost:5000/api
let me know if this works for anyone..

fyi bower.json is back in client (and kept in root) because you can't add/copy symlink in dockerfile

### DIFF
--- a/client/bower.json
+++ b/client/bower.json
@@ -1,1 +1,35 @@
-../bower.json
+{
+  "name": "superdesk-client",
+  "dependencies": {
+    "requirejs": "2.1.15",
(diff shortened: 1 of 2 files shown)


amagdas:
:+1:

ioanpocol:
:+1:

PR #662, opened 2015-06-01
[SD-2373][SD-2376] ANPA Output Format and file system distribution me…

marwoodandrew:
…thod

### DIFF
--- a/server/apps/publish/formatters/anpa_formatter.py
+++ b/server/apps/publish/formatters/anpa_formatter.py
@@ -21,97 +21,97 @@ class AAPAnpaFormatter(Formatter):
 
             pub_seq_num = superdesk.get_resource_service('output_channels').generate_sequence_number(output_channel)
 
-            anpa = b''
+            anpa = []
 
             # Selector code specific optional
             if selector_codes and len(selector_codes) > 0:
-                anpa = b'\x05'
+                anpa.append(b'\x05')
                 sel_codes = selector_codes[output_channel['_id']]
-                anpa += ' '.join(sel_codes).lower().encode('ascii')
-                anpa += b'\x0D\x0A'
+                anpa.append(' '.join(sel_codes).lower().encode('ascii'))
+                anpa.append(b'\x0D\x0A')
 
             # start of message header (syn syn soh)
-            anpa += b'\x16\x16\x01'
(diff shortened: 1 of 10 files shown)


sivakuna-aap:
:+1:

server/apps/publish/formatters/anpa_formatter.py
```diff
@@ -0,0 +1,119 @@
...
+            anpa += str(pub_seq_num).zfill(4).encode('ascii')
+            # field seperator
+            anpa += b'\x0A'  # -LF
+            anpa += article.get('priority', 'r').encode('ascii')
+            anpa += b'\x20'
+
+            category = article.get('anpa-category', {})
+            if 'qcode' in category:
+                anpa += article.get('anpa-category', {})['qcode'].encode('ascii')
+            # TODO TODO TODO
```
mdhaman: maybe create a ticket for this TODO.

amagdas:
:+1:

server/apps/publish/formatters/anpa_formatter.py
```diff
@@ -0,0 +1,119 @@
...
+            anpa += (b'\x20' + sign_off) if len(sign_off) > 0 else b''
+            anpa += b'\x0D\x0A'
+
+            anpa += b'\x03'  # ETX
+
+            # time and date
+            anpa += datetime.datetime.now().strftime('%d-%m-%y %H-%M-%S').encode('ascii')
+
+            anpa += b'\x04'  # EOT
+            anpa += b'\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A'
```
plamut: Repeatedly appending new chunks to an existing string is a slow (`O(n^2)`) operation. It's much better to use `bytearray`, a _mutable_ binary string data type for that ([official docs](https://docs.python.org/3/faq/programming.html#what-is-the-most-efficient-way-to-concatenate-many-strings-together)).

plamut:
Just for the reference - I created a synthetic test and concatenating chunks using a `bytearray` is approximately 40% faster compared to using a list.

Nevertheless, both are _significantly_ faster than concatenating immutable `bytes` strings.

:+1:

PR #647, opened 2015-05-27
modals.less: style issue with flex positioning fixed

vied12:
I removed these lines to fix this issue we have on Liveblog:
modal

PR #663, opened 2015-06-01
Fix css for modal windows

Aca-jov:


plamut:
:+1:

ioanpocol:
:+1:

vied12:
This PR reverts my PR https://github.com/superdesk/superdesk/pull/647.
Can I know what does it fix @Aca-jov ?

Aca-jov:
We had broken modal windows (on adding widgets), and we lost scroll on modal windows (on window resize), so this css is needed after all...

vied12:
And do you have any clue why it breaks our styles like https://github.com/superdesk/superdesk/pull/647#issue-81429753 ?

Aca-jov:
No... Is this the only broken modal, or every modal is like this?

vied12:
This is the only one we have.
here is the html http://pastie.org/private/cjtref5umqxdinwzjcneq
(from http://liveblog.sd-test.sourcefabric.org/#/liveblog on a "Create
Blog" click)

Aca-jov:
You need to add width:100% in form, that will resolve width problem, but on window resize, there is scroll problem (scroll is not appearing), so compare our modals with yours and try to find why it works on superdesk and not on liveblog..

https://master.sd-test.sourcefabric.org on top right you have add widget button, clik there for modal.

PR #664, opened 2015-06-01
feat(e2e test): Added e2e test for subject codes

pavlovicnemanja92:
SD-2403

### DIFF
--- a/client/spec/search_spec.js
+++ b/client/spec/search_spec.js
@@ -48,23 +48,15 @@ describe('Search', function() {
         var subject = element.all(by.css('.dropdown-nested')).first();
         var subjectToggle = subject.element(by.css('.dropdown-toggle'));
 
-        filterPanelButton.click().then(
-            function () {
-                return element.all(by.css('[ng-click="toggleModule()"]')).first().click();
-            }).then(
-            function () {
-                return subjectToggle.click();
-            }).then(
-            function () {
-                return subject.all(by.css('.nested-toggle')).first().click();
-            }).then(
-            function () {
-                return subject.all(by.repeater('term in activeTree')).first().click();
-            }).then(function () {
-            expect(element.all(by.repeater('item in items._items')).count()).toBe(0);
-            expect(element.all(by.repeater('parameter in tags.selectedParameters')).count()).toBe(1);
-            expect(element.all(by.repeater('t in item[field]')).count()).toBe(1);
-        });
+        filterPanelButton.click();
+        element.all(by.css('[ng-click="toggleModule()"]')).first().click();
+        subjectToggle.click();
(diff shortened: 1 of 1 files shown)


plamut:
:+1:

ioanpocol:
There are a few minor jshint errors.

client/spec/search_spec.js
```diff
@@ -39,6 +39,33 @@ describe('Search', function() {
         expect(element.all(by.repeater('items._items')).count()).toBe(1);
         expect(element.all(by.repeater('parameter in tags.selectedKeywords')).count()).toBe(1);
     });
+    
+    it('can search by subject codes field', function() {
+        workspace.switchToDesk('SPORTS DESK').then(content.setListView);
```
ioanpocol: In protractor docs is specified that '.then()' is not needed and most (but not all :) ) of the tests works without it.
Probably it is better to avoid them and use only where problems appears. 
@petrjasek Probably we have to decide about the style used to write e2e tests.
Also all selections and actions should be moved on a helper file. This is an issue for all tests from this file, so when/if it will be done, it should be done for all tests.

ioanpocol:
:+1:  (except tests style issues)

client/spec/search_spec.js
pavlovicnemanja92: I had a problem with expected result and because of that I create structure with then() ... Test take result before it completes filtering and still displaying 2 items (it need to be 0, because there are no specified subject codes)

client/spec/search_spec.js
```diff
@@ -40,6 +40,33 @@ describe('Search', function() {
...
 
+    it('can search by subject codes field', function () {
+        workspace.switchToDesk('SPORTS DESK').then(content.setListView);
+        expect(element.all(by.repeater('items._items')).count()).toBe(2);
+
+        var filterPanelButton = element(by.css('.fitler-trigger'));
+        var subject = element.all(by.css('.dropdown-nested')).first();
+        var subjectToggle = subject.element(by.css('.dropdown-toggle'));
+
+        filterPanelButton.click().then(
```
petrjasek: imo it would be better to avoid `then` calls, it should work same like:
```
filterPanelButton.click()
element.all(by.css('[ng-click="toggleModule()"]')).first().click();
subjectToggle.click();
subject.all(by.css('.nested-toggle')).first().click();
subject.all(by.repeater('term in activeTree')).first().click();

expect(element.all(by.repeater('item in items._items')).count()).toBe(0);
expect(element.all(by.repeater('parameter in tags.selectedParameters')).count()).toBe(1);
expect(element.all(by.repeater('t in item[field]')).count()).toBe(1);
```
protractor should handle the promises

amagdas:
:+1:

PR #668, opened 2015-06-03
(fix) Authoring: Allow duplicate of item even if the item belongs to another desk.

mdhaman:


plamut:
:+1:

akintolga:
:+1:

plamut:
Perhaps also add test(s) for the discovered (and now fixed) bugs so that we can be sure they don't appear again in the future.

amagdas:
Merging this as more PR's are coming for this feature(publishing).

PR #669, opened 2015-06-03
[sd-2293]-Killed item destination groups drop down should not be active(handled scheduler in the meta data be inactive in view mode)

sjunaid:


sivakuna-aap:
Changes related to `Killed item destination groups drop down should not be active` are missing I guess. I see changes for disabling date and time pickers only.

sjunaid:
Siva, that dropdown part was previously handled and merged in another ticket SD-2365.

sivakuna-aap:
ok :+1:

amagdas:
:+1:

PR #672, opened 2015-06-04
(fix) Authoring: Minor fixes

mdhaman:
- Fixing the not for publication flag.
- Strip html from the headline and abstract
- Fixing the unit test.

### DIFF
--- a/client/app/scripts/superdesk-archive/views/media-box-list.html
+++ b/client/app/scripts/superdesk-archive/views/media-box-list.html
@@ -32,7 +32,7 @@
             <div class="takekey" ng-if="item.anpa_take_key">{{ item.anpa_take_key }}</div>
             <div class="category" ng-if="item['anpa-category'].name">{{ item['anpa-category'].name }}</div>
             <div class="provider" sd-meta-ingest data-item="item"></div>
-            <div ng-if="item.marked_for_not_publication == true" class="not_for_publication" translate>Not for publication</div>
+            <div ng-if="item.marked_for_not_publication" class="state-label not-for-publication">Not For Publication</div>
             <div class="expires" ng-if="item.is_spiked"><span translate>expires</span> {{ item.expiry | reldate }}</div>
             <div ng-if="item.archived"><div sd-fetched-desks data-item="item"></div></div>
         </div>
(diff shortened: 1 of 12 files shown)


client/app/scripts/superdesk-archive/views/media-box-list.html
```diff
@@ -32,7 +32,7 @@
             <div class="takekey" ng-if="item.anpa_take_key">{{ item.anpa_take_key }}</div>
             <div class="category" ng-if="item['anpa-category'].name">{{ item['anpa-category'].name }}</div>
             <div class="provider" sd-meta-ingest data-item="item"></div>
-            <div ng-if="item.marked_for_not_publication == true" class="not_for_publication" translate>Not for publication</div>
+            <div ng-if="item.marked_for_not_publication" class="state-label not-for-publication">Not For Publication</div>
```
plamut: Was this `translate` here deleted by an accident?

client/app/scripts/superdesk-archive/views/metadata-view.html
```diff
@@ -1,7 +1,10 @@
 <div class="metadata-view">
     <dl>
         <dt translate>State</dt>
-        <dd><div class="state-label state-{{item.state}}">{{item.state}}</div></dd>
+        <dd>
+            <div class="state-label state-{{item.state}}">{{item.state}}</div>
+            <div ng-if="item.marked_for_not_publication" class="state-label not-for-publication">Not For Publication</div>
```
plamut: Missing `translate`?

plamut:
I think it's good (apart from a few missing  `translate` directives).
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -722,6 +728,15 @@
                     }
                 }, true);
 
+                $scope.$watch('item.marked_for_not_publication', function(newValue, oldValue) {
+                    if (newValue !== oldValue) {
```
petrjasek: imo this check is extra, if this is not true, watch shouldn't be triggered

petrjasek:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
plamut: Actually a watch _can_ be triggered in the initialization phase with both old and new values equal. If triggering the handler is not desired in such case, this type of check is necessary.

amagdas:
:+1:

PR #673, opened 2015-06-04
SD 2389 : (plus SD-2390) -UI: Reload the privileges, desks and stages when there's a RELOAD message 

sjunaid:


client/app/scripts/superdesk/notification/notification.js
```diff
@@ -29,6 +43,101 @@
         };
     }
 
-    return angular.module('superdesk.notification', [])
+    ReloadService.$inject = ['$window', '$rootScope', 'session', 'desks'];
+    function ReloadService($window, $rootScope, session, desks) {
+        var userDesks;
```
akintolga: All looking good. Few tests for this new service would be very good. :+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -867,6 +906,19 @@
...
+                    var changeMsg = msg;
+                    authoring.saveWorkConfirmation($scope.origItem, $scope.item, $scope.dirty, changeMsg)
+                    .then(function(res) {
+                        desks.setCurrentDesk(null);
+                        $location.url('/workspace/content');
+                        referrer.setReferrerUrl('/workspace/content');
+                        $window.location.reload(true);
+                    }, function(response) {
+                        notify.error(gettext('Error: Saving work on configuration changes.'));
+                        $window.location.reload(true);
```
plamut: This can be moved to `.finally()` to avoid duplicate code in both handlers.

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -29,6 +43,101 @@
...
+        var stageEvents = {
+            'stage': 'Stage is created/updated/deleted',
+            'stage_visibility_updated': 'Stage visibility change'
+        };
+
+        $rootScope.$on('reload', function(event, msg) {
+
+            var _result = reloadIdentifier(msg);
+
+            if (_result.reload === true) {
```
plamut: Explicit comparison to `true` kind of redundant here?

plamut:
I agree with @akintolga on tests. Looks good in general, though (the remarks I made are not that significant).

:+1:

PR #675, opened 2015-06-04
public API: slightly improve the invalid date error message

plamut:
I added a word "valid" to the error message to make it more clear that a date like 2015-04-31 is rejected not because of its format, but because it is does not represent a _valid_ date (due to the incorrect number of days in this case).


ioanpocol:
:+1:

amagdas:
:+1:

PR #676, opened 2015-06-04
feat(authoring): Implemented saving/loading state to aggregate widget.

eozan:
SD-2394

### DIFF
--- a/client/app/scripts/superdesk-desks/aggregate.js
+++ b/client/app/scripts/superdesk-desks/aggregate.js
@@ -268,7 +268,7 @@
         };
     }
 
-    angular.module('superdesk.aggregate.sidebar', ['superdesk.authoring.widgets'])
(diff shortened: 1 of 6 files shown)


plamut:
Tests are missing, the rest looks good to me.

ioanpocol:
:+1:  (except e2e tests)

plamut:
No, rather prefer unit tests. :wink: (the best of course is to have both)

The reason is behave tests are _slow_ and thus do not encourage running them repeatedly when you write code. Unit tests, on the other hand, run much quicker and give you feedback in a much more reasonable time. On top of that, they also point you to the exact part of the code that failed, making  it easier to locate bugs.

Both of the above-mentioned properties of unit tests are essential for efficient TDD.

plamut:
:+1:

petrjasek:
:+1:

PR #677, opened 2015-06-04
Feat multi fetch

petrjasek:
implementing multi fetch + fetchAs

### DIFF
--- a/client/app/scripts/superdesk-archive/module.js
+++ b/client/app/scripts/superdesk-archive/module.js
@@ -19,7 +19,7 @@ define([
          *
          * @param {Object} item
          */
-        this.isSelected = function isSelected(item) {
+        this.isSelected = function(item) {
             return !!_.find(items, {_id: item._id});
         };
 
@@ -28,7 +28,7 @@ define([
          *
          * @param {Object} item
          */
-        this.toggle = function toggle(item) {
+        this.toggle = function(item) {
             item.selected = !this.isSelected(item);
             if (item.selected) {
                 items = _.union(items, [item]);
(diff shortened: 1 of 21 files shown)


client/app/scripts/superdesk-archive/module.js
```diff
@@ -57,9 +57,15 @@ define([
             this.count = 0;
         };
 
+        this.items = getItems;
```
plamut: What's with this public alias for the `getItems` function? Or was it perhaps meant to be a function call, e.g. `getItems()`?

plamut:
Looking good so far IMO.

client/app/scripts/superdesk-archive/module.js
petrjasek: yep exactly. but it's kind of weird when mixed together with `this.reset = function () { ... };`
will clean this a bit

client/app/scripts/superdesk-archive/module.js
```diff
@@ -57,6 +57,13 @@ define([
             this.count = 0;
         };
 
+        /**
+         * Get list of selected items
+         */
+        this.getItems = function getItems() {
```
plamut: Method name redundant here?

petrjasek: it's better for stack traces / profiling

plamut: Yeah, but don't stack traces already contain the name that is assigned to the method when it is bound to controller?

petrjasek: yep you're right, no need for the names there

ioanpocol:
:+1:

PR #678, opened 2015-06-05
[SD-2351] - NewsML 1.2 Formatting for published items

akintolga:


server/apps/publish/formatters/newsml_1_2_formatter.py
```diff
@@ -0,0 +1,163 @@
...
+    def __format_news_envelope(self, article, news_envelope, pub_seq_num):
+        SubElement(news_envelope, 'TransmissionId').text = str(pub_seq_num)
+        SubElement(news_envelope, 'DateAndTime').text = self.now.strftime('%Y%m%dT%H%M%S+0000')
+        SubElement(news_envelope, 'Priority', {'FormalName': article.get('priority', '')})
+
+    def __format_identification(self, article, news_item):
+        revision = self.__process_revision(article)
+        identification = SubElement(news_item, "Identification")
+        news_identifier = SubElement(identification, "NewsIdentifier")
+        SubElement(news_identifier, 'ProviderId').text = 'aap.com.au'
```
mdhaman: Extract to the formatter settings at output channel level.

client/app/scripts/superdesk-publish/views/output-channels.html
```diff
@@ -54,6 +54,7 @@ <h3 translate ng-hide="outputChannel._id" translate>Add New Output Channel</h3>
                             <label translate>Format</label>
                             <select id="outputChannel-format" class="input" required ng-model="outputChannel.format">
                                 <option value="nitf" translate>NITF</option>
+                                <option value="newsml12" translate>NewsML 1.2</option>
```
amagdas: I think the name should be 'AAP NewsML 1.2' as it's specific to AAP.

server/apps/publish/formatters/newsml_1_2_formatter.py
```diff
@@ -0,0 +1,163 @@
...
+                 'omissions in or reliance on AAP content.  The globe symbol and "AAP" are registered trade marks. ' \
+                 'Further this AAP content is supplied to the direct recipient pursuant to an Information Supply ' \
+                 'Agreement with AAP (AAP Information Supply Agreement).  The direct recipient has a non-exclusive, ' \
+                 'non-transferable right to display this AAP content in accordance with and subject to the ' \
+                 'terms of the AAP Information Supply Agreement.'
+    LIMITATIONS = 'The direct recipient must comply with the limitations specified in the AAP Information ' \
+                  'Supply Agreement relating to the AAP content including, without limitation, not permitting ' \
+                  'redistribution and storage of the content (outside the terms of the Agreement) and not ' \
+                  'permitting deep hyperlinking to the content, framing of the content on a web site, ' \
+                  'posting the content to usenet newsgroups or facilitating such actions.'
```
plamut: If you prefer, you can get rid of all these ending backslashes by using implicit string concatenation (but it's more a matter of style than anything else):
```python
long_string = ('abcde.........'
               'next line in the string'
               'and another line'
               'this is the last line')
```

sivakuna-aap: Or we can use Multiline String feature of python

server/apps/publish/formatters/newsml_1_2_formatter.py
```diff
@@ -0,0 +1,163 @@
...
+
+            newsml = etree.Element("NewsML")
+            SubElement(newsml, "Catalog", {'Href': 'http://www.aap.com.au/xml-res/aap-master-catalog.xml'})
+            news_envelope = SubElement(newsml, "NewsEnvelope")
+            news_item = SubElement(newsml, "NewsItem")
+
+            self.__format_news_envelope(article, news_envelope, pub_seq_num)
+            self.__format_identification(article, news_item)
+            self.__format_news_management(article, news_item)
+            self.__format_news_component(article, news_item)
```
plamut: IMO it's better and more pythonic to simply use a single dash to prefix the names of the "private" methods (as they are never really private anyway). It makes subclassing easier, for instance, and as Guido says _"we are all consenting adults here"_.

A single underscore is enough for any adult to understand that a method is meant to be private. :)

sivakuna-aap: As per python [documentation](https://docs.python.org/2/tutorial/classes.html#private-variables-and-class-local-references), for defining a private function in a module then it's a single underscore and for defining a private method in a class then it's a double underscore.

server/apps/publish/formatters/newsml_1_2_formatter.py
```diff
@@ -0,0 +1,163 @@
...
+        SubElement(copyright, 'CopyrightHolder').text = article.get('source', article.get('original_source', ''))
+        SubElement(copyright, 'CopyrightDate').text = self.now.strftime("%Y")
+
+        usage_rights = SubElement(rights_metadata, "UsageRights")
+        SubElement(usage_rights, 'UsageType').text = self.USAGE_TYPE
+        SubElement(usage_rights, 'Geography').text = article.get('place', article.get('located', ''))
+        SubElement(usage_rights, 'RightsHolder').text = article.get('source', article.get('original_source', ''))
+        SubElement(usage_rights, 'Limitations').text = self.LIMITATIONS
+        SubElement(usage_rights, 'StartDate').text = self.now.strftime('%Y%m%dT%H%M%S+0000')
+        SubElement(usage_rights, 'EndDate').text = self.now.strftime('%Y%m%dT%H%M%S+0000')
```
plamut: This datetime format specification occurs at multiple places, perhaps move it to some class level "constant"?

server/apps/publish/formatters/newsml_1_2_formatter.py
```diff
@@ -0,0 +1,163 @@
...
+
+        for subject in article.get('subject', []):
+            SubElement(subject_code, 'Subject', {'FormalName': subject.get('qcode', '')})
+
+        # For now there's only one category
+        SubElement(descriptive_metadata, 'Property',
+                   {'FormalName': 'Category', 'Value': article['anpa-category']['qcode']})
+
+        # TODO: Subcategory
+        # TODO: Locator
```
plamut: Forgotten TODOs or just a reminder for future tickets?

plamut:
There is a test missing for the `newml12FormatterError`, but the rest looks good (the remarks are just suggestions, nothing serious).

:+1:

server/apps/publish/formatters/newsml_1_2_formatter.py
plamut: Well, it's not really private, it's just the name that changes so that directly referencing such a variable as `instance.__private_var` does not work, but it's still accessible if one really wants to get to it (even the documentation says that).

In general it is considered a discouraged practice in Python.


Actually Python doesn't have private members. The only thing that happens is name mangling so

server/apps/publish/formatters/newsml_1_2_formatter.py
plamut: You mean the triple quotes? That's an option, too, indeed.
Perhaps its only downside is that it considers all indentation whitespace as a part of the string, which is probably not desirable for a human-readable text.

server/apps/publish/formatters/newsml_1_2_formatter.py
akintolga: I'd like to keep those TODOs in place for the time being if they're not a big problem. Subcategory and Locator fields need a re-visit at some stage.

server/apps/publish/formatters/newsml_1_2_formatter.py
akintolga: I separated providerId as well as the copyright text to aap_settings.py file which I think can later be used for AAP specific settings.

akintolga:
Thanks all for the review.

amagdas:
:+1:

server/apps/publish/formatters/newsml_1_2_formatter.py
plamut: In that case we should keep them, yes, just wanted to make sure.
Thanks for explaining!

plamut:
:+1:

PR #682, opened 2015-06-06
[SD-2261]  Expire Published Items and Expire Archive Collection Items.

sivakuna-aap:
1. Expire Published Items and Expire Archive Collection Items.
2. Modified Vocabularies API to pop out 'is_active' property from response and filter out inactive vocabulary items


server/apps/common/models/base_model.py
```diff
@@ -157,28 +157,31 @@ def update(self, filter, doc, etag=None):
         return res
 
     def replace(self, filter, doc, etag=None):
```
plamut: Since we are changing this file already, should we also rename the `filter` parameter in a couple of methods, because it clashes with the name of a built-in class?

server/apps/legal_archive/components/archive_component.py
```diff
@@ -21,24 +20,22 @@ def __init__(self, app):
...
 
-    def update(self, original, updates):
-        updated = copy(original)
-        updated.update(updates)
-        updated['_id_document'] = original['_id']
-        del updated['_id']
+    def replace(self, filter, doc, etag=None):
         try:
-            return get_model(LegalArchiveModel).create([updated])
+            return get_model(LegalArchiveModel).replace(filter, doc, etag=None)
```
plamut: Same question here (the name of the `filter`parameter).

server/apps/publish/published_item.py
```diff
@@ -174,24 +293,162 @@ def insert_into_text_archive(self, doc):
...
+        lookup = {config.ID_FIELD: legal_archive_doc[config.ID_FIELD]}
+        article_in_legal_archive = legal_archive_component.find_one(lookup)
+        if article_in_legal_archive:
+            legal_archive_component.replace(lookup, legal_archive_doc)
+        else:
+            legal_archive_component.create([legal_archive_doc])
+
+        get_component(LegalArchiveVersions).create(legal_archive_doc_versions)
+        get_component(FormattedItems).create(formatted_items)
+        get_component(PublishQueue).create(queue_items)
```
petrjasek: i'm a bit lost here, seems like calling create on items that exist already, can you maybe add some comment here to explain this part?

server/apps/publish/published_item.py
```diff
@@ -174,24 +293,162 @@ def insert_into_text_archive(self, doc):
...
+            del queue_item[config.ETAG]
+            queue_item['output_channel_id'] = output_channels[str(queue_item['output_channel_id'])]['name']
+            queue_item['subscriber_id'] = subscribers[str(queue_item['subscriber_id'])]['name']
+        logging.info(
+            'De-normalized the Transmission Detail records of article %s' % legal_archive_doc.get('unique_name'))
+
+        # De-normalizing the legal archive doc
+        self.__denormalize_user_dg_desk(legal_archive_doc)
+
+        # Get Version History
```
petrjasek: maybe this could be a method on archive_versions service

server/apps/publish/published_item.py
```diff
@@ -174,24 +293,162 @@ def insert_into_text_archive(self, doc):
...
+
+        # Fetch Publish Queue Items
+        formatted_item_ids = [str(formatted_item[config.ID_FIELD]) for formatted_item in formatted_items]
+        query = {'$and': [{'formatted_item_id': {'$in': formatted_item_ids}}]}
+        queue_items = list(get_resource_service('publish_queue').get(req=None, lookup=query))
+        assert len(queue_items) > 0, \
+            "Transmission Details are empty for published item %s" % legal_archive_doc[config.ID_FIELD]
+        logging.info('Fetched transmission details for article %s' % legal_archive_doc.get('unique_name'))
+
+        output_channel_ids = list(set([str(queue_item['output_channel_id']) for queue_item in queue_items]))
```
plamut: We can use set comprehension here directly, avoiding the need for the "inner" list.

petrjasek:
:+1:

server/apps/publish/published_item.py
```diff
@@ -174,24 +293,162 @@ def insert_into_text_archive(self, doc):
...
+        query = {'$and': [{config.ID_FIELD: {'$in': output_channel_ids}}]}
+        output_channels = list(get_resource_service('output_channels').get(req=None, lookup=query))
+        output_channels = {str(output_channel[config.ID_FIELD]): output_channel for output_channel in output_channels}
+
+        subscriber_ids = list(set([str(queue_item['subscriber_id']) for queue_item in queue_items]))
```
plamut: Same here, no need for the inner list comprehension.

plamut:
My comments and suggestions (summary):

1. It's good to see docstrings, nice job!
2. `filter` parameter should probably be renamed in a couple of places to avoid shadowing the built-in with the same name.
3. A few redundant list comprehensions spotted.
4. As per general convention, "private" class methods in Python should be prefixed by a single underscore. Double underscores are only used in rare cases were name clashes with subclasses' methods can occur.

Other than that, it looks good to me in general, though I can't say whether I'm missing something in the big picture, there are quite a lot of changes in this PR.

amagdas:
Closing this, will open a new PR with latest updates from Eve.

PR #684, opened 2015-06-08
feat(multi highlights): Mark for highlights multiple items

ioanpocol:


client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -117,6 +117,40 @@
...
+                            item.highlights = [highlight._id].concat(item.highlights);
+                        } else {
+                            return;
+                        }
+                        highlightsService.mark_item(highlight._id, item._id);
+                    });
+                    multi.reset();
+                };
+
+                scope.is_marked = function is_marked(highlight) {
```
plamut: Redundant function name? (here and and a few lines above)
Since the function is bound to scope, it is not anonymous anymore...

Also, camelCase should be used for consistency.

plamut:
Apart for a minor remark it looks good IMO.

:+1:

client/app/scripts/superdesk-archive/tests/archive_spec.js
```diff
@@ -22,7 +22,7 @@ describe('content', function() {
 
     describe('multi service', function() {
         it('can reset on route change', inject(function(multi, $rootScope) {
-            multi.toggle({_id: 1});
+            multi.toggle({_id: 1, selected: true});
```
mugurrus: This is a bit confusing: toggle should switch the selection state. If you want to set a selection state given by you create a separate method.

mugurrus:
:+1:  (the test needs to be fixed)

client/app/scripts/superdesk-archive/tests/archive_spec.js
ioanpocol: The item.selected is set as a model for selection so it will be changed directly when the selection change is performed.

PR #685, opened 2015-06-08
feat(docker): add mail service for user activation

actionless:


docker/docker-compose.yml
```diff
@@ -29,6 +29,14 @@ elastic:
   volumes:
    - ../data/elastic:/usr/share/elasticsearch/data
 
+postfix:
+  image: catatnight/postfix
+  ports:
+  - "25:25"
```
petrjasek: is it possible to do this via link and not exposing the port outside?

actionless: yes

petrjasek:
:+1:

PR #688, opened 2015-06-09
Refactor legal archive. Upgrade to latest Eve.

amagdas:
Replaces https://github.com/superdesk/superdesk/pull/682


plamut:
@amagdas So basically everything is the same as in #682 , except for the bumped Eve version? So no need to review it again?

sivakuna-aap:
@plamut Not exactly. Apart from bumping up eve version, he also refactored component and model of legal archive to use multi-db support.

plamut:
OK, thanks, I'll have a closer look at it then (when I manage at least...).

server/apps/archive/commands.py
```diff
@@ -28,26 +28,29 @@ def remove_expired_content(self):
         logger.info('Removing expired content')
         now = date_to_str(utcnow())
         items = self.get_expired_items(now)
+
         while items.count() > 0:
             for item in items:
                 logger.info('deleting {} expiry: {} now:{}'.format(item['_id'], item['expiry'], now))
                 superdesk.get_resource_service('archive').delete_action({'_id': str(item['_id'])})
```
mdhaman: Are we checking if the spiked item is part of a package? Are we removing the referenced from the package?

marwoodandrew:
:+1:

mdhaman:
:+1:

akintolga:
:+1:

PR #690, opened 2015-06-09
Fix (item fetched by multiple desks looks)

Aca-jov:
SD-1881

### DIFF
--- a/client/app/scripts/superdesk-archive/directives.js
+++ b/client/app/scripts/superdesk-archive/directives.js
@@ -231,7 +231,7 @@
                 }
             };
         }])
-        .directive('sdFetchedDesks', ['familyService', function(familyService) {
+        .directive('sdFetchedDesks', ['desks', 'familyService', '$location', function(desks, familyService, $location) {
             return {
                 scope: {
                     item: '='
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-archive/directives.js
```diff
@@ -243,6 +243,15 @@
                             familyService.fetchDesks(scope.item, false)
                                 .then(function(desks) {
                                     scope.desks = desks;
+                                    scope.tooltipText = '';
+
+                                    for (var i = 1; i < desks.length; i++) {
```
plamut: Is `i`Initialized to 1 on purpose or should it be zero?
(BTW, using the built-in [Array.prototype.forEach](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) avoids these kinds of problems altogether, less error prone)

Aca-jov: It's on purpose, first desk should be shown as link, and other desks as tooltip.

plamut: Oh, OK then. It's probably good to add a comment, though, to explain that this is intentional.

petrjasek: it could be also written in func style..
```
scope.tooltipText = desks.map(function(desk) { return desk.name; }).slice(1).join(', ');
```

plamut: Or perhaps even shorter, without the mapping function:
```js
scope.tooltipText = _.map(desks, 'name').slice(1).join(', ');
```

Aca-jov:
This will be reworked, so no need to merge.

ioanpocol:
:+1:

mugurrus:
:+1:

PR #692, opened 2015-06-09
fix(fetch): fix fetch as for single item

petrjasek:


ioanpocol:
:+1:

amagdas:
rebase needed...

PR #693, opened 2015-06-09
fix(templates): let client send template desk=null

petrjasek:
so that template can be set from desk template to personal


plamut:
:+1:

ioanpocol:
:+1:

PR #694, opened 2015-06-09
fix(templates): Added 'none' option to desk dropdown in 'save as temp…

eozan:
…late', changed any to none in all.

SD-2484


plamut:
:+1:

amagdas:
:+1:

PR #695, opened 2015-06-09
fix(storage): disable task to check consistency

petrjasek:
it's using too much memory atm which makes oom-killer stop either mongo or
elasticsearch when running this.


petrjasek:
@superdesk/developers @akintolga

akintolga:
:+1: That makes sense. When we need it again we can always turn it back on for few days.

amagdas:
:+1:  Restarted the e2e tests, will merge if it's green.

PR #696, opened 2015-06-09
feat(package): Create package from multiple items selection

ioanpocol:


client/app/scripts/superdesk-search/search.js
```diff
@@ -1215,6 +1215,17 @@
                         multiEdit.create(multi.getIds());
                         multiEdit.open();
                     };
+
+                    scope.createPackage = function() {
+                        packages.createPackageFromItems(multi.getItems())
+                        .then(function(new_package) {
+                            superdesk.intent('author', 'package', new_package);
+                        }, function(response) {
+                            if (response.status === 403 && response.data && response.data._message) {
```
plamut: What about other types of errors, do we really want to silence them?

ioanpocol: There are some package related validation that should be showed to the user and all those errors has 403 error code. 
For other critical errors we don't show any other message. In the application there is a mixed style re: critical errors. We have to decide what style to adopt and use it consistently.

plamut: I agree with that, we need to revise the error codes and make them consistent across the application.

client/app/scripts/superdesk-search/search.js
```diff
@@ -1202,8 +1202,8 @@
             };
         })
 
-        .directive('sdMultiActionBar', ['asset', 'multi', 'multiEdit',
-        function(asset, multi, multiEdit) {
+        .directive('sdMultiActionBar', ['asset', 'multi', 'multiEdit', 'packages', 'superdesk', 'notify',
+        function(asset, multi, multiEdit, packages, superdesk, notify) {
             return {
                 controller: 'MultiActionBar',
                 controllerAs: 'action',
```
plamut: I noticed that some of the methods and values are bound to the scope, while the others are bound to the controller object - I suggest we make this consistent.

(preferably bind everything to controller, it's a more modern pattern in Angular).

client/app/scripts/superdesk-search/search.js
ioanpocol: https://dev.sourcefabric.org/browse/SD-2508

plamut: That's good, thanks for opening it.

plamut:
It would be a bit better to bind things to the controller rather than the scope, but as long as it's consistent, it's fine.

:+1:

mugurrus:
:+1:

plamut:
Just noticed - there is a semi-serious problem with the `sdMultiActionBar` directive, and that it's that it lacks its own scope. If there is, for example, a method named `send` defined in the parent element's scope, this directive replaces it, which can cause some hard-to-trace bugs.

This should be fixed.

petrjasek:
@plamut it was made to split authoring view which was too compex, it's not ment to be reusable component

plamut:
Hm, OK, that explains it a bit. Would be nice though if this was documented somewhere, even a simple directive docstring would already be useful. :)

ioanpocol:
The issue has been fixed by PR #699

plamut:
Yeah, that's good.

PR #697, opened 2015-06-09
[WIP] RSS ingest: extract referenced images' URLs from RSS entries

plamut:
I am adding support for ingesting external images referenced by entries (items) in an RSS feed. The first step is to find all URLs that reference an external image.

Please check the MIME types and the file extensions for the image types we want to support and let me know if anything should be changed. Please also verify that the list of RSS tags we take into account when looking for image URLs is indeed complete.

Thanks!

### DIFF
--- a/server/superdesk/io/rss.py
+++ b/server/superdesk/io/rss.py
@@ -60,6 +60,23 @@ class RssIngestService(IngestService):
     * type - field's data type
     """
 
+    IMG_MIME_TYPES = (
+        'image/bmp',
+        'image/gif',
+        'image/jpeg',
+        'image/png',
+    )
+    """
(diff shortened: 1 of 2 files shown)


mugurrus:
Link element defines the hyperlink to the item, not media content, see: http://www.w3schools.com/webservices/rss_tag_title_link_description_item.asp

Enclosure item is used for media content: http://www.w3schools.com/webservices/rss_tag_enclosure.asp

plamut:
The `feedparser` library places the parsed enclosure tags into the `.links` attribute on the parsed item (along with other links), hence the code that examines this list.

mugurrus:
:+1:

plamut:
Closing this, will issue a new PR with the next part of this feature.

PR #698, opened 2015-06-10
SD-2321 As a user I want to select multiple items and perform SPIKE or UNSPIKE actions

mugurrus:


client/spec/helpers/content.js
```diff
@@ -23,6 +23,10 @@ function Content() {
         });
     };
 
+    this.getItems = function(item) {
```
ioanpocol: Why item is needed?

client/spec/helpers/content.js
```diff
@@ -23,6 +23,10 @@ function Content() {
...
+        return element.all(by.repeater('items._items'));
+    };
+
     this.getItem = function(item) {
         return element.all(by.repeater('items._items')).get(item);
```
ioanpocol: Can be refactored to use getItems.

ioanpocol:
:+1:

amagdas:
:+1:

plamut:
This should be refactored, it suffers from the same problem that Ioan fixed in #699 (the `sdMultiActionBar` directive might accidentally overwrite things in the parent element's scope). It would be best to bind these spike/unspike methods to the controller as well, just like in the linked PR.

ioanpocol:
Already done on last update of #699

plamut:
Ah, good then!

PR #699, opened 2015-06-10
fix(package): Add local scope to sdMultiActionBar directive

ioanpocol:


plamut:
Hm, it seems that things up in the scope hierarchy want to access `scope.multi`, thus the failed test. Probably two-way binding of the `multi` object would do the trick.

plamut:
:+1:

amagdas:
:+1:  Rebase  needed

PR #701, opened 2015-06-10
RSS ingest: add a method for fetching remote images from given URLs

plamut:
I would like to get feedback on handling the image download errors - should we silently ignore them and simply exclude these images from the result, or should we perhaps raise an error and completely abort the ingesting of an RSS entry as a content package?

### DIFF
--- a/server/superdesk/io/rss.py
+++ b/server/superdesk/io/rss.py
@@ -14,6 +14,7 @@ import requests
 
 from calendar import timegm
 from collections import namedtuple
+from concurrent import futures
 from datetime import datetime
 
 from superdesk.errors import IngestApiError, ParserError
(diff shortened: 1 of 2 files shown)


amagdas:
@plamut For the other ingest channels, we abort the whole entry/item if the referenced image failed to download. I think we should be consistent on this one. This is already handled in https://github.com/superdesk/superdesk/blob/master/server/superdesk/io/commands/update_ingest.py#L325 , IMO the only work to be done is to pass the data from RSS in the proper shape so `ingest_item` does it work properly.

plamut:
OK, I'll change the behavior then and raise an error instead.

BTW, is creating a _package_ different in any way? I simply return a properly structured dictionary from the *_update()* method and that should be it?

amagdas:
Yup, returning a package like dictionary should be it. The rest of the ingest stack will take care of it.

PR #702, opened 2015-06-11
[SD-2025] - Managing user configuration changes

akintolga:


amagdas:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -265,6 +265,22 @@
...
 
+        this.saveWorkConfirmation = function saveWorkAuthoring(orig, diff, isDirty, message) {
+            var promise = $q.when();
+            if (isDirty) {
+                if (this.isEditable(diff)) {
+                    promise = confirm.confirmSaveWork(message)
+                        .then(angular.bind(this, function save() {
+                            return this.saveWork(orig, diff);
+                        }), function(err) { // cancel
+                            return $q.when();
```
petrjasek: it would always call `onsuccess` in:
```
saveWorkConfirmation().then(onsuccess);
```
even if there is an error. so there should be `$q.reject()` or you can just remove that error handler.

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -22,6 +34,9 @@
         ws.onmessage = function(event) {
             var msg = angular.fromJson(event.data);
             $rootScope.$broadcast(msg.event, msg.extra);
+            if (_.has(ReloadEvents, msg.event)) {
+                $rootScope.$broadcast('reload', msg);
```
petrjasek: I think it would be more explicit if you just inject reload service and call some method there, or are you expecting other consumers?

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -29,6 +44,102 @@
...
+                        }
+                    }
+                }
+            }
+            return result;
+        };
+    }
+
+    return angular.module('superdesk.notification', ['superdesk.desks'])
+        .service('reloadService', ReloadService)
```
petrjasek: imo there is no need to call it `reloadService`, `reload` should be fine (doesn't matter if it's service, factory etc)

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -10,8 +10,20 @@
 (function() {
     'use strict';
 
-    WebSocketProxy.$inject = ['$rootScope', 'config'];
-    function WebSocketProxy($rootScope, config) {
+    WebSocketProxy.$inject = ['$rootScope', 'config', 'desks'];
+    function WebSocketProxy($rootScope, config, desks, session) {
+        var ReloadEvents = {
+            'user_disabled': 'User is disabled',
```
petrjasek: those messages could use gettext

petrjasek:
:+1: looks good, just added a few hints

client/app/scripts/superdesk/notification/notification.js
sjunaid: actually, we tried here injecting reloadService but , lot of tests (more than 20) was starts failing, if you notice ozan's commit above as well for that prpose, so after that come up that solution.

client/app/scripts/superdesk/notification/notification.js
sjunaid: Since there are other variables and event using in this work with the name 'reload', that's why using 'reloadService' as service name here.

client/app/scripts/superdesk/notification/notification.js
sjunaid: done

client/app/scripts/superdesk-authoring/authoring.js
sjunaid: removed the handler.

ioanpocol:
:+1:

PR #708, opened 2015-06-11
fix(multi content config close): Fix closing button on multi content config modal

pavlovicnemanja92:
There is no task in Jira for, this is update for task releated to workspace multi content modal


ioanpocol:
:+1:

PR #709, opened 2015-06-11
[WIP] RSS ingest: create a package if RSS entry contains links to images

plamut:
This PR creates the actual content package from the RSS item and the images it references.

I would need to know though if the dictionary created in the `_wrap_into_package` method is correct, as well as how to resolve several TODOs I placed in it (hence the missing unit tests for this method).

Thanks for all the feedback in advance!

SD-2482

### DIFF
--- a/server/superdesk/io/rss.py
+++ b/server/superdesk/io/rss.py
@@ -61,6 +61,7 @@ class RssIngestService(IngestService):
     * type - field's data type
     """
 
+    # TODO: remove bmp, add tiff (to be consistent with media.renditions)
     IMG_MIME_TYPES = (
         'image/bmp',
         'image/gif',
@@ -114,17 +115,21 @@ class RssIngestService(IngestService):
             if t_entry_updated <= t_provider_updated:
                 continue
 
-            # If the entry references any images, create a package from it,
-            # otherwise treat it as a simple text item (even if it might
-            # reference other media types, e.g. video clips).
+            item = self._create_item(entry, field_aliases)
+            self.add_timestamps(item)
+
+            # If the entry references any images, create picture items from
+            # them and create a package containing them and the entry itself.
+            # If there are no image references, treat entry as a simple text
+            # item, even if it might reference other media types, e.g. videos.
             image_urls = self._extract_image_links(entry)
             if image_urls:
-                images = self._fetch_images(image_urls, provider)  # TODO: test is called
-                raise NotImplementedError(images)  # TODO: create a package
-            else:
-                item = self._create_item(entry, field_aliases)
-                self.add_timestamps(item)
+                image_items = self._create_image_items(image_urls, item)
+                new_items.extend(image_items)
                 new_items.append(item)
+                item = self._create_package(item, image_items)
+
+            new_items.append(item)
 
         return [new_items]
 
@@ -193,6 +198,7 @@ class RssIngestService(IngestService):
 
         return list(img_links)
 
+    # TODO: delete this, the downloads will be handled elsewhere
     def _fetch_images(self, image_urls, provider):
         """Fetch images from the given list of URLs.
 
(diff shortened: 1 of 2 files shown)


server/superdesk/io/rss.py
```diff
@@ -267,5 +268,47 @@ def _create_item(self, data, field_aliases=None):
...
+            package as picture items
+        :return: the created content package
+        :rtype: dict
+        """
+        package = {
+            'type': 'composite',
+            'groups': [
+                {
+                    'id': 'root',
+                    'role': 'grpRole:NEP',  # TODO: what group role here?
```
petrjasek: this is the right one

petrjasek:
imo this should first save all images as picture items and after that it should create package referencing those items, and not just embedding the images in a single package item. (that's what it does, right?)

plamut:
Yes, that's right. I create picture items from images (well, I want to at least) and then bundle these items with the text item into a single package and return it.

Are there any good examples on how to create a picture item from raw image data? Ideally there is some method for it that I can use here.

petrjasek:
actually we download images from [reuters](https://github.com/superdesk/superdesk/blob/master/server/superdesk/io/commands/update_ingest.py#L350-L355), so as long as you populate renditions in an item it will fetch those images

plamut:
Hm... so for each picture item I should create a renditions dictionary containing a single rendition (`'baseImage'`) with a link to the image and the rest of the stack will take care of it?

On a related note - if I understand the linked `ingest_item` function correctly, images are downloaded sequentially (each new picture item is processed individually). Since I already have all the images (downloaded in parallel), would it make sense to modify that function so that it will use the already downloaded image, if available?

amagdas:
I would say to use the existing mechanism, not download the images differently. That means, when parsing the RSS feed, just create the proper package dictionary and pass it to `ingest_items`, if there are any fixes or optimization we want to do to image download, we do it in only one place.

plamut:
OK, I'll make a ticket for optimizing the download process when we know in advance we want to fetch multiple files when building a package.

As for the package creation - is the following correct?
* invoke the `ingest_item` for each picture item so that they are created in DB and get IDs assigned to them (IDs are needed for the intra-package references, i.e. `residRef`, right?)
* create a package dictionary with a single "main" group referencing the picture items created in the previous step, and the text item containing the RSS entry's text
* invoke `ingest_item' with this package dictionary to create the package in DB

BTW, can these steps be merged into a single step by invoking `ingest_items` instead? Item URLs can be used for their GUIDs which can be used for `residRef` values within the package - would that work?

amagdas:
`ingest_items` is automatically called, so all that is needed is to return proper data from `provider.update`.

plamut:
OK, OK, but I need to know what "proper data" looks like in the first place. :) Can the `guid` field be used for intra-package references?

petrjasek:
yep guid is used in package to reference other items

plamut:
OK, thanks, will come up with something and re-submit for the review.

plamut:
Updated, please review again.

Basically, if there are image links, "empty" image items are created with each containing only a single "baseImage" rendition with a link to the corresponding image. The `firstcreated` and `versioncreated` fields are set to the same values as in the "main" text item.

Then, when we have both the text item and image items, they are all added to the new items list (returned by the `_update` method), and a package is created containing the references to all these items.

Is this the correct way of doing it?

server/superdesk/io/rss.py
```diff
@@ -114,17 +115,21 @@ def _update(self, provider):
...
-            # If the entry references any images, create a package from it,
-            # otherwise treat it as a simple text item (even if it might
-            # reference other media types, e.g. video clips).
+            item = self._create_item(entry, field_aliases)
+            self.add_timestamps(item)
+
+            # If the entry references any images, create picture items from
+            # them and create a package containing them and the entry itself.
+            # If there are no image references, treat entry as a simple text
+            # item, even if it might reference other media types, e.g. videos.
```
amagdas: Actually, video and audio content is treated exactly the same as images, the processing for all media types is unified when ingesting.

plamut: Yeah, I had the same thought when reading the specifications, since pictures can essentially be (and in fact, as you pointed out, _are_) treated in the same way as any other media file.
@hlmnrmr should we reconsider this and add support for other media types as well?

@amagdas Are the created item/package dictionaries correct otherwise? Nothing  to change?

petrjasek:
looks good. ps: there are some helper methods in [package service](https://github.com/superdesk/superdesk/blob/master/server/apps/packages/package_service.py) for package creation

plamut:
OK, thanks for confirming.

Too bad the p.s. came a bit late, but thanks for mentioning it nevertheless, note taken. :)

server/superdesk/io/rss.py
hlmnrmr: Other media types should go as a separate issue, the scope of this one is strictly support for images, any extra effort no matter how trivial means solving a different problem.

PR #710, opened 2015-06-12
(fix) Eve : Modified _version field to _current_version to avoid clash with elasticseach field

mdhaman:
[SD-2525] Modified _version field to _current_version to avoid clash with elasticseach _version field

Reason: ElasticSearch and Eve both use _version field for version information. There is no issue when we try to create/update document. But, when we try to bulk insert into elasticsearch using app:index_from_mongo command the operation fails due to version conflict error. 

After merging this PR, 
- please drop elasticsearch "Superdesk" index. 
- Remove all document from ingest, archive, published, publish_queue, formatted_item, archive_versions, archive_autosave.


sivakuna-aap:
:+1:

amagdas:
:+1:

ioanpocol:
:+1:

mdhaman:
I will have to make changes to EVE. We have following in the GET method.
    if req.if_none_match:
        if (resource_def['versioning'] is False) or \
           (document['_version'] == document['_latest_version']):

Don't know why behave tests passed.

amagdas:
:+1:

ioanpocol:
:+1:

PR #711, opened 2015-06-12
chore(scaffold) - add scaffolding command

amagdas:


server/apps/legal_archive/resource.py
```diff
@@ -18,41 +18,29 @@
...
 
-class LegalArchiveResource(Resource):
-    endpoint_name = LEGAL_ARCHIVE_NAME
+class LegalResource(Resource):
     schema = {}
     resource_methods = ['GET']
     item_methods = ['GET']
-    resource_title = endpoint_name
-    internal_resource = True
+    privileges = {'GET': LEGAL_ARCHIVE_NAME}
```
sivakuna-aap: This may not work as SuperdeskTokenAuth.check_permissions() returns True if HTTP Method is 'GET'.

Please check https://github.com/superdesk/superdesk/blob/master/server/apps/auth/auth.py#L94

amagdas: Oh, thanks for catching this :+1:

server/apps/prepopulate/app_scaffold_data.py
```diff
@@ -0,0 +1,71 @@
...
+        for i, desk in enumerate(desks):
+            self.logger.info('Adding items for desk:' + str(desk['_id']))
+            self.ingest_items_for(desk, no_of_stories, i + 1)
+
+    def ingest_items_for(self, desk, no_of_stories, skip_index):
+        desk_id = desk['_id']
+        stage_id = desk['incoming_stage']
+
+        min_size = 100
+        bucket_size = min_size if min_size < no_of_stories else no_of_stories
```
petrjasek: `max(min_size, no_of_stories)`

server/apps/prepopulate/app_scaffold_data.py
```diff
@@ -0,0 +1,71 @@
...
+
+        no_of_buckets = len(range(0, no_of_stories, bucket_size))
+
+        for x in range(0, no_of_buckets):
+            skip = x * bucket_size * skip_index
+            print('Page : {}, skip: {}'.format(x + 1, skip))
```
petrjasek: maybe it could use logger?

petrjasek:
:+1:

plamut:
Missing test, but otherwise good IMO.

amagdas:
I don't think it's worth mocking so much stuff, I would rather do an integration test but for that we need text_archive to be around.

akintolga:
:+1:

PR #713, opened 2015-06-12
Each MediaStorage should be able to choose the way they return links

vied12:


server/superdesk/upload.py
```diff
@@ -37,11 +37,14 @@ def get_upload_as_data_uri(media_id):
...
-        response.cache_control.max_age = cache_for
-        response.cache_control.s_max_age = cache_for
+        try:
+            response.content_length = media_file.length
+            response.last_modified = media_file.upload_date
+            response.set_etag(media_file.md5)
+            response.cache_control.max_age = cache_for
+            response.cache_control.s_max_age = cache_for
+        except Exception:
+            pass
```
plamut: Do we really want to silence the exception?

plamut:
Test is missing. ;)

vied12:
S3 feature isn't tested because it requires a valid account.
And if you provide encrypted credentials to travis, you will loose the ability to  run tests for PR from other repos than superdesk/superdesk

plamut:
Why? Just mock it. :)

mugurrus:
:+1:  A test would be nice although it was the responsibility of the person that implemented Amazon storage in the first place :)

vied12:
Right, but I can add a unit test for the new method

ioanpocol:
:+1:

PR #714, opened 2015-06-12
[SD-2514]  Removing associations while removing expired articles which are spiked

sivakuna-aap:


server/apps/archive/archive.py
```diff
@@ -412,6 +413,78 @@ def can_edit(self, item, user_id):
...
+                if str(package[config.ID_FIELD]) not in processed_packages:
+                    processed_packages.extend(self.remove_refs_in_package(package, doc_id, processed_packages))
+
+    def get_packages(self, doc_id):
+        """
+        Retrieves if an article identified by doc_id is referenced in a package.
+        :return: articles of type composite
+        """
+
+        query = {'query': {'filtered': {'filter': {'and': [{'term': {'type': 'composite'}}]},
```
amagdas: I think we should fetch the data from Mongo in all the background jobs.
The spiked document already has a reference to all the packages in which is it included into, so we should be able to fetch the packages from Mongo by `_id`.

server/apps/archive/archive.py
```diff
@@ -412,6 +413,78 @@ def can_edit(self, item, user_id):
...
+        for group in new_groups:
+            if group['id'] == 'root':
+                group['refs'] = new_root_refs
+                break
+
+        updates = {config.LAST_UPDATED: utcnow(), 'groups': new_groups}
+        resolve_document_version(updates, SOURCE, 'PATCH', package)
+
+        self.patch(package[config.ID_FIELD], updates)
+        insert_into_versions(id_=package[config.ID_FIELD])
```
amagdas: Maybe we should move this into PackagesService for this to happen on all package changes.

mdhaman: Agree with @amagdas as it would help me with handling of spikes in takes package.

mdhaman:
:+1:

amagdas:
:+1:

PR #717, opened 2015-06-15
[SD-2490] Create a NewsML G2 Output Format

marwoodandrew:


sivakuna-aap:
Looks good. It would be good if we can externalize `AAP_PROVIDER_ID` also.

marwoodandrew:
Yes, I think it should become just PROVIDER_ID and move into the settings, there is nothing AAP specific in the formatter then.

ioanpocol:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

PR #719, opened 2015-06-16
(feat) SD-1637 - Messaging (Backend)

sivakuna-aap:
- [x] [SD-2426]  Backend: Create the chat resource
- [x] [SD-2425] Send a notification when a new user is added to a chat session
- [x] [SD-2427] Remove expired chat sessions

### DIFF
--- a/server/apps/messaging/messages.py
+++ b/server/apps/messaging/messages.py
@@ -7,7 +7,6 @@
 # For the full copyright and license information, please see the
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
-from eve.utils import config
 
 from apps.archive.common import get_user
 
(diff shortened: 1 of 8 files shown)


server/apps/messaging/messages.py
```diff
@@ -24,18 +39,43 @@ class MessageService(Service):
...
             sent_user = doc.get('sender', None)
-            user = g.user
-            if sent_user and sent_user != str(user.get('_id')):
+            user = get_user()
+            if sent_user and str(sent_user) != str(user.get('_id')):
                 message = 'Sending a message on behalf of someone else is prohibited.'
                 raise SuperdeskApiError.forbiddenError(message)
             doc['sender'] = str(user.get('_id'))
 
+            chat_session = get_resource_service(CHAT_SESSIONS).find_one(req=None, _id=doc['chat_session'])
```
amagdas: This is already handled by Eve as specified in the schema.

amagdas:
When fetching a chat session we should return the denormalized list of users, this is needed by UI.

server/apps/messaging/messages.py
sivakuna-aap: Ah! right. Forgot about that.

server/apps/messaging/messages_test.py
```diff
@@ -0,0 +1,52 @@
...
+    def setUp(self):
+        super().setUp()
+
+        with self.app.app_context():
+            self.app.data.insert('users', self.users)
+            self.app.data.insert('groups', self.groups)
+            self.app.data.insert('desks', self.desks)
+
+    def test_resolve_recipients(self):
+        with self.app.app_context():
```
plamut: If more tests are added that need this `with self.app.app_context()` line, it might make sense to push the application context in `setUp()` and pop it in `tearDown()`, avoiding the need for repeating that line in every test method.

(not that it's an issue here, but I noticed this pattern spreading several other tests...)

server/settings.py
```diff
@@ -124,6 +124,10 @@ def env(variable, fallback_value=None):
         'task': 'apps.publish.content_purge',
         'schedule': crontab(minute=30)
     },
+    'messaging:remove_expired': {
+        'task': 'apps.messaging.purge_chat_sessions',
+        'schedule': crontab(hour=0, minute=0)
```
amagdas: We will have to provide some time here, maybe 24 hours.

sivakuna-aap: In Cron Expression all intervals start from 0. In this case, the job runs every day midnight 1200 AM.

amagdas: :+1: Thanks for put shed some light on my ignorance :smile:

PR #733, opened 2015-06-17
fix(docker): fix unittests for publicapi in docker

petrjasek:
it was missing config with prefix which works fine on localhost,
but not when mongo is running in a container

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -45,8 +45,8 @@ backend:
    - logstash
    - postfix
   environment:
-   - MONGOLAB_URI=mongodb://mongodb:27017/test
-   - LEGAL_ARCHIVE_URI=mongodb://mongodb:27017/test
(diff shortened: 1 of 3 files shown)


plamut:
:+1: 

Just curious, though, why Docker environment is/was different from e.g. the Travis builds? Is it possible that some other things are different, too?

IIRC we've already had a few cases when things worked, but not when deployed with Docker due to some differences in the environment.

petrjasek:
on travis you have mongo on `localhost` (and that's default setting so it was working), but in docker it's on mongodb container ip address so it has to read it from env

plamut:
Ah, makes sense.

Are you perhaps aware of any other differences that might exist between our "localhost" builds and Docker builds? (we should bridge as much of these as possible)

petrjasek:
afaik that should be it

sivakuna-aap:
:+1:

mugurrus:
:+1:

PR #734, opened 2015-06-17
fix(editor): after find/replace trigger model update

petrjasek:
make editor listen to blur events and not only input


plamut:
:+1:

mugurrus:
:+1:

PR #735, opened 2015-06-17
Add more metadata to RSS-ingested package (versioncreated, guid, etc.)

plamut:
It turned out that RSS packages lacked metadata and were filtered out because they were recognized as expired, and they also lacked the `guid` attribute. This PR fixes that.

### DIFF
--- a/server/superdesk/io/rss.py
+++ b/server/superdesk/io/rss.py
@@ -299,6 +299,9 @@ class RssIngestService(IngestService):
         not the items themselves. In the list of references, the reference to
         the text item preceeds the references to image items.
 
+        Package's `firstcreated` and `versioncreated` fields are set to values
+        of these fields in `text_item`.
+
         :param dict text_item: item representing the text content
         :param list image_items: list of items (dicts) representing the images
             related to the text content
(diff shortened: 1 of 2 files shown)


server/superdesk/io/rss.py
```diff
@@ -307,6 +310,9 @@ def _create_package(self, text_item, image_items):
         """
         package = {
             'type': 'composite',
+            'guid': 'package:' + text_item['guid'],
```
plamut: Is this an OK guid for a package?

amagdas: IMO we should use this: https://github.com/superdesk/superdesk/blob/master/server/apps/archive/common.py#L88

mugurrus:
:+1: but use what Adrian suggested for guid generation

server/superdesk/io/rss.py
plamut: This is exactly what I was looking for, thanks. I'll update.

plamut: BTW, what to use for the "hint" to the linked function, simply the GUID_TAG from the settings?

plamut:
Updated, `generate_guid()` is now used for items' and packages' `guid` field values.

amagdas:
:+1:

PR #736, opened 2015-06-18
(fix) [SD-2542]

sivakuna-aap:
Issues fixed:

1. No message is shown indicating why the desk containing sages has not been deleted. There can be no desk without a stage as incoming stage is mandatory. Originally, as per the requirement user should be able to delete if the desk doesn't have any content.
2. When there are more users in the system and desk as well the system takes time in loading the list. While loading if user clicks on Next the system will send request payload as members: [] resulting in 0 members in the desk although there were users in the desk initially. Same case can occur in case of stages as well.


amagdas:
:+1:

mugurrus:
:+1:

mdhaman:
:+1:

PR #738, opened 2015-06-18
[SD-2426] De-normalizing recipients on every create and update.

sivakuna-aap:
This adds capability to find chat sessions which a user is participating.


plamut:
:+1: 

(somebody just please re-run the build, that `multi mark for highlights` test failed again, it happens occasionally...)

amagdas:
:+1:

PR #741, opened 2015-06-18
Recent templates

eozan:
Tests and fix for broken tests coming soon.

### DIFF
--- a/client/app/scripts/superdesk-templates/templates.js
+++ b/client/app/scripts/superdesk-templates/templates.js
@@ -69,6 +69,11 @@
 
             return api.content_templates.query(params)
             .then(function(result) {
+                if (result && result._items) {
+                    result._items.sort(function(a, b) {
+                        return templateIds.indexOf(a._id) - templateIds.indexOf(b._id);
+                    });
+                }
                 return result;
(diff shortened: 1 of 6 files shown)


plamut:
Tests should be added, otherwise it seems fine.
(and extra points if docstrings are added ;) )

client/app/scripts/superdesk-templates/templates.js
```diff
@@ -54,6 +55,57 @@
...
+
+            var criteria = [];
+            _.each(templateIds, function(id) {
+                criteria.push({_id: id});
+            });
+
+            var params = {
+                max_results: PAGE_SIZE,
+                page: 1,
+                where: JSON.stringify({'$or': criteria})
```
petrjasek: fyi there is also `{_id: {$in: templateIds}}` query

petrjasek:
looks fine, just wondering are those last used templates ordered somehow?

eozan:
Last used template is added to the top of the list, That's the only ordering used here.

petrjasek:
:+1:

ioanpocol:
:+1:

PR #742, opened 2015-06-18
auth.js: use require.toUrl to set the reset password template

vied12:


plamut:
:+1:

ioanpocol:
One unittest is failing:
PhantomJS 1.9.8 (Linux) auth can use routes with auth=false without identity FAILED

plamut:
Yeah, unexpected GET request due to the changed template URL, the latter now includes the full path (starting from `/base/app...`).

vied12:
I would like to use `superdesk.asset` but at this point the `config` isn't set...
Do you have any clue ?

PR #745, opened 2015-06-19
feat(monitoring): Add the option to set the max items per group

ioanpocol:


mugurrus:
:+1:  aside from the following fixes:
- replace iit with it in multicontent_spec.js
- fix selection in setMaxItems() from helpers/multicontent.js

amagdas:
:+1:

PR #749, opened 2015-06-22
Legal archive UI

amagdas:


server/apps/legal_archive/resource.py
```diff
@@ -18,41 +19,30 @@
...
 
-class LegalArchiveResource(Resource):
-    endpoint_name = LEGAL_ARCHIVE_NAME
+class LegalResource(Resource):
     schema = {}
     resource_methods = ['GET']
     item_methods = ['GET']
-    resource_title = endpoint_name
-    internal_resource = True
+    privileges = {'GET': LEGAL_ARCHIVE_NAME}
```
sivakuna-aap: This privilege will not work as by default `GET` is allowed for all users. For more information, please look https://github.com/superdesk/superdesk/blob/master/server/apps/auth/auth.py#L94

amagdas: Thanks, forgot about it, will fix.

akintolga:
:+1:

mdhaman:
:+1:

PR #755, opened 2015-06-23
Fix privilege check for Legal Archive

amagdas:


mdhaman:
:+1:

server/apps/legal_archive/service.py
```diff
@@ -8,10 +8,45 @@
...
+
+    def get(self, req, lookup):
+        self.check_get_access_privilege()
+        return super().get(req, lookup)
+
+    def check_get_access_privilege(self):
+        if not hasattr(g, 'user'):
+            return
+
+        privileges = g.user.get('privileges', {})
```
sivakuna-aap: `auth.py` sets `active_privileges` on `g.user` not `privileges`. So privileges variable will always be {}. Please refer https://github.com/superdesk/superdesk/blob/master/server/apps/users/services.py#L296

sivakuna-aap:
:+1:

PR #756, opened 2015-06-23
feat(dictionary): add word to user dict in authoring

petrjasek:
user can add word that is not part of global dictionaries
into their personal dictionaries. there can be one dictionary
per user and language.


sivakuna-aap:
:+1:

petrjasek:
now together with `active` switch for global dictionaries

amagdas:
:+1:

PR #757, opened 2015-06-23
Automate the registration of the ingest classes as providers

plamut:
This is now handled by a metaclass, which also checks across all ingest classes for any duplicate `PROVIDER˙ names defined. The behavior of this metaclass is described in more details in its docstring.

Resolves [SD 2535](https://dev.sourcefabric.org/browse/SD-2535).

### DIFF
--- a/server/app.py
+++ b/server/app.py
@@ -26,6 +26,7 @@ from superdesk.storage.desk_media_storage import SuperdeskGridFSMediaStorage
 from superdesk.validator import SuperdeskValidator
 from raven.contrib.flask import Sentry
 from superdesk.errors import SuperdeskError, SuperdeskApiError
+from superdesk.io import providers
 from logging.handlers import SysLogHandler
 from settings import LOG_SERVER_ADDRESS, LOG_SERVER_PORT
 
@@ -117,6 +118,11 @@ def get_app(config=None):
 
     app.sentry = sentry
     sentry.init_app(app)
+
+    # instantiate registered provider classes (leave non-classes intact)
+    for key, provider in providers.items():
+        providers[key] = provider() if isinstance(provider, type) else provider
+
     return app
 
 if __name__ == '__main__':
--- a/server/superdesk/io/ingest_service.py
+++ b/server/superdesk/io/ingest_service.py
@@ -14,12 +14,62 @@ import superdesk
 from datetime import datetime
 from superdesk.utc import utc, utcnow
 from superdesk.errors import SuperdeskApiError, SuperdeskIngestError
+from superdesk.io import register_provider
 
 
 logger = logging.getLogger(__name__)
 
 
-class IngestService():
+class AutoRegisteredMeta(type):
+    """Metaclass for automatically registering the ingest service instances.
+
+    An instance of an ingest class is registered as a provider when the class
+    is instantiated - but only if the class defines a `PROVIDER` attribute
+    containing the name under which to register the class instances.
+
+    Every ingest provider class needs to define an `ERRORS` attribute
+    representing a list of <error_number, error_message> pairs that might be
+    raised by the class instances' methods.
+
+    If another provider class with the same `PROVIDER` name has already been
+    defined, a TypeError is raised.
+    """
+    _providers = {}  # keep track of all PROVIDER names defined in classes
+
+    def __new__(metacls, name, bases, attrs):
+        provider_name = attrs.get('PROVIDER')
+
+        if provider_name is not None:
+            if 'ERRORS' not in attrs:
+                raise AttributeError(
+                    "Provider class {} must define "
+                    "the ERRORS list attribute.".format(name))
+
+            if provider_name in metacls._providers:
+                raise TypeError(
+                    "PROVIDER {} already exists (class {}).".format(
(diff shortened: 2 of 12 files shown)


petrjasek:
:+1: looks good.
would make sense to do this also for resource/service/.. registration later if we want to use metaclasses to make things consistent.

plamut:
I'll gladly do that as well, could you create a ticket in the backlog?

Because I agree, registering a family of classes is one of those cases where using a metaclass actually makes a lot of sense.

server/superdesk/io/ingest_service.py
```diff
@@ -14,12 +14,62 @@
...
+    def __call__(cls, *args, **kwargs):
+        # NOTE: Registration cannot be done in __new__ or __init__, because the
+        # class cannot be instantiated there yet (a RuntimeError occurs), thus
+        # we need to perform the registration here in __call__ (when a new
+        # class instance is going to be created).
+
+        new_instance = super().__call__(*args, **kwargs)
+
+        if hasattr(cls, 'PROVIDER'):
+            register_provider(cls.PROVIDER, new_instance, cls.ERRORS)
```
ioanpocol: The __call__ method will run only when a provider class will be instantiated so a provider will be registered only when you create an instance of that provider. 
Where are you creating the instances of the providers?

plamut: They should be instantiated at application startup?

The problem is that is not possible to create instances of the just-created class in `__new__()` or `__init__()` if that class invokes `super().__init__()` in its own `__init__()` method, a `RuntimeError` occurs. See [this discussion](http://stackoverflow.com/questions/13126727/how-is-super-in-python-3-implemented) for more details. This behavior is considered a bug in the Python interpreter by some people (I'm using CPython 3.3.6 on Ubuntu).

Before sending this PR, I circumvented the problem by registering the ingest classes on their instantiation, but if that's not OK, I can revert back to registering the classes in metaclass'es `__new__()` - if we can live with the limitation that `PROVIDER` ingest classes will not be able to call `super().__init__()` in their own `__init__()` method (FWIW, the former doesn't do anything really, since the base `IngestService` class doesn't have its own `__init__()` method).

petrjasek: missed that. atm it's registering instances not classes. there is no need to have instances, those can be created when needed, but it needs to register provider name - it's used for allowed_values check in ingest_provider resource

plamut: Yeah, I was wondering that, too, why an instance is needed when registering a new provider, and not just the class. But refactoring that (to create an instance if/when needed) would probably mean too much work?

The drawback I described in the 2nd paragraph in my previous comments is probably not a deal breaker? And, in fact, it was only the `RssIngestService` that invoked `super().__init__()` causing the error on registration, and it works just as good without it.

I'm also posting a question on StackOverflow for any additional hints...

ioanpocol: __init__ and __new__ are called at class creation.
So you can change __init__ where you call the register and send the class object to it.
When UpdateIngest is started you can create a provider instance of every provider if it is not already created.

petrjasek: imo it's not a deal breaker, and it's less work than modifying the code to create instances when needed

server/superdesk/io/commands/update_ingest.py
```diff
@@ -151,7 +151,19 @@ class UpdateIngest(superdesk.Command):
...
+        """Return an instance of `thing` if `thing` is a class, otherwise
+        return the `thing` itself.
+        """
+        return thing() if isinstance(thing, type) else thing
+
     def run(self, provider_type=None):
+        # instantiate any registered providers classes that have not been
+        # instantiated yet
+        for name, provider in providers.items():
+            providers[name] = self.instantiate_if_class(provider)
```
plamut: What's the best place to instantiate the registered provider classes? To me this does not seem like the best place to do it. Is there some kind of an application "init()" that is run when the application is started?

server/app.py
```diff
@@ -117,6 +118,11 @@ def server_error_handler(error):
 
     app.sentry = sentry
     sentry.init_app(app)
+
+    # instantiate registered provider classes (leave non-classes intact)
+    for key, provider in providers.items():
+        providers[key] = provider() if isinstance(provider, type) else provider
```
plamut: Is this indeed a good place for instantiating all registered ingest classes, i.e. when the application object is created?

sivakuna-aap:
Needs rebase. Build fails as python3-ldap 0.9.7 is not available anymore.

server/app.py
ioanpocol: The initialization is needed also on celery because on update_provider(celery task) you access the list of providers.
Better to create a method on superdesk.io, and the method returns the providers by name. In the method if provider is not created, you can create it.

ioanpocol:
:+1:  (except init providers and rebase)

server/app.py
petrjasek: celery is using app from factory, this should work

plamut:
I've rebased. I left the instantiating intact, following @petrjasek's comment.

amagdas:
:+1:

PR #770, opened 2015-06-25
Fix (mentioned users scroll and modal window hide animation)

Aca-jov:
SD-744 && SD-2098

### DIFF
--- a/client/app/scripts/superdesk-users/users.js
+++ b/client/app/scripts/superdesk-users/users.js
@@ -1297,6 +1297,15 @@
                     scope.selectUser = function(user) {
                         return '@' + user.username;
                     };
+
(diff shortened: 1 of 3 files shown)


client/app/scripts/superdesk-users/users.js
```diff
@@ -1297,6 +1297,15 @@
                     scope.selectUser = function(user) {
                         return '@' + user.username;
                     };
+
+                    scope.$watchCollection(
```
petrjasek: not sure it's a good idea to use jquery in a watcher, is there maybe some js event it could listen to?

petrjasek:
:+1:

ioanpocol:
:+1:

PR #767, opened 2015-06-24
feat monitoring

petrjasek:
add monitoring view, it's a new page (like content) in the workspace,
showing same content like monitoring widget (former aggregate).

PR #771, opened 2015-06-25
feat(publish): Implemented publish queue preview sidebar.

eozan:
SD-2375

### DIFF
--- a/client/app/scripts/superdesk-publish/views/publish-queue.html
+++ b/client/app/scripts/superdesk-publish/views/publish-queue.html
@@ -104,28 +104,28 @@
                 </div>
             </div> 
        </div>
-    </div>
-    <div class="preview-pane content-item-preview" ng-init="tab = 'content'">
-        <header>
-            <ul class="nav nav-tabs">
-                <li ng-class="{active: tab === 'content'}">
-                    <button ng-click="tab = 'content'"><span translate>Content</span></button>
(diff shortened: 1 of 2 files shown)


ioanpocol:
:+1:  (a behave test will help)

client/app/scripts/superdesk-publish/views/publish-queue.html
```diff
@@ -105,5 +105,28 @@ <h3 class="page-nav-title" ng-hide="multiSelectCount > 0" translate>Publish Queu
             </div> 
        </div>
     </div>
+    <div class="preview-pane content-item-preview" ng-init="tab = 'content'">
```
petrjasek: you can use `sd-item-preview` directive it was added in #767

eozan:
Changed to use new sd-item-preview directive.

ioanpocol:
:+1:

plamut:
Seems fine, but please add the test(s), too.

sivakuna-aap:
:+1: Might need rebase

ioanpocol:
Per Ozan feedback he will add the tests in a different PR

plamut:
I don't think this is the right approach, tests should be shipped along with the code so that it is always possible to automatically test a new commit. It kind of becomes tempting to postpone that separate PR with tests to "later"...

PR #785, opened 2015-06-30
Fix (order of right and left items is fixed)

Aca-jov:
SD-2467


plamut:
:+1:  (just check whether the failing behave test is just a random occurrence or something that needs to be fixed before merging)

ioanpocol:
:+1:

PR #787, opened 2015-07-01
undo/redo fix for medium editor

sjunaid:


client/app/scripts/superdesk/editor/editor.js
```diff
@@ -23,6 +23,14 @@ function EditorService() {
 
     var vm = this;
 
+    this.historyBodyHtml = [];
+    this.historyHeadline = [];
+    this.historyAbstract = [];
+
+    this.index = -1;
+    this.indexHeadline = -1;
+    this.indexAbstract = -1;
```
plamut: It probably makes sense to use an alias for `this` (i.e. `vm`) if we already defined it IMO.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -360,9 +389,17 @@ angular.module('superdesk.editor', [])
...
                     editorElem.on('input blur', function(event) {
                         $timeout.cancel(updateTimeout);
-                        updateTimeout = $timeout(updateModel, 300, false);
+                        updateTimeout = $timeout(function() { updateModel(event); }, 300, false);
+                    });
+                    editorElem.on('keyup', function(event) {
+                        if (event.ctrlKey) {
+                            if (event.keyCode === 90) {
+                                doUndo();
+                            } else if (event.keyCode === 89) {
```
plamut: I suggest creating an alias for these key codes for better readability, perhaps something like the following example:
```js
var KEY_CODES = Object.freeze({
    Y: 89,
    X: 90
});

if (event.keyCode === KEY_CODES.X) { ... }
```

I would also consider putting the key codes definition into an Angular constant or something, because they are used at multiple places throughout the app...

plamut:
Tests are missing, but otherwise this seems promising.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -21,6 +21,19 @@ function EditorService() {
     this.addEventListeners = addEventListeners;
     this.removeEventListeners = removeEventListeners;
 
+    // Undo Redo related variables
+    this.historyBodyHtml = [];
```
amagdas: This can be refactored to a dict where the key is the field name, this will allow to support undo/redo automatically on all the places where the editor is used.
The map would look like `{'headline': {'stack': [], 'index': 1}}`

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -426,15 +468,124 @@ angular.module('superdesk.editor', [])
                 scope.$on('spellcheck:run', renderSpellcheck);
                 scope.$on('key:ctrl:shift:s', renderSpellcheck);
 
-                function updateModel() {
+                function doUndo() {
+                    scope.$applyAsync(function() {
+                        if (elem[0].id === 'bodyhtml') {
```
amagdas: Using the approach I mentioned above, this code duplication will be removed.

akintolga:
It is all good. My only opinion is that Undo implementation is inherently an implementation of a stack (last-in first-out) so maybe in next iteration (or next visit to this feature) that can be refactored. But I don't think it is a show stopper.

amagdas:
:+1:

amagdas:
:+1:

sivakuna-aap:
:+1:

PR #788, opened 2015-07-01
fix(dictionaries): On add new dictionary fails with 400 error code

ioanpocol:


client/app/scripts/superdesk-dictionaries/dictionaries.js
```diff
@@ -63,7 +63,15 @@
...
 
         this.update = function update(dictionary, data, success, error) {
-            return api.save('dictionaries', dictionary, data).then(success, error);
+            var sendData = {};
+            angular.forEach(data, function(val, key) {
+                if (key !== 'content') {
+                    sendData[key] = val === null? val: val.toString();
+                } else {
+                    sendData[key] = val;
+                }
```
plamut: This can be shortened to a one-liner:
```js
sendData[key] = (key !== 'content' && val !== null) ? val.toString() : val;
```

ioanpocol: Yes, it can be this way, but for clarity I prefer the current version.

client/app/scripts/superdesk-dictionaries/dictionaries.js
```diff
@@ -143,7 +151,7 @@
                 $scope.origDictionary = result;
                 $scope.dictionary = _.create(result);
                 $scope.dictionary.content = _.create(result.content || {});
-                $scope.dictionary.is_active = $scope.dictionary.is_active !== false;
+                $scope.dictionary.is_active = $scope.dictionary.is_active !== 'false';
```
plamut: Wouldn't it be better to convert the `is_active` flag to a boolean when data is fetched to the server? It's more error-prone if it stays as a string, e.g. somebody might do:

```js
if (!$scope.dictionary.is_active) {...}
```

...which would not execute because `!'false'` would evaluate to `false` instead of the expected `true`.

ioanpocol: This was the issue, on upload file the boolean is interpreted as string on server and not as a boolean.
So changed the data to be string instead of boolean.

server/apps/dictionaries/resource.py
```diff
@@ -38,8 +38,8 @@ class DictionariesResource(Resource):
         },
         'user': Resource.rel('users', nullable=True),
         'is_active': {
-            'type': 'boolean',
-            'default': True,
+            'type': 'string',
+            'default': 'true',
```
plamut: This is related to the comment above. Wouldn't it be better to leave this as a boolean and perform the conversion to a string only when needed? (e.g. when sending a query to the backend)

It's more natural to work with booleans for properties that are, well, "naturally" boolean.

ioanpocol: See above comment.

client/app/scripts/superdesk-dictionaries/dictionaries.js
plamut: Hm, clarity was actually the reason for proposing a change. With `if` you have two different paths where `sendData[key]` can be set to `val`, which is redundant and hence a bit less readable IMO.

ioanpocol: IMO your proposal is not so readable because the statement is too long and with if statement is more easier to  understood to code.

client/app/scripts/superdesk-dictionaries/dictionaries.js
plamut: Hm... I think this should be handled on the server then, if it is not too much work. Having semantically "incorrect" data types makes it harder to work with such objects elsewhere...

ioanpocol: On server it is much more harder because the type checking is performed automatically by eve.

client/app/scripts/superdesk-dictionaries/dictionaries.js
plamut: It's not about the length per se, it's about how many execution paths you have to walk through in your head before you deduce the value of `sendData[key]` in all four possible cases.

As I said, to me personally at least a compact version is more readable, because it makes it very explicit in which case `val.toString()` is used and that `val` is used in all other cases.

client/app/scripts/superdesk-dictionaries/dictionaries.js
plamut: Oh, OK then, I guess we'll have to live with it then. 

As it stands, there's probably no good way of transparently doing the `bool` to `str` (and vice versa) conversions? E.g. when data enters/leaves the application boundary?

client/app/scripts/superdesk-dictionaries/dictionaries.js
ioanpocol: There are just two paths and I think that this is really readable. Indeed multiple imbricated "if" statements can become hard to read. But is not the case. And not always compact is better, it depends.

plamut: OK, I guess we simply read things differently, can't argue that.
(FWIW, I mentally "see" a nested _if-then-else_ inside the outer `if`, while the compact form reads just as a single _if-then-else_).

ioanpocol: Yes, this is better

ioanpocol: or even better
sendData[key] = (key !== 'content' && val !== null) ? val.toString() : val;

plamut: Erm, Ioan, but this is exactly what was proposed in the first comment. :wink:

ioanpocol: Yes, my mistake

mugurrus:
:+1:  (The string/bool issue is related to Eve and needs an Eve fix, after that is_active can be changed to bool)

client/app/scripts/superdesk-dictionaries/dictionaries.js
ioanpocol: The change is included in the next PR, the one related personal dictionaries.

PR #789, opened 2015-07-01
fix(media items): Fix link for preview of media icons from package

pavlovicnemanja92:
SD-1959

### DIFF
--- a/client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
+++ b/client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
@@ -17,7 +17,7 @@
             <div ng-hide=":: item.headline || item.slugline || item.description" translate>Blank headline received</div>
         </div>
 
-        <a ng-href="#/{{ :: type !== 'composite' ? 'authoring' : 'packaging'}}/{{ :: item.guid }}/view?_id={{ :: item.guid }}" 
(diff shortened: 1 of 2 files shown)


client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
```diff
@@ -17,7 +17,7 @@
             <div ng-hide=":: item.headline || item.slugline || item.description" translate>Blank headline received</div>
         </div>
 
-        <a ng-href="#/{{ :: type !== 'composite' ? 'authoring' : 'packaging'}}/{{ :: item.guid }}/view?_id={{ :: item.guid }}" 
+        <a ng-href="#/{{ :: type !== 'composite' ? 'authoring' : 'packaging'}}/{{ :: type !== 'video' || 'picture' ? item.residRef : item.guid}}/view?_id={{ :: item.guid }}"
```
plamut: This always evaluates to a truthy value (either `true`or `'picture'`).

pavlovicnemanja92: It works in this way if it is picture or video it results item.residRef and if it is neither it wil result item.guid. In my test example, where I have picture, video and text it works in this way.

pavlovicnemanja92: Sorry, you were right :D ... When I was writing I was accidentally added !== instead of === ... But I found that for text elements item.residRef is equal to item.quid ... it is different just for picture and video items... So I synchronized code now without condition.

plamut:
:+1:

amagdas:
:+1:

PR #791, opened 2015-07-02
(imp) SD-2579

sivakuna-aap:
[SD-2579] SPIKE - Investigate output channels and subscriber design
[SD-2608] Modify the Subscriber Endpoint and Management Screen to meet new business requirement
[SD-2609] Changes to Authoring UI	 Technical task
[SD-2611] Modify the publishing workflow
[SD-2643] Error 400 thrown on adding new subscriber when SEQUENCE NUMBER SETTINGS are not specified

`Need to run vocabularies command after the PR is merged`

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
@@ -1,6 +1,7 @@
 <div class="dropdown" dropdown>
 	<div class="dropdown-toggle" dropdown-toggle>{{ item[field].name || item[field]}} </div>
 	<ul class="dropdown-menu pull-right">
-		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled = "disabled">{{ o.name }}</button></li>
+		<li ng-click="select()"><button>None</button></li>
+		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled="disabled">{{ o.name }}</button></li>
 	</ul>
 </div>
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
@@ -1,7 +1,7 @@
 <div class="dropdown" dropdown>
 	<div class="dropdown-toggle" dropdown-toggle>{{ item[field].name || item[field]}} </div>
 	<ul class="dropdown-menu pull-right">
-		<li ng-click="select()"><button>None</button></li>
-		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled="disabled">{{ o.name }}</button></li>
+		<li ng-click="select()" translate><button>None</button></li>
+		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled="disabled">{{::o.name}}</button></li>
 	</ul>
 </div>
--- a/server/apps/content.py
+++ b/server/apps/content.py
@@ -8,6 +8,7 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
+from collections import namedtuple
 
 import superdesk
 from superdesk.resource import Resource
@@ -21,6 +22,9 @@ ITEM_TYPE_COMPOSITE = 'composite'
 LAST_TAKE = 'last_take'
 
 not_analyzed = {'type': 'string', 'index': 'not_analyzed'}
+
+pub_status = ['usable', 'withhold', 'canceled']
+PUB_STATUS = namedtuple('PUBSTATUS', ['USABLE', 'HOLD', 'CANCELED'])._make(pub_status)
 
 metadata_schema = {
     # Identifiers
@@ -186,8 +190,8 @@ metadata_schema = {
     },
     'pubstatus': {
         'type': 'string',
-        'allowed': ['usable', 'withhold', 'canceled'],
-        'default': 'usable',
+        'allowed': pub_status,
+        'default': PUB_STATUS.USABLE,
         'mapping': not_analyzed
     },
     'signal': {
--- a/server/apps/content.py
+++ b/server/apps/content.py
@@ -94,6 +94,7 @@ metadata_schema = {
     # Category Details
     'anpa-category': {
         'type': 'dict',
+        'nullable': True,
         'mapping': {
             'type': 'object',
             'properties': {
--- a/server/apps/content.py
+++ b/server/apps/content.py
@@ -24,7 +24,7 @@ LAST_TAKE = 'last_take'
 not_analyzed = {'type': 'string', 'index': 'not_analyzed'}
 
 pub_status = ['usable', 'withhold', 'canceled']
-PUB_STATUS = namedtuple('PUBSTATUS', ['USABLE', 'HOLD', 'CANCELED'])._make(pub_status)
+PUB_STATUS = namedtuple('PUBSTATUS', ['USABLE', 'HOLD', 'CANCELED'])(*pub_status)
 
 metadata_schema = {
     # Identifiers
--- a/server/apps/prepopulate/data_initialization/vocabularies.json
+++ b/server/apps/prepopulate/data_initialization/vocabularies.json
@@ -148,5 +148,22 @@
                 "usageTerms": "The direct recipient must comply with the limitations specified in the AAP Information\r\nSupply Agreement relating to the AAP content including, without limitation, not permitting\r\nredistribution and storage of the content (outside the terms of the Agreement) and not\r\npermitting deep hyperlinking to the content, framing of the content on a web site,\r\nposting the content to usenet newsgroups or facilitating such actions."
             }
         ]
+    },
+    {
+        "_id": "geographical_restrictions",
+        "items": [
+            {"is_active": true, "name": "New South Wales", "value": "New South Wales"},
+            {"is_active": true, "name": "Victoria", "value": "Victoria"},
+            {"is_active": true, "name": "Queensland", "value": "Queensland"},
+            {"is_active": true, "name": "Tasmania", "value": "Tasmania"},
+            {"is_active": true, "name": "South Australia", "value": "South Australia"},
+            {"is_active": true, "name": "New Zealand", "value": "New Zealand"}
+        ]
+    },
+    {
+        "_id": "subscriber_types",
+        "items": [
+            {"is_active": true, "name": "Internet", "value": "Internet"}
+        ]
     }
 ]
--- a/server/apps/prepopulate/data_initialization/vocabularies.json
+++ b/server/apps/prepopulate/data_initialization/vocabularies.json
@@ -157,6 +157,8 @@
             {"is_active": true, "name": "Queensland", "value": "Queensland"},
             {"is_active": true, "name": "Tasmania", "value": "Tasmania"},
             {"is_active": true, "name": "South Australia", "value": "South Australia"},
+            {"is_active": true, "name": "Western Australia", "value": "Western Australia"},
+            {"is_active": true, "name": "Northern Territory", "value": "Northern Territory"},
             {"is_active": true, "name": "New Zealand", "value": "New Zealand"}
         ]
     },
--- a/server/apps/publish/archive_publish.py
+++ b/server/apps/publish/archive_publish.py
@@ -12,8 +12,9 @@ from copy import copy
 import logging
 
 from eve.versioning import resolve_document_version
-from eve.utils import config
+from eve.utils import config, ParsedRequest
 from eve.validation import ValidationError
+from apps.content import PUB_STATUS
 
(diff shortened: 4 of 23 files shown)


server/apps/publish/archive_publish.py
```diff
@@ -282,48 +302,92 @@ def publish(self, doc, updates, target_output_channels=None):
...
+        # Step 2(a)
+        if self.published_state == 'killed':
+            queued_items = get_resource_service('publish_queue').get(
+                req=None, lookup={'item_id': updated[config.ID_FIELD]})
+
+            if queued_items.count():
+                # Step 2(a)(i)
+                subscribers = self._find_subscribers(queued_items)
+                recipients = [s['email'] for s in subscribers]
+                send_article_killed_email(doc, recipients)
```
akintolga: I think the killed email should be sent to all subscribers, not only those who received the original.

server/apps/publish/archive_publish.py
```diff
@@ -282,48 +302,92 @@ def publish(self, doc, updates, target_output_channels=None):
...
+                # Step 2(b)(i)
+                subscribers = self._find_subscribers(queued_items)
+                no_formatters, queued = self.queue_transmission(updated, subscribers)
+
+                # Step 2(b)(ii)
+                active_subscribers = list(get_resource_service('subscribers').get(req=None, lookup={'is_active': True}))
+                subscribers_yet_to_receive = []
+
+                for subscriber in active_subscribers:
+                    found = False
```
akintolga: What about one liner here? But maybe less pythonic.
```python
subscribers_yet_to_receive = [sub for sub in active_subscribers if not any(sub['_id'] == s['_id'] for s in subscribers)]
```

akintolga:
:+1:

server/apps/publish/archive_publish.py
```diff
@@ -282,48 +302,92 @@ def publish(self, doc, updates, target_output_channels=None):
...
+                    for s in subscribers:
+                        if s[config.ID_FIELD] == subscriber[config.ID_FIELD]:
+                            found = True
+                            break
+
+                    if not found:
+                        subscribers_yet_to_receive.append(subscriber)
```
plamut: How about using the `for...else` construct?

```py
for subscriber in active_subscribers:
    for s in subscribers:
        if s[config.ID_FIELD] == subscriber[config.ID_FIELD]:
            break
    else: # the for loop did not end with break (i.e. "s" was not found in "subscribers")
        subscribers_yet_to_receive.append(subscriber)
```

client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,6 +1,7 @@
 <div class="dropdown" dropdown>
 	<div class="dropdown-toggle" dropdown-toggle>{{ item[field].name || item[field]}} </div>
 	<ul class="dropdown-menu pull-right">
-		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled = "disabled">{{ o.name }}</button></li>
+		<li ng-click="select()"><button>None</button></li>
```
plamut: Does this need to be translated?

client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,6 +1,7 @@
...
+		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled="disabled">{{ o.name }}</button></li>
```
plamut: `track by`?
Also, can we use a one-time binding for `o.name`?

server/apps/publish/archive_publish.py
```diff
@@ -282,48 +302,92 @@ def publish(self, doc, updates, target_output_channels=None):
                 updates['source'] = desk['source'] if desk and desk.get('source', '') \
                     else DEFAULT_SOURCE_VALUE_FOR_MANUAL_ARTICLES
 
+            updates['pubstatus'] = 'canceled' if self.publish_type == 'killed' else 'usable'
```
plamut: Would it make sense to define the list of possible `pubstatus` states in some kind of enum-ish object and use that instead of plain strings? Would provide a better protection against typos etc.

server/apps/publish/archive_publish.py
```diff
@@ -282,48 +302,92 @@ def publish(self, doc, updates, target_output_channels=None):
...
+                    found = False
+                    for s in subscribers:
+                        if s[config.ID_FIELD] == subscriber[config.ID_FIELD]:
+                            found = True
+                            break
+
+                    if not found:
+                        subscribers_yet_to_receive.append(subscriber)
+
+                if len(subscribers_yet_to_receive):
```
plamut: Perhaps make it more explicit by comparing the len to `> 0`...
(it would also make it consistent with a similar comparison a few lines below)

server/apps/publish/archive_publish.py
```diff
@@ -282,48 +302,92 @@ def publish(self, doc, updates, target_output_channels=None):
...
-                b. queue the formatted article
-        :return:
-            1. format list for which formatters are not found,
-            2. True if the article queued for transmissions False otherwise.
+        Method formats and then queues the article for transmission to the passed subscribers.
+        ::Important Note:: Format Type across Subscribers can repeat. But we can't have formatted item generated once
+        based on the format_types configured across for all the subscribers as the formatted item must have a published
+        sequence number generated by Subscriber.
+
+        :param: target_media_type - dictate if the doc being queued is a Takes Package or an Individual Article.
```
plamut: In Sphinx-style docstrings the parameter name should be placed _inside_ the colons AFAIK (`:param [type] name:`).

server/apps/publish/archive_publish_tests.py
```diff
@@ -349,7 +353,7 @@ def test_queue_transmission(self):
             self.assertEquals(0, queue_items.count())
```
plamut: Not a part of this PR, but nevertheless - `assertEquals()` is deprecated. If you feel like it, you can replace all occurrences of it in the file with `assertEqual()`.

server/apps/publish/archive_publish_tests.py
```diff
@@ -544,6 +548,12 @@ def test_remove_published_and_killed_content_separately(self):
             published_items = published_service.get_other_published_items(original[config.ID_FIELD])
             self.assertEquals(1, published_items.count())
 
+            article_in_production = get_resource_service(ARCHIVE).find_one(req=None, _id=original[config.ID_FIELD])
+            self.assertIsNotNone(article_in_production)
+            self.assertEquals(article_in_production[config.CONTENT_STATE], 'published')
+            self.assertEquals(article_in_production[config.VERSION], original[config.VERSION] + 1)
```
plamut: `assertEquals()` has been deprecated in Python3.3, I suggest replacing it with `assertEqual()` throughout the file.

server/apps/vocabularies/vocabularies.py
```diff
@@ -50,23 +51,19 @@ def on_fetched(self, doc):
...
         """
 
-        self.__filter_inactive_vocabularies(doc)
+        self._filter_inactive_vocabularies(doc)
 
-    def __filter_inactive_vocabularies(self, item):
-        active_vocabularies = []
+    def _filter_inactive_vocabularies(self, item):
+        active_vocabularies = [{k: voc[k] for k in voc.keys() if k != 'is_active' and voc.get('is_active', True)}
+                               for voc in item['items']]
```
plamut: IMO this list should be converted to a generator, because the latter is iterated over again in the very next line, resulting in another list (rendering the initial list kind of redundant).

server/apps/publish/archive_publish.py
sivakuna-aap: Enums are introduced in Python 3.4. We test against Python 3.3 also and the build will fail with enums. Thats the reason I didn't add enums. Otherwise, I prefer not only for this there are other properties like content state, article types,... where we can use Enums

server/apps/publish/archive_publish.py
sivakuna-aap: Thanks for this Tolga. I will add it.

server/apps/publish/archive_publish.py
sivakuna-aap: Thanks for the tip. However that's the convention being followed in most of the Superdesk modules. So used the same. I will add it.

server/apps/publish/archive_publish.py
plamut: I know, that's why I said enum _-ish_. ;) `collections.namedtuple`, for instance, is a good candidate, too.

server/apps/publish/archive_publish.py
plamut: Seems like we need to revise the convention then. But yeah, this would be out of the scope of this PR, thus the consistency argument prevails here, I agree.

server/apps/prepopulate/data_initialization/vocabularies.json
```diff
@@ -125,5 +125,22 @@
                 "usageTerms": "The direct recipient must comply with the limitations specified in the AAP Information\r\nSupply Agreement relating to the AAP content including, without limitation, not permitting\r\nredistribution and storage of the content (outside the terms of the Agreement) and not\r\npermitting deep hyperlinking to the content, framing of the content on a web site,\r\nposting the content to usenet newsgroups or facilitating such actions."
             }
         ]
+    },
+    {
+        "_id": "geographical_restrictions",
+        "items": [
```
mdhaman: imo, add Western Australia and Northern Territory in vocabularies. That's other half of australia. :)

server/apps/publish/archive_publish.py
mdhaman: Same as Tolga's comment. We need to broadcast kill email.

client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
sivakuna-aap: We can use one-time binding but not `track by` as the data we receive in this directive may not have `id`

plamut: Would it be possible to uniquely identify objects by some other mean, e.g. by calculating some kind of a hash for each of them?

server/apps/content.py
```diff
@@ -22,6 +23,9 @@
 
 not_analyzed = {'type': 'string', 'index': 'not_analyzed'}
 
+pub_status = ['usable', 'withhold', 'canceled']
+PUB_STATUS = namedtuple('PUBSTATUS', ['USABLE', 'HOLD', 'CANCELED'])._make(pub_status)
```
plamut: Hm, probably better not to use methods that are meant to be private, the same can be achieved by simply instantiating the class created by `namedtuple(...)`:

```py
namedtuple('PUBSTATUS', ['USABLE', 'HOLD', 'CANCELED'])(*pub_status)
```

sivakuna-aap: Right. Thanks for catching it.

server/apps/publish/archive_publish.py
```diff
@@ -282,51 +303,94 @@ def publish(self, doc, updates, target_output_channels=None):
...
+                # Step 2(b)(i)
+                queued_items = list(queued_items)
+                subscriber_ids = {queued_item['subscriber_id'] for queued_item in queued_items}
+
+                query = {'$and': [{config.ID_FIELD: {'$in': list(subscriber_ids)}}]}
+                subscribers = list(get_resource_service('subscribers').get(req=None, lookup=query))
+                no_formatters, queued = self.queue_transmission(updated, subscribers)
+
+                # Step 2(b)(ii)
+                active_subscribers = list(get_resource_service('subscribers').get(req=None, lookup={'is_active': True}))
```
plamut: Probably redundant to convert to a list here? It's enough if `active_subscribers` can be iterated over (as it happens on the next line).

plamut:
I spotted one or two extra things left, but in general this has been polished well IMO.  :+1:

client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
sivakuna-aap: Since it's a reusable component am bit hesitant to add track by as each object can have it's own structure.

client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,6 +1,7 @@
 <div class="dropdown" dropdown>
 	<div class="dropdown-toggle" dropdown-toggle>{{ item[field].name || item[field]}} </div>
 	<ul class="dropdown-menu pull-right">
-		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled = "disabled">{{ o.name }}</button></li>
+		<li ng-click="select()" translate><button>None</button></li>
```
plamut: Just to make sure - how does the `translate` directive work? Does it work with nested elements, too, or does it have to be directly applied to an element that contains the text node we want to translate?
(in other words - does it need to be moved onto the `<button>` element?)

amagdas:
:+1:

PR #792, opened 2015-07-02
Image publishing

amagdas:


sivakuna-aap:
:+1:

server/apps/publish/formatters/ninjs_formatter.py
```diff
@@ -35,8 +39,14 @@ def format(self, article, subscriber):
             for copy_property in self.direct_copy_properties:
                 if copy_property in article:
                     ninjs[copy_property] = article[copy_property]
+
+            if 'body_html' in article:
+                ninjs['body_xhtml'] = article['body_html']
```
plamut: `body_html` is a perfectly fine field name, it's allowed by the NINJS standard. The "xhtml" suffix feels kind of archaic. :)

amagdas: I agree, forgot to fix it.

plamut:
Sorry for commenting directly on [Adrian's commit](https://github.com/amagdas/superdesk/commit/aa579f8f03a0a4a540abf3cf15d84d4d6decc224#diff-28155d4a88abd17ccba1a6ef4ef4ac93R54). :/
I opened it in new tab and then started adding the comments there instead of here in the PR.

If I find some time over the weekend I'll post the same comments here for better readability.

plamut:
@amagdas There are a few more e2e tests for the API coming, I will submit a PR most likely tomorrow (Monday). If you plan to do any changes there, please wait for a bit more with that, if possible, in order to avoid duplicate work.

amagdas:
I have already a bunch of changes coming, mostly around date handling.

amagdas:
@plamut I will try to address your comments in one pass.
Here we go:
1. Eve has a default format: https://github.com/nicolaiarocci/eve/blob/develop/eve/default_settings.py#L93
DATE_FORMAT = '%a, %d %b %Y %H:%M:%S GMT'
If the format is not set and the date does not match the format, the date is saved as a string in mongo.
Long story short, PublicAPI currently saves `versioncreated` as a string which is quite bad, I think from there all the issues with the need to do a +1 on `start_date`.

2. We have a bunch of tests around of error codes but without proper e2e tests we missed some stuff:
- error codes are not properly serialized from PublicAPI, basically the client gets the stacktrace instead of the json
- the before mentioned issue around date handling

Currently I'm fixing all the tests around this issues and adding more behave tests for error and date formatting. 
I know that you prefer unit tests, but without e2e tests we don't test what the client is actually getting and using.

amagdas:
@plamut Please check the latest changes as it solved the issues mentioned before.

plamut:
Well regarding the tests, I'm only advocating that the majority of them should be unit tests because of the faster and more accurate feedback they provide, encouraging the real rapid TDD cycles (e2e tests are too slow for that). :) Despite that, e2e tests are needed as well, of course, just not as a replacement for unit tests but rather as a complement to them.

Re: dates & errors - if the dates have been indeed stored as strings, then yes, they should be converted to datetimes. Since you made quite a few changes and it's not a WIP anymore, I'll wait with my branch and rebase when this is merged. I'll probably review the updates to this PR right after the scrum.

server/superdesk/publish/publicapi.py
```diff
@@ -47,7 +50,33 @@ def _transmit(self, queue_item, subscriber):
 
     def _fix_dates(self, item):
         for field, value in item.items():
-            if isinstance(value, dict) and '$date' in value:
-                item[field] = datetime.utcfromtimestamp(value['$date'] / 1000)
+            if isinstance(value, str):
+                try:
+                    new_value = datetime.strptime(value, config.DATE_FORMAT)
+                    item[field] = new_value
+                except:
```
plamut: Should we only catch `ValueError` here? Or is it fine to silence an error if e.g. `DATE_FORMAT` does not exist in `config` for some reason?

server/publicapi/features/items.feature
```diff
@@ -14,3 +14,129 @@ Feature: Public content
...
+                            "title": "Item",
+                            "href": "items\/tag:example.com,0000:newsml_BRE9A605"
+                        }
+                    }
+                }
+           ]
+        }
+        """
+        When we get "/items"
+        Then we get list with 0 items
```
plamut: Is there a good way to test that by default only today's items are retrieved? E.g. by mocking current date for this particular case?

As it stands, the behave tests current don't really test this because of the fixed dates in given items.

server/publicapi/features/returning_errors.feature
```diff
@@ -2,16 +2,29 @@ Feature: Returning error on invalid request parameters
 
     Scenario: Sending an unknown parameter when retrieving items
         When we get "/items?parameter_x=foo"
-        Then we get error 422
+        Then we get error 10001
```
plamut: Shouldn't we be returning HTTP 422 in such cases? At least that's the API specifications says.

(the same question for the other scenarios, too)

plamut:
There are a few unit tests missing (is the coverage still 100%, for the publicapi at least?) and I made a couple of remarks, please check if it's anything significant. Otherwise it seems fine in general, and I can add the missing tests when I rebase and submit my changes.

:+1:

server/superdesk/publish/publicapi.py
amagdas: I think it's fine to silence the error.

server/publicapi/features/items.feature
amagdas: I will make a behave test with orders created for the current date.

server/publicapi/features/returning_errors.feature
amagdas: I think I prefer to be consistent with how SD API treats errors, this is something we can discuss, having standalone error codes looks better to me.

amagdas:
Merging, will add more behave tests in the next PR.

server/publicapi/features/returning_errors.feature
plamut: I think we decided for regular HTTP codes because it provides better semantics, and because the content API is treated as an independent service that does not share much with the "internal" API that serves the Superdesk client. These two have different goals and do not have to share the error codes and other features, @hlmnrmr can elaborate on that.

Anyway, thanks for merging, I'll rebase my e2e API tests branch now.

PR #794, opened 2015-07-02
SD-2463 As a user I want to see the item history - backend operations

ioanpocol:
Issues:

    ITEM_RESTORE - the client uses PATCH instead of REPLACE
    did not implement ITEM_LOCK, ITEM_UNLOCK and ITEM_ADD_TO_PACKAGE because these operations do not create a new document version

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -900,11 +900,12 @@
 
                 $scope.beforeSend = function() {
                     $scope.sending = true;
-                    return $scope.save()
-                    .then(function() {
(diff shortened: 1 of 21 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -900,11 +900,7 @@
 
                 $scope.beforeSend = function() {
                     $scope.sending = true;
-                    return $scope.save()
```
amagdas: What was the issue here?

amagdas:
:+1: needs rebase

mugurrus:
Opened another PR for rebase and fixes

PR #795, opened 2015-07-02
feat(authoring): Implemented solo mode for monitoring widget.

eozan:
SD-2503


eozan:
Because of some styling issues (setting height based on available space, without breaking normal mode), scrolling is a bit tricky. We should probably fine tune this behaviour, but functionality is complete.

Suggestions on how to address this are welcome.

plamut:
Not sure about the styling, perhaps open a ticket for the UI team, but otherwise it looks good (perhaps only add some docstrings as a cherry on top).

:+1:

amagdas:
:+1:

PR #796, opened 2015-07-03
fix(closing items): Closing item that was opened from left search panel should return to initial state

pavlovicnemanja92:
SD-2022 AND SD-2087

Open for suggestions!

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -894,7 +894,12 @@
                 $scope.close = function() {
                     _closing = true;
                     authoring.close($scope.item, $scope.origItem, $scope.dirty).then(function() {
-                        $location.url($scope.referrerUrl);
+                        if (sessionStorage.getItem('previewUrl')) {
+                            $location.url(sessionStorage.getItem('previewUrl'));
+                            sessionStorage.removeItem('previewUrl');
+                        } else {
+                            $location.url($scope.referrerUrl);
+                        }
                     });
                 };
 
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -894,7 +894,12 @@
                 $scope.close = function() {
                     _closing = true;
                     authoring.close($scope.item, $scope.origItem, $scope.dirty).then(function() {
-                        $location.url($scope.referrerUrl);
+                        if (sessionStorage.getItem('previewUrl')) {
+                            $location.url(sessionStorage.getItem('previewUrl'));
+                            sessionStorage.removeItem('previewUrl');
```
plamut: Just checking - does this get executed when the site URL changes?

plamut:
Tests are missing. :wink:

client/app/scripts/superdesk-authoring/authoring.js
pavlovicnemanja92: No, this is executed when you press close button. And it triggers URL change.

plamut: I was referring to the `sessionStorage.removeItem('previewUrl')` line - does it get executed when `$location.url(...)` on the previous line already does a redirect?

pavlovicnemanja92: Yeah, it removes value from sessionStorage... Maybe is good for transparency that I move it before $location change. I will make this change when I finish writting e2e/unt tests.

plamut: OK, it's fine then. I also like the idea of moving it before the URL change line to avoid any doubt.

amagdas:
:+1:

PR #797, opened 2015-07-06
fix(publish_filter) - Publish filter dependency checks

akintolga:


server/apps/publish/publish_filters/filter_condition.py
```diff
@@ -80,6 +80,18 @@ def on_update(self, updates, original):
         self._check_equals([doc])
         self._check_parameters([doc])
 
+    def delete(self, lookup):
+        referenced_filters = self._get_referenced_filter_conditions(lookup.get('_id'))
+        if referenced_filters.count() > 0:
+            references = ''.join([pf['name'] for pf in referenced_filters])
```
sivakuna-aap: It's better to join the publish filter names with a separator when we are expecting to have more than one in Mongo.

server/apps/publish/publish_filters/publish_filter.py
```diff
@@ -70,8 +69,46 @@ def create(self, docs, **kwargs):
...
+    def update(self, id, updates, original):
+        publish_filter = dict(original)
+        publish_filter.update(updates)
+        self._validate_no_circular_reference(publish_filter, publish_filter['_id'])
+        super().update(id, updates, original)
+
+    def delete(self, lookup):
+        referenced_filters = self._get_referenced_publish_filters(lookup.get('_id'))
+        if referenced_filters.count() > 0:
+            references = ''.join([pf['name'] for pf in referenced_filters])
```
sivakuna-aap: Same as above

server/apps/publish/publish_filters/publish_filter.py
```diff
@@ -70,8 +69,46 @@ def create(self, docs, **kwargs):
...
+            raise SuperdeskApiError.badRequestError('Publish filter has been referenced in {}'.format(references))
+        return super().delete(lookup)
+
+    def _validate_publish_filters(self, docs):
+        for doc in docs:
+            if not doc.get('name', None):
+                raise SuperdeskApiError.badRequestError('Name cannot be empty')
```
sivakuna-aap: Cerberus can handle this. Use `'required': True and 'empty': False`. `'empty': False` will prevent user from sending empty strings.

akintolga: True, since I use the post for testing against an article, I removed it from schema definition. I may need to create a new end-point for further tests (against production stories) then I'll put the name back to the schema.

server/features/filter_condition.feature
```diff
@@ -102,4 +102,22 @@ Feature: Filter Condition
...
+    """
+
+    Then we get latest
+    Given empty "publish_filters"
+    When we post to "/publish_filters" with success
+    """
+    [{"publish_filter": [{"expression": {"fc": ["#filter_conditions._id#"]}}], "name": "soccer"}]
+    """
+    When we delete "/filter_conditions/#filter_conditions._id#"
+    Then we get error 400
```
amagdas: We should check for the error here to be sure it's not some general error.

amagdas:
:+1: except the minor comment

server/features/filter_condition.feature
```diff
@@ -102,4 +102,25 @@ Feature: Filter Condition
...
+    """
+    {"_status": "ERR", "_message": "Filter condition has been referenced in pf:soccer"}
```
akintolga: Thanks @amagdas, this should clarify the error message.

marwoodandrew:
:+1:

server/apps/publish/publish_filters/filter_condition.py
```diff
@@ -80,6 +80,18 @@ def on_update(self, updates, original):
         self._check_equals([doc])
         self._check_parameters([doc])
 
+    def delete(self, lookup):
+        referenced_filters = self._get_referenced_filter_conditions(lookup.get('_id'))
+        if referenced_filters.count() > 0:
+            references = ','.join([pf['name'] for pf in referenced_filters])
```
plamut: **A tip:** in cases like this when an iterable is expected as an argument, there is no need to construct and pass in a (temporary) list, a generator expression achieves the same with less overhead.

(I know I'm late to the party with this PR already merged, but it might be useful in the future)

PR #800, opened 2015-07-06
(fix) Fix issue with GPSInfo exif metadata extraction

thnkloud9:
Fixes the error "documents must have only string keys" encountered when ingesting images with GPSInfo exif data.

Added additional unit test cases for image processing.


server/superdesk/media/image_tests.py
```diff
@@ -24,7 +24,49 @@ def test_extract_meta_json_serialization(self):
...
+    fixtures = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'fixtures')
+    # this image has UserComment exif data value as 'test'
+    img = os.path.join(fixtures, 'iphone_gpsinfo_exif.JPG')
+
+    def test_extract_meta_json_serialization(self):
+
+        with open(self.img, mode='rb') as f:
+            meta = get_meta(f)
+
+        self.assertIsNone(meta.get('UserComment', None))
```
plamut: If the second argument to `get()` is omitted, it already defaults to `None` - but it's true that the form above is more explicit, which is often not a bad thing.

Nevertheless, if you don't care about the case when `meta['UserComment'] == None` and are actually checking if `UserComment` exists at all, an even more Pythonic way to do it is:
```py
self.assertNotIn('UserComment', meta)
```
(I think that's the case here, since you are popping this key from the dictionary in `media/image.py`.

plamut:
Looks good. :+1:

amagdas:
:+1:

PR #804, opened 2015-07-07
fix(bamboo): use environment for db config

petrjasek:
on bamboo there is no db on localhost, it's configured via environment.


petrjasek:
@mdhaman it's reverting change in d8f37ce, not sure what was the problem

sivakuna-aap:
@petrjasek Before this change, when we are executing Nosetests/Behave tests in local, it was overwriting the superdesk database. This change was put in to prevent that.

PR #806, opened 2015-07-07
fix(bamboo): use environment for db config

petrjasek:
on bamboo there is no db on localhost, it's configured via environment.

refactored #804 
@sivakuna-aap

### DIFF
--- a/server/superdesk/tests.py
+++ b/server/superdesk/tests.py
@@ -42,16 +42,24 @@ test_user = {
 }
 
 
+def get_mongo_uri(key, dbname):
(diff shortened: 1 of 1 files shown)


sivakuna-aap:
Looks good. We are doing the same thing in `settings.py` also. So, can we move `get_mongo_uri()` to settings.py and reuse the function in both `settings.py` and `tests.py`?

amagdas:
:+1:

mdhaman:
:+1:

petrjasek:
@sivakuna-aap we don't change the name there we just read it from env

PR #807, opened 2015-07-07
test(bamboo): show mongo config in case test fails to connect

petrjasek:
it's only for test setup so no info should be visible in production.

previous commit didn't fix the issue..

### DIFF
--- a/server/superdesk/tests.py
+++ b/server/superdesk/tests.py
@@ -10,8 +10,10 @@
 
 
 import os
+import pymongo
 import unittest
 import elasticsearch
+
 from app import get_app
 from base64 import b64encode
 from flask import json
@@ -38,7 +40,6 @@ test_user = {
             'category': 'notifications',
             'enabled': True}
     }
-
 }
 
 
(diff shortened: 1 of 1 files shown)


server/superdesk/tests.py
```diff
@@ -87,6 +87,8 @@ def drop_mongo(app):
             app.data.mongo.pymongo(prefix='LEGAL_ARCHIVE').cx.drop_database(app.config['LEGAL_ARCHIVE_DBNAME'])
         except AttributeError:
             pass
+        except pymongo.errors.ConnectionFailure:
+            raise ValueError('Invalid mongo config: uri=%s db=%s' % (app.config['MONGO_URI'], app.config['MONGO_DBNAME']))
```
plamut: Could this error message be somewhat misleading, if for instance `ConnectionFailure` occurs because of the e.g. Mongo not running for some reason?

amagdas:
:+1:

ioanpocol:
:+1:

PR #808, opened 2015-07-07
(feat) [SD-2573] Restricting content delivery based on subscriber metadata

sivakuna-aap:
Need to update vocabularies after this ticket it merged.

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
@@ -88,10 +88,20 @@
 			<label translate>Genre</label>
 			<div sd-meta-terms data-item="item" data-field="genre" data-unique="value" data-list="metadata.genre" ng-disabled="!_editable" data-change="autosave(item)" data-postprocessing="processGenre()"></div>
 		</li>
-		<li>
-			<label translate>Restrictions</label>
-			<div class="data visible" ng-if="metadata.geographical_restrictions">
-				<div sd-meta-dropdown data-list="metadata.geographical_restrictions" data-item="item" data-field="restrictions" data-change="autosave(item)" ng-disabled="!_editable"></div>
+		<li class="terms-box">
+			<label translate>Targeted For</label>
+			<div ng-if="metadata.targeted_for">
+				<span sd-check ng-model="item.negation" tooltip="Adds Negation" tooltip-placement="right" ng-disabled="!_editable"></span>
+				<select id="targeted_for_list" ng-model="item.targeted_for_value" ng-disabled="!_editable">
+					<option value=""></option>
+					<option value="{{g.value}}" ng-repeat="g in metadata.targeted_for track by g.name">{{:: g.name}}</option>
+				</select>
+				<button class="btn btn-small btn-primary" type="button" ng-click="addTargeted()" ng-disabled="!_editable" translate>Add</button>
+				<div class="terms" ng-if="item['targeted_for'].length">
+					<ul>
+						<li ng-repeat="t in item['targeted_for']" ng-click="removeTargeted(t)">{{ :: t['allow'] == false ? t.name : 'Not ' + t.name}}<i class="icon-close-small"></i></li>
+					</ul>
+				</div>
 			</div>
 		</li>
 		<li>
@@ -144,7 +154,6 @@
 			<div class="data">{{item.creditline}}</div>
 		</li>
 	</ul>
-
 
 	<div sd-toggle-box data-title="Task details" data-open="true" ng-if="item.task">
 		<ul class="basic-list task-info">
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
@@ -94,12 +94,12 @@
 				<span sd-check ng-model="item.negation" tooltip="Adds Negation" tooltip-placement="right" ng-disabled="!_editable"></span>
 				<select id="targeted_for_list" ng-model="item.targeted_for_value" ng-disabled="!_editable">
 					<option value=""></option>
-					<option value="{{g.value}}" ng-repeat="g in metadata.targeted_for track by g.name">{{:: g.name}}</option>
+					<option value="{{:: g.value}}" ng-repeat="g in metadata.targeted_for track by g.name">{{:: g.name}}</option>
 				</select>
 				<button class="btn btn-small btn-primary" type="button" ng-click="addTargeted()" ng-disabled="!_editable" translate>Add</button>
 				<div class="terms" ng-if="item['targeted_for'].length">
 					<ul>
-						<li ng-repeat="t in item['targeted_for']" ng-click="removeTargeted(t)">{{ :: t['allow'] == false ? 'Not ' + t.name : t.name}}<i class="icon-close-small"></i></li>
+						<li ng-repeat="t in item['targeted_for']" ng-click="removeTargeted(t)">{{ :: t['allow'] === false ? 'Not ' + t.name : t.name}}<i class="icon-close-small"></i></li>
 					</ul>
 				</div>
 			</div>
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
@@ -99,7 +99,7 @@
 				<button class="btn btn-small btn-primary" type="button" ng-click="addTargeted()" ng-disabled="!_editable" translate>Add</button>
 				<div class="terms" ng-if="item['targeted_for'].length">
 					<ul>
-						<li ng-repeat="t in item['targeted_for']" ng-click="removeTargeted(t)">{{ :: t['allow'] == false ? t.name : 'Not ' + t.name}}<i class="icon-close-small"></i></li>
+						<li ng-repeat="t in item['targeted_for']" ng-click="removeTargeted(t)">{{ :: t['allow'] == false ? 'Not ' + t.name : t.name}}<i class="icon-close-small"></i></li>
 					</ul>
 				</div>
 			</div>
--- a/client/app/scripts/superdesk/filters.js
+++ b/client/app/scripts/superdesk/filters.js
@@ -106,5 +106,15 @@ define([
                     return 'danger';
                 }
             };
-        }]);
+        }])
+        .filter('mergeTargets', function() {
+            return function(array) {
+                var merged = [];
+                _.forEach(array, function(item) {
+                    merged.push(item.allow === false ? item.name : 'Not ' + item.name);
+                });
+
+                return merged.join(', ');
+            };
+        });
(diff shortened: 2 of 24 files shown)


server/apps/publish/archive_publish.py
```diff
@@ -342,12 +348,24 @@ def publish(self, doc, updates, target_media_type=None):
                                               if not any(a[config.ID_FIELD] == s[config.ID_FIELD] for s in subscribers)]
 
                 if len(subscribers_yet_to_receive) > 0:
+                    # Step 2(b)(iii)
+                    if updated.get('targeted_for'):
+                        subscribers_yet_to_receive = [s for s in subscribers_yet_to_receive
+                                                      if s.get('subscriber_type', '') != SUBSCRIBER_TYPES.DIGITAL]
```
amagdas: Should we handle `BROADCAST` subscribers as well?

server/superdesk/publish/odbc_test.py
```diff
@@ -15,7 +17,7 @@
 
 
 class ODBCTests(TestCase):
-    subscribers = [{"_id": "1", "name": "Test", "can_send_takes_packages": False, "media_type": "media",
+    subscribers = [{"_id": "1", "name": "Test", "subscriber_type": SUBSCRIBER_TYPES.WIRE, "media_type": "media",
```
amagdas: Is `media_type` field still needed?

server/apps/publish/archive_publish.py
sivakuna-aap: As per the ticket, `Any story restricted by state or region will automatically be restricted to internet clients.` That's the reason we can't restrict `Broadcast` subscribers here.

server/superdesk/publish/odbc_test.py
sivakuna-aap: Not sure. Need to check with Andrew Lasky and Vince. But at the moment it's not playing any role in Superdesk.

amagdas: Most probably it's replaced by `subscriber_type`.

amagdas:
:+1:

server/apps/publish/subscribers.py
```diff
@@ -18,6 +20,9 @@
 
 logger = logging.getLogger(__name__)
 
+subscriber_types = ['broadcast', 'digital', 'wire']
+SUBSCRIBER_TYPES = namedtuple('SUBSCRIBER_TYPES', ['BROADCAST', 'DIGITAL', 'WIRE'])(*subscriber_types)
```
amagdas: Just to be sure that I understand this
WIRE - publish to wire clients
DIGITAL - publish to digital clients, used mostly for takes/master files
BROADCAST - same as digital but in the case of takes will receive the take stories beside the take package, the only broadcast subscriber we know of currently, is PublicAPI
Are these assumptions correct?

sivakuna-aap: Wire is correct and for Digital Subscribers, the Takes Package is transmitted not the individual takes. Need to check with business if broadcast subscribers can receive the takes package. Example given by business for a broadcast subscriber is a TV Station or Radio Station.

server/superdesk/publish/odbc_test.py
sivakuna-aap: @akintolga I checked with Andrew on this. He is not sure if Media Type becomes redundant after introducing Subscriber Type. He wants you to confirm on this part. If you say Media Type is redundant then I will remove the property from both Frondend and Backend.

mdhaman: We need something to determine media client like Fairfax, Guardian and
non-media like Commbank, Myer for embargo stories or something similar. This
was few weeks ago.

On Wednesday, 8 July 2015, Siva Kuna <notifications@github.com> wrote:

> In server/superdesk/publish/odbc_test.py
> <https://github.com/superdesk/superdesk/pull/808#discussion_r34113360>:
>
> > @@ -15,7 +17,7 @@
> >
> >
> >  class ODBCTests(TestCase):
> > -    subscribers = [{"_id": "1", "name": "Test", "can_send_takes_packages": False, "media_type": "media",
> > +    subscribers = [{"_id": "1", "name": "Test", "subscriber_type": SUBSCRIBER_TYPES.WIRE, "media_type": "media",
>
> @akintolga <https://github.com/akintolga> I checked with Andrew on this.
> He is not sure if Media Type becomes redundant after introducing Subscriber
> Type. He wants you to confirm on this part. If you say Media Type is
> redundant then I will remove the property from both Frondend and Backend.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/superdesk/superdesk/pull/808/files#r34113360>.
>


-- 

*Mayur Dhamanwala | *Senior Analyst Programmer


*Australian Associated Press*
Level 5, 3 Rider Boulevard, Rhodes NSW 2138, Australia
PO Box 3411, Rhodes NSW 2138, Australia
*T *+61 2 9322 8582  *F *+61 2 9322 8086
mdhamanwala@aap.com.au | www.aap.com.au

-- 
*This email may contain information that is confidential. If you receive an 
email in error please delete it immediately.*

client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
```diff
@@ -88,10 +88,20 @@
...
-			<label translate>Restrictions</label>
-			<div class="data visible" ng-if="metadata.geographical_restrictions">
-				<div sd-meta-dropdown data-list="metadata.geographical_restrictions" data-item="item" data-field="restrictions" data-change="autosave(item)" ng-disabled="!_editable"></div>
+		<li class="terms-box">
+			<label translate>Targeted For</label>
+			<div ng-if="metadata.targeted_for">
+				<span sd-check ng-model="item.negation" tooltip="Adds Negation" tooltip-placement="right" ng-disabled="!_editable"></span>
+				<select id="targeted_for_list" ng-model="item.targeted_for_value" ng-disabled="!_editable">
+					<option value=""></option>
+					<option value="{{g.value}}" ng-repeat="g in metadata.targeted_for track by g.name">{{:: g.name}}</option>
```
plamut: `g.value` is probably fixed, too, and we can use one-time binding on it?

client/app/scripts/superdesk/filters.js
```diff
@@ -106,5 +106,15 @@ define([
                     return 'danger';
                 }
             };
-        }]);
+        }])
+        .filter('mergeTargets', function() {
+            return function(array) {
+                var merged = [];
+                _.forEach(array, function(item) {
+                    merged.push(item.allow === false ? item.name : 'Not ' + item.name);
```
plamut: Just checking - in the metadata widget template there is a snippet `t['allow'] == false ? 'Not ' + t.name : t.name`, which does the opposite from this line, i.e. prepending `'Not '` if the item is _not_ allowed. Is this the way it should be?

Also, `item.allow === false` could perhaps be written simply as `!item.allow`, but nothing significant.

server/apps/publish/subscribers.py
amagdas: If that's the case then we need  a new type for which the takes package is sent as well as the takes themselves. One example of such a subscriber is PublicApi.

server/superdesk/publish/odbc_test.py
amagdas: Isn't this the same as digital/internet?

server/apps/publish/archive_publish.py
```diff
@@ -342,12 +348,24 @@ def publish(self, doc, updates, target_media_type=None):
...
                     formatters_not_found, queued_new_subscribers = \
                         self.queue_transmission(updated, subscribers_yet_to_receive, target_media_type)
                     no_formatters.extend(formatters_not_found)
         elif self.published_state == 'published':  # Step 2(c)
             # Step 2(c)(i)
             subscribers = list(get_resource_service('subscribers').get(req=None, lookup={'is_active': True}))
+
+            # Step 2(c)(ii)
+            if updated.get('targeted_for'):
+                subscribers = [s for s in subscribers if s.get('subscriber_type', '') != SUBSCRIBER_TYPES.DIGITAL]
```
plamut: This filtering of subscribers that are not `DIGITAL` occurs in two places, perhaps we can create a function for that to avoid repeating the list comprehensions with identical logic, e.g.:
```py
from functools import partial
non_digital = partial(filter, lambda s: s.get('subscriber_type', '') != SUBSCRIBER_TYPES.DIGITAL)

...
subscribers_yet_to_receive = list(non_digital(subscribers_yet_to_receive))
...
subscribers = list(non_digital(subscribers))
```

client/app/scripts/superdesk-authoring/metadata/views/metadata-widget.html
```diff
@@ -88,10 +88,20 @@
...
+				</select>
+				<button class="btn btn-small btn-primary" type="button" ng-click="addTargeted()" ng-disabled="!_editable" translate>Add</button>
+				<div class="terms" ng-if="item['targeted_for'].length">
+					<ul>
+						<li ng-repeat="t in item['targeted_for']" ng-click="removeTargeted(t)">{{ :: t['allow'] == false ? 'Not ' + t.name : t.name}}<i class="icon-close-small"></i></li>
```
plamut: Based on the rest of the code, shouldn't we be using the identity comparison (`===`) here as well?

client/app/scripts/superdesk/filters.js
sivakuna-aap: Thanks for catching it. I forgot

server/superdesk/publish/odbc_test.py
mdhaman: You mean to say that Digital/Internet means non-media client and rest are media clients.
@amagdas I think it is better to get confirmation from POs.

sivakuna-aap: I checked with Andrew Lasky and he needs confirmation from Tolga.

mdhaman: Let @akintolga decide :100: :+1:

plamut:
:+1: 
(can't say much on the `media_type` discussion, but the rest seems good to me)

akintolga:
My recollection from Digital/Internet was they were the same. So we have three types of subscribers:
- Broadcast
- Digital (internet)
- Wire
On top of that I agree with @mdhaman that media type (although we haven't discussed in the scope of this task with Siva and Lasky) will probably be needed. So I think we can keep that field and introduce relevant restriction parameter (i.e. Media) so if necessary we can target only to non-media clients like myer. 
Common sense suggests that media/non-media is only applicable to Digital type subscribers. But we can also treat media/non-media as independent property than subscriber-type for the time being. This would be a safer bet.

mdhaman:
Media/Non Media property of subscriber came up as a discussion for embargo stories (only use case mentioned so far). Cyber currently sends embargo stories to Media clients as the Media Clients respect embargo dates. And for non-media clients it sends transmission after embargo is passed.

Question for

amagdas:
:+1:

mdhaman:
:+1:

PR #811, opened 2015-07-08
feat(workspace): add multi workspace support

petrjasek:
there is new resource `/workspaces` that can handle different types of workspaces (user, desk, etc)


server/apps/auth/auth.py
```diff
@@ -103,7 +103,8 @@ def check_permissions(self, resource, method, user):
         # Step 3: Intrinsic Privileges
         intrinsic_privileges = get_intrinsic_privileges()
         if intrinsic_privileges.get(resource) and method in intrinsic_privileges[resource]:
-            authorized = get_resource_service(resource).is_authorized(user_id=request.view_args.get('_id'))
+            service = get_resource_service(resource)
+            authorized = service.is_authorized(user_id=str(user.get('_id')), _id=request.view_args.get('_id'))
```
sivakuna-aap: There were use cases where the `user_id` is being passed in `request.view_args` but unfortunately I don't remember one.

PR #812, opened 2015-07-08
(fix) Fixed Issues related to Tasks

sivakuna-aap:
Fixed:
1. Avoiding API Call to archive when there are no updates
2. When article is submitted to a desk the version is incremented but the version isn't inserted into archive_versions

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -327,10 +327,7 @@
                     return item;
                 });
             } else {
-                var d = $q.defer();
-                d.resolve(origItem);
-
-                return d.promise;
+                return $q.when(origItem);
             }
         };
 
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -319,11 +319,19 @@
 
             stripHtml(diff);
             autosave.stop(item);
-            return api.save('archive', item, diff).then(function(_item) {
-                item._autosave = null;
-                item._locked = lock.isLocked(item);
-                return item;
-            });
+
+            if (_.size(diff) > 0) {
+                return api.save('archive', item, diff).then(function(_item) {
+                    item._autosave = null;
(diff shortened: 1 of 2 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -319,11 +319,19 @@
...
+                return api.save('archive', item, diff).then(function(_item) {
+                    item._autosave = null;
+                    item._locked = lock.isLocked(item);
+                    return item;
+                });
+            } else {
+                var d = $q.defer();
+                d.resolve(origItem);
+
+                return d.promise;
```
petrjasek: you can use `return $q.when(origItem);` which returns a promise that's resolved with given value

mdhaman:
:+1: 
Do we have behave test to check if archive_version exists?

server/apps/tasks.py
```diff
@@ -94,10 +91,11 @@ def send_to(doc, update=None, desk_id=None, stage_id=None, user_id=None):
             task['status'] = destination_stage['task_status']
 
     if update:
-        update['task'] = task
+        update.setdefault('task', {})
+        update.get('task').update(task)
         update['expiry'] = get_expiry(desk_or_stage_doc=calculate_expiry_from)
     else:
-        doc['task'] = task
+        doc.get('task', {}).update(task)
```
amagdas: This won't assign the updated task on doc if the task does not exist already.

mdhaman: you are right @amagdas.

sivakuna-aap:
We don't have behave tests.

server/apps/tasks.py
sivakuna-aap: Yes, it won't assign. It will update the task in the doc. The assumption is that the doc already has task or assigned in the first line of send_to `original_task = doc.setdefault('task', {})`

amagdas: Then no need for calling `.get`. Doing so it's a bit confusing.

petrjasek:
:+1:

PR #813, opened 2015-07-09
(fix) [SD-2269] Unclear error message when killing an item

sivakuna-aap:


server/apps/publish/archive_publish.py
```diff
@@ -318,8 +318,8 @@ def publish(self, doc, updates, target_media_type=None):
 
                 # Step 2(a)(i)
                 subscribers = list(get_resource_service('subscribers').get(req=None, lookup=None))
-                recipients = [s['email'] for s in subscribers]
-                send_article_killed_email(doc, recipients, queued_items[0].get('completed_at'))
+                recipients = [s.get('email') for s in subscribers]
```
plamut: This can be a generator, since it's only used as an iterable passed to `filter()` in the subsequent line, right?

plamut:
:+1: 

(test for this probably already exists, right?)

server/apps/publish/archive_publish.py
sivakuna-aap: I used the generator like below
```python
recipients = list((s.get('email', '') for s in subscribers))
```
The above generator generates None. That's the reason I used like that.

sivakuna-aap:
Behave tests are available.

server/apps/publish/archive_publish.py
plamut: We could even avoid that by adding the filtering condition to the expression itself, making the use of `filter()` unnecessary, what do you think?

```py
recipients = list(s.get('email') for s in subscribers if s.get('email'))
```

sivakuna-aap: Right.

plamut:
OK, so we have this covered, good. We can merge this as far as I'm concerned, the other comment was just a suggestion.

petrjasek:
:+1:

PR #815, opened 2015-07-09
test(travis): use beta chrome for testing

petrjasek:
instead of hardcoded version use devel channel to stay up to date

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,10 @@ before_install:
         npm install -g grunt-cli bower ;
       fi
     - if [ "${TARGET}" != "${TARGET/e2e/}" ]; then
+        pip install requests &&
         export DISPLAY=:99.0 &&
         /sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1920x1080x24 ;
-        wget -N --no-verbose https://storage.googleapis.com/chromium-browser-continuous/Linux_x64/335875/chrome-linux.zip && unzip chrome-linux.zip &&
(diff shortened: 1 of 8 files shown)


scripts/get_chrome_url
```diff
@@ -0,0 +1,24 @@
...
+BUILDS_URL = 'https://storage.googleapis.com/chromium-browser-continuous/Linux_x64/{}/chrome-linux.zip'
+VERSIONS_URL = 'https://omahaproxy.appspot.com/all.json'
+
+
+def get_build_no(data):
+    for platform in data:
+        if platform['os'] == OS:
+            for version in platform['versions']:
+                if version['channel'] == CHANNEL:
+                    return version['branch_base_position']
```
plamut: Ugh, this nesting is kind of hard to read. :smile: Should we perhaps flatten it by converting it to a _collection pipeline_ (as per Martin Fowler in his [artcile](http://martinfowler.com/articles/refactoring-pipelines.html))?


```py
from functools import partial
from operator import itemgetter
...

# construct mapping and filtering helper functions
valid_platforms = partial(filter, lambda platform: platform['os'] == OS)
valid_versions = partial(filter, lambda ver: ver['channel'] == CHANNEL)
platf_versions = partial(map, itemgetter('versions'))
...

def get_build_no(data):
    versions = valid_versions(platf_versions(valid_platforms(data)))  # <-- the pipeline itself
    return next(versions, {}).get('branch_base_position')
```

plamut:
Would be good to add tests for the `get_build_no()` function, but generally looks good to me.

petrjasek:
found out that this api can do the filtering. so no need for that anymore

plamut:
OK, even better for us then.

:+1:

amagdas:
:+1:

PR #818, opened 2015-07-10
[SD-2626] Assign multiple categories to content items

marwoodandrew:


sivakuna-aap:
that's a massive one

server/apps/publish/formatters/anpa_formatter.py
```diff
@@ -18,94 +18,91 @@
...
+
+                anpa.append(b'\x03')  # ETX
+
+                # time and date
+                anpa.append(datetime.datetime.now().strftime('%d-%m-%y %H-%M-%S').encode('ascii'))
+
+                anpa.append(b'\x04')  # EOT
+                anpa.append(b'\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A\x0D\x0A')
+
+                docs.append((pub_seq_num, b''.join(anpa)))
```
akintolga: Do we need to increment the pub_seq_number as every one of the formatted_item will be a delivery? The worst case would be the same subscriber getting the same story with different anpa-category but same sequence.

akintolga:
:+1:

amagdas:
:+1:

client/app/scripts/superdesk-archive/views/metadata-view.html
```diff
@@ -78,7 +78,7 @@
     <!-- category details -->
     <dl ng-if="item['anpa-category']">
         <dt translate>ANPA category</dt>
-        <dd>{{ item['anpa-category'].name }}</dd>
```
mdhaman: since we are changing type of the field from dict to list of dict. Can we change the field name to anpa_category instead of anpa-category?

marwoodandrew: Probably as good a time as any!

server/apps/content.py
```diff
@@ -97,7 +97,7 @@
 
     # Category Details
     'anpa-category': {
-        'type': 'dict',
+        'type': 'list',
         'nullable': True,
```
sivakuna-aap: Can we schema definition of each dict here so cerberus validates it?

server/superdesk/io/commands/update_ingest.py
```diff
@@ -214,12 +214,13 @@ def update_provider(provider, rule_set=None, routing_scheme=None):
 def process_anpa_category(item, provider):
     try:
         anpa_categories = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='categories')
-        if anpa_categories:
-            for anpa_category in anpa_categories['items']:
-                if anpa_category['is_active'] is True \
-                        and item['anpa-category']['qcode'].lower() == anpa_category['qcode'].lower():
-                    item['anpa-category'] = {'qcode': item['anpa-category']['qcode'], 'name': anpa_category['name']}
-                    break
+        for item_category in item['anpa-category']:
```
sivakuna-aap: Better to move the `if` condition before `for`.

server/superdesk/io/commands/update_ingest.py
```diff
@@ -214,12 +214,13 @@ def update_provider(provider, rule_set=None, routing_scheme=None):
 def process_anpa_category(item, provider):
     try:
         anpa_categories = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='categories')
```
sivakuna-aap: Better to move the `if` condition before `for`.

sivakuna-aap:
:+1: apart from minor observations it looks good.

server/superdesk/io/iptc7901_tests.py
```diff
@@ -31,10 +31,10 @@ def test_open_iptc7901_file(self):
...
         self.assertEqual('062', item['ingest_provider_sequence'])
-        self.assertEqual('i', item['anpa-category']['qcode'])
+        self.assertEqual('i', item['anpa_category'][0]['qcode'])
         self.assertEqual(211, item['word_count'])
         self.assertEqual('Germany Social Democrats: Coalition talks with Merkel could fail =', item['headline'])
         self.assertRegex(item['body_html'], '^\n   Berlin')
         self.assertEqual('Germany-politics', item['slugline'])
         self.assertEquals('R', item['priority'])
-        self.assertEquals({'qcode': 'i'}, item['anpa-category'])
+        self.assertEquals([{'qcode': 'i'}], item['anpa_category'])
```
plamut: `assertEquals()` is deprecated since Python 3.3, I suggest replacing it with `assertEqual()` everywhere.

plamut:
This one is really massive, I might have easily overlooked something, but generally my scan did not yield much, apart from the deprecated `assertEquals()`.

:+1:

PR #819, opened 2015-07-10
(fix) [SD-2684] Making destinations mandatory in Subscribers. Otherwise, publish fails by throwing "Missing key - destinations" exception

sivakuna-aap:


plamut:
Nice and clean. :+1:

amagdas:
:+1:

PR #821, opened 2015-07-10
Do not show Fetch buttons in search results if current desk is 'personal'

plamut:
As @sivakuna-aap explained in the chat, fetching ingested items to the `'personal'` desk is not allowed. This PR removes the "Fetch" and "Fetch As..." action buttons from the ingested items listed on the search results pane.

Resolves [SD-2601](https://dev.sourcefabric.org/browse/SD-2601).

### DIFF
--- a/client/app/scripts/superdesk-ingest/module.js
+++ b/client/app/scripts/superdesk-ingest/module.js
@@ -1358,7 +1358,11 @@ define([
                     {action: 'list', type: 'ingest'}
                 ],
                 privileges: {fetch: 1},
-                key: 'f'
+                key: 'f',
+                additionalCondition: ['desks', function (desks) {
+                    // fetching to 'personal' desk is not allowed
+                    return (desks.getCurrentDeskId() !== 'personal');
+                }]
             })
             .activity('externalsource', {
                 label: gettext('Get from external source'),
--- a/client/.jshintrc
+++ b/client/.jshintrc
@@ -30,6 +30,7 @@
     "afterEach": false,
     "inject": false,
     "expect": false,
+    "fail": false,
     "describe": false,
     "xdescribe": false,
     "requirejs": false,
--- a/client/app/scripts/superdesk-ingest/ingest.spec.js
+++ b/client/app/scripts/superdesk-ingest/ingest.spec.js
@@ -56,4 +56,47 @@ describe('ingest', function() {
             expect(api.save.calls.count()).toBe(2);
         }));
     });
+
+    describe('registering activities in superdesk.ingest module', function () {
+
+        beforeEach(module('superdesk.ingest'));
+
+        describe('the "archive" activity', function () {
+            var activity;
+
+            beforeEach(inject(function (superdesk) {
+                activity = superdesk.activities.archive;
+                if (angular.isUndefined(activity)) {
+                    fail('Activity "archive" is not registered.');
+                }
+            }));
+
+            it('is allowed if the current desk is not "personal"', function () {
+                var extra_condition = activity.additionalCondition,
+                    fakeDesks;
+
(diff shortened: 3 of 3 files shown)


plamut:
Where should the tests be put, I didn't find any files with tests for the `activity list` directive?

sivakuna-aap:
:+1: 

I was rethinking about the functionality of "Fetch As"...I think "Fetch As" should be available as it forces user to select a desk and stage. I will confirm this with POs on Monday.

We should be able to add e2e test, checking if buttons are available or not based on the current user desk.

ioanpocol:
@plamut Those conditions should be added where you define fetch as action and not here.
There are a few places where an action with different show conditions is defined.

plamut:
@ioanpocol Is [this](https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-ingest/module.js#L1342-1344) perhaps the right place to put the logic in? I'm not familiar with this part of the system (the actions) and the architecture is not well documented, unfortunately.

Thanks!

plamut:
@superdesk/superdesk-developers Since Ioan is on well-deserved holidays, could anybody else comment on the question above? Where should we put the logic "current desk checking" logic?

Thanks.

amagdas:
I would say this is your best bet: https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-authoring/authoring.js#L421

amagdas:
Actually here: https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-ingest/module.js#L1345
Sorry for misleading...

plamut:
Oh, so it should be defined declaratively? Does anything like filtering by current desk exist already or do I need to create a new filter? Can you also point me to the logic of these filters? Much appreciated.

sivakuna-aap:
I agree with @amagdas. Each activity takes 'condition' property where we can check the current desk of user. 

For reference, you can look at https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-archive/module.js#L211

As mentioned earlier, we need Fetch As even if user is on Personal desk but it should hide Fetch icon.

amagdas:
Yup, what @sivakuna-aap said :+1:

plamut:
Ah, the `condition` option is just what I need, thanks!

On a related note - what if a user selects 'personal' desk when asked where to fetch the item to? That option will probably need to be hidden (or is this already the case?).

sivakuna-aap:
When user uses fetch as option personal desk will not be displayed in desks list

plamut:
OK, good then. Thanks to both again!

plamut:
I have now updated the commit, please re-check. I basically moved the logic to where the activity is configured in the `superdesk.ingest` module and added the corresponding tests.

client/app/scripts/superdesk-ingest/module.js
```diff
@@ -1358,7 +1358,11 @@ define([
                     {action: 'list', type: 'ingest'}
                 ],
                 privileges: {fetch: 1},
-                key: 'f'
+                key: 'f',
+                additionalCondition: ['desks', function (desks) {
+                    // fetching to 'personal' desk is not allowed
+                    return (desks.getCurrentDeskId() !== 'personal');
+                }]
```
plamut: I've seen in the linked examples that `condition` is a function that accepts the `item`parameter, while the `additionalCondition` can accept arbitrary dependencies specified by the Angular-like array syntax. I hence used the latter, is that fine?

sivakuna-aap: That should be fine.

sivakuna-aap:
:+1:

amagdas:
:+1:

PR #826, opened 2015-07-13
custom workspaces

petrjasek:
it enables desk workspaces and lets user create 1+ custom workspace
meanwhile it avoids hardcoded personal workspace, all custom workspaces are _personal_


petrjasek:
includes #811

client/app/scripts/superdesk-workspace/workspace.js
```diff
@@ -0,0 +1,238 @@
...
+        this.active = null;
+        this.edited = null;
+        this.create = create;
+        this.save = save;
+        this.cancelEdit = cancelEdit;
+        this.setActive = setActiveWorkspace;
+        this.setActiveDesk = setActiveDesk;
+        this.getActive = getActiveWorkspace;
+        this.getActiveId = getActiveWorkspaceId;
+        this.queryUserWorkspaces = queryUserWorkspaces;
```
plamut: Perhaps just use the `vm` alias here, since we already defined it?

plamut:
Just checking - would this conflict with #821 in any way? The `currentDeskId` can now be `null`, too?

server/superdesk/__init__.py
```diff
@@ -113,3 +113,20 @@ def register_default_user_preference(preference_name, preference):
 
 def register_default_session_preference(preference_name, preference):
     default_session_preferences[preference_name] = preference
+
+
+def register_resource(name, resource, service, backend=None, privilege=None):
```
amagdas: `service` can default to `None` as well, in that case use the default Service, we have plenty of resources configured with it.

amagdas:
:+1:  :100:

petrjasek:
@plamut right with this PR it can be `null` or desk id, it's never set to `personal`

client/app/scripts/superdesk-workspace/workspace.js
petrjasek: I don't like aliases so I'm using it sparingly ;)
those vars should probably go below this public api declaration so it might not be obvious what vm is.

plamut:
Re: `null` - it shouldn't be difficult to change that, right? The condition should simply check for `null` instead of `'personal'` and that's more or less it.

petrjasek:
I think so, that's how I was fixing it at few places

plamut:
OK, good, then just rebase and fix the activity configuration.

Oh, BTW, I think the PR itself s fine (though a bit lengthy and I might have missed something).

:+1:

server/superdesk/__init__.py
```diff
@@ -113,3 +113,24 @@ def register_default_user_preference(preference_name, preference):
 
 def register_default_session_preference(preference_name, preference):
     default_session_preferences[preference_name] = preference
+
+
+def register_resource(name, resource, service=None, backend=None, privilege=None):
```
sivakuna-aap: can we have `BaseService` as default value for `service` instead of `None`?

petrjasek: yep makes more sense

amagdas:
:+1:

client/app/scripts/superdesk-ingest/module.js
```diff
@@ -1361,7 +1361,7 @@ define([
                 key: 'f',
                 additionalCondition: ['desks', function (desks) {
                     // fetching to 'personal' desk is not allowed
```
plamut: This comment should've been updated, it's now obsolete...

PR #827, opened 2015-07-13
SD-2463 As a user I want to see the item history - backend operations - rebased

mugurrus:


server/apps/archive/archive.py
```diff
@@ -362,17 +367,15 @@ def _duplicate_item(self, original_doc):
...
         del new_doc['guid']
 
+        new_doc[ITEM_OPERATION] = ITEM_DUPLICATE
         item_model = get_model(ItemModel)
 
         on_duplicate_item(new_doc)
-        item_model.create([new_doc])
-        self._duplicate_versions(original_doc['guid'], new_doc)
-
+        resolve_document_version(new_doc, 'archive', 'PATCH', new_doc)
```
sivakuna-aap: Any specific reason for moving this put of if condition? 
We need not to increment the version when the article state is 'Submitted'. 
With this change it looks like for each copy and duplicate operation we increment the version which means there won't be any change in the last two versions.

server/apps/archive/common.py
```diff
@@ -41,6 +41,13 @@
 SEQUENCE = 'sequence'
 PUBLISH_STATES = ['published', 'killed', 'corrected', 'scheduled']
 CUSTOM_HATEOAS = {'self': {'title': 'Archive', 'href': '/archive/{_id}'}}
+ITEM_OPERATION = 'operation'
+ITEM_CREATE = 'create'
+ITEM_UPDATE = 'update'
+ITEM_RESTORE = 'restore'
+ITEM_DUPLICATE = 'duplicate'
+ITEM_DESCHEDULE = 'deschedule'
+item_operations = [ITEM_CREATE, ITEM_UPDATE, ITEM_RESTORE, ITEM_DUPLICATE, ITEM_DESCHEDULE]
```
sivakuna-aap: As suggested by @plamut in one of the earlier PR's it would be good if we can create a `namedtuple()` for this instead of list

server/apps/item_lock/components/item_lock.py
```diff
@@ -80,7 +80,8 @@ def unlock(self, item_filter, user_id, session_id, etag):
             self.app.on_item_unlock(item, user_id)
 
             # delete the item if nothing is saved so far
-            if item[config.VERSION] == 0 and item['state'] == 'draft':
+            # version 1 created on lock item
+            if item[config.VERSION] == 1 and item['state'] == 'draft':
```
sivakuna-aap: It should be version 0 only. For items created in personal desk it starts with version 0. 

@petrjasek correct me if am wrong.

server/apps/tasks.py
```diff
@@ -228,10 +231,14 @@ def on_update(self, updates, original):
         old_stage_id = original.get('task', {}).get('stage', '')
         new_user_id = updates.get('task', {}).get('user', '')
         if new_stage_id and new_stage_id != old_stage_id:
+            updates[ITEM_OPERATION] = ITEM_SEND
             send_to(doc=original, update=updates, desk_id=None, stage_id=new_stage_id, user_id=new_user_id)
-        update_version(updates, original)
+        resolve_document_version(updates, ARCHIVE, 'PATCH', original)
```
sivakuna-aap: Better to cross check with @petrjasek. Logic in `update_version()` is specific to superdesk. The `resolve_document_version` will fail if there is no version property in `updates` dict

server/apps/tasks.py
```diff
@@ -228,10 +231,14 @@ def on_update(self, updates, original):
...
 
     def on_updated(self, updates, original):
+        updated = copy(original)
+        updated.update(updates)
+        insert_versioning_documents(ARCHIVE, updated)
```
sivakuna-aap: If the article is assigned to a new desk then the `updated` dict is inserted twice. Please refer 
https://github.com/superdesk/superdesk/blob/master/server/apps/tasks.py#L249-L252

server/features/content_copy.feature
```diff
@@ -28,7 +28,7 @@ Feature: Copy Content in Personal Workspace
       {"state": "draft"}
       """
       When we get "/archive/#copy._id#?version=all"
-      Then we get list with 3 items
+      Then we get list with 4 items
       When we get "/archive/"
```
sivakuna-aap: As per the Scenario, there should be 3 items only in archive_versions. Because of the changes in `duplicate_item()` in `archive.py` module we see 4 items which is not desired behavior.

server/features/content_duplication.feature
```diff
@@ -58,10 +65,10 @@ Feature: Duplication of Content within Desk
       When we get "/archive/#duplicate._id#"
       Then we get existing resource
       """
-      {"state": "submitted", "_current_version": 1, "task": {"desk": "#desks._id#", "stage": "#desks.incoming_stage#", "user": "#CONTEXT_USER_ID#"}}
+      {"state": "submitted", "_current_version": 2, "task": {"desk": "#desks._id#", "stage": "#desks.incoming_stage#", "user": "#CONTEXT_USER_ID#"}}
```
sivakuna-aap: Same as my previous comment on `content_copy.feature`.

server/apps/archive/archive.py
mugurrus: There will be no inconsistency: the duplicated article will have version 4 and 4 documents in the arcticle_versions collection (_duplicate_versions adds this version). The duplicated article needs an extra version in order to record the operation.

"We need not to increment the version when the article state is 'Submitted'. "
This is done, see above.

server/apps/tasks.py
mugurrus: update_version sets the 'version' field, but _current_version is used for versioning. Will add back update_version but the resolve_document_version call is needed too.

server/apps/tasks.py
mugurrus: Will remove the indicated lines; a new version is created every time as it is needed for recording the operation.

server/features/content_copy.feature
mugurrus: Copy operation creates an extra version in the copied article in order to record the operation.

server/features/content_duplication.feature
mugurrus: Same as above

server/apps/archive/archive.py
sivakuna-aap: 1. Yes, you were right. the `_duplicate_versions()` inserts the last version.
2. Not sure if POs agree to increment the version for recording the operation.
3. I looked into `duplicate_item()` it does increment the version irrespective of state. However, as mentioned by you we need extra version for recording the operation.

server/apps/tasks.py
sivakuna-aap: Why don't we change the `update_version()` to modify `_current_version` itself? In that case, `update_version()` will does what `resolve_document_version()` does.

mugurrus: Because I don't know the side effects of modifying update_version() to set the '_current_version' field instead of the 'version' field.

plamut: The tests should warn us about any bugs that might be introduced by such change - well, in theory at least. :smile:

mugurrus: Seems the 'version' field has some functionality so I'll leave it as it is for now: https://github.com/superdesk/superdesk/blob/master/server/apps/archive/archive.py#L159

sivakuna-aap: Check with @petrjasek once. He added this functionality to fix [SD-2177] (https://dev.sourcefabric.org/browse/SD-2177)

server/apps/archive/common.py
mugurrus: Sorry, I don't see the gain in clarity in this case

server/apps/tasks.py
```diff
@@ -260,4 +266,9 @@ def on_deleted(self, doc):
     def assign_user(self, item_id, user):
         return self.patch(item_id, {'task': user})
 
+    def _stage_changed(self, updates, original):
+        new_stage_id = updates.get('task', {}).get('stage', '')
+        old_stage_id = original.get('task', {}).get('stage', '')
+        return new_stage_id and new_stage_id != old_stage_id
```
amagdas: Sorry, already commented on the commit itself.
We need to cast to string so we know the comparison is reliable, usually the fetched item from Mongo will have the ID as ObjectId while the one from the request is string so will always return false.

amagdas:
Besides the comment around stage_id comparison, it looks good :+1:

petrjasek:
:+1:

PR #830, opened 2015-07-14
Package story as a take when sending to digital subscribers

amagdas:
SD-2674

### DIFF
--- a/server/apps/publish/archive_publish.py
+++ b/server/apps/publish/archive_publish.py
@@ -98,7 +98,6 @@ class BasePublishService(BaseService):
         """
         Handles workflow of each Publish, Corrected and Killed.
         """
-
         try:
             user = get_user()
             last_updated = updates.get(config.LAST_UPDATED, utcnow())
@@ -114,19 +113,15 @@ class BasePublishService(BaseService):
                 package_id = TakesPackageService().get_take_package_id(original)
 
                 if package_id:
-                    # process the takes to form digital master file content
-                    package, package_updates = self.process_takes(updates_of_take_to_be_published=updates,
-                                                                  original_of_take_to_be_published=original,
-                                                                  package_id=package_id)
-
-                    self._set_version_last_modified_and_state(package, package_updates, last_updated)
-                    self._update_archive(package, package_updates)
-                    package.update(package_updates)
-
-                    # send it to the digital channels
-                    queued_digital = self.publish(doc=package, updates=None, target_media_type=DIGITAL)
-
-                    self.update_published_collection(published_item_id=package[config.ID_FIELD])
+                    queued_digital = self._publish_takes_package(package_id, updates, original, last_updated)
+                else:
+                    # if item is going to be sent to digital subscribers, package it as a take
+                    if self.sending_to_digital_subscribers(updates):
+                        updated = copy(original)
+                        updated.update(updates)
+                        # create a takes package
+                        package_id = TakesPackageService().package_story_as_a_take(updated, {}, None)
+                        queued_digital = self._publish_takes_package(package_id, updates, original, last_updated)
 
                 # queue only text items
                 queued_wire = \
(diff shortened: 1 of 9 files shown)


server/apps/publish/archive_publish.py
```diff
@@ -385,6 +377,80 @@ def publish(self, doc, updates, target_media_type=None):
 
         return queued
 
+    def sending_to_digital_subscribers(self, doc):
+        subscribers, subscribers_yet_to_receive = self.get_subscribers(doc)
+        filtered_subscribers = self.filter_subscribers(doc, subscribers, DIGITAL)
+        return len(filtered_subscribers) > 0
+
+    def get_subscribers(self, doc):
+        subscribers, subscribers_yet_to_receive = [], []
```
sivakuna-aap: Instead of multiple if conditions, it would be better to have this method in base class and override the same in sub-classes.

server/apps/publish/archive_publish.py
```diff
@@ -385,6 +377,80 @@ def publish(self, doc, updates, target_media_type=None):
...
+            # Step 2(c)(i)
+            subscribers = list(get_resource_service('subscribers').get(req=None, lookup={'is_active': True}))
+
+            # Step 2(c)(ii)
+            if doc.get('targeted_for'):
+                subscribers = list(self.non_digital(subscribers))
+
+        return subscribers, subscribers_yet_to_receive
+
+    def filter_subscribers(self, doc, subscribers, target_media_type):
```
sivakuna-aap: Better to filter first and then invoke `queue_transmission()`

server/apps/publish/archive_publish.py
```diff
@@ -114,19 +115,16 @@ def update(self, id, updates, original):
...
-
-                    self.update_published_collection(published_item_id=package[config.ID_FIELD])
+                    queued_digital = self._publish_takes_package(package_id, updates, original, last_updated)
+                else:
+                    # if item is going to be sent to digital subscribers, package it as a take
+                    if self.sending_to_digital_subscribers(updates):
+                        updated = copy(original)
+                        updated.update(updates)
+                        # create a takes package
+                        package_id = TakesPackageService().package_story_as_a_take(updated, {}, None)
```
sivakuna-aap: @amagdas correct me if am wrong. Looks like we create Takes Package in both publish and correct. Shouldn't it to be 1 takes package with both published and corrected stories as package items?

amagdas: In a take package only one version of a take can exist at a  time, I'm handling the correction case if the story changed and now will be sent to a new digital subscriber, subscriber not matched the first time the story was published.

server/apps/publish/archive_publish.py
amagdas: Yup, rookie mistake, will  fix.

server/apps/publish/archive_publish.py
amagdas: I didn't want to change the current publish logic to much so basically I just extracted some functions and reused them.
I will look into optimizing the publishing flow so we don't do unneeded subscriber filtering.

server/apps/publish/archive_publish.py
```diff
@@ -385,6 +377,80 @@ def publish(self, doc, updates, target_media_type=None):
...
 
+    def sending_to_digital_subscribers(self, doc):
+        subscribers, subscribers_yet_to_receive = self.get_subscribers(doc)
+        filtered_subscribers = self.filter_subscribers(doc, subscribers, DIGITAL)
+        return len(filtered_subscribers) > 0
+
+    def get_subscribers(self, doc):
+        subscribers, subscribers_yet_to_receive = [], []
+        if self.published_state == 'corrected':
+            query = {'$and': [{'item_id': doc[config.ID_FIELD]}, {'publishing_action': 'published'}]}
```
mdhaman: include `{'publishing_action': {'$in': ['published', 'correction']}}` for correction of a correction.

sivakuna-aap:
:+1:

akintolga:
:+1:

amagdas:
Merging, will address the comments in the next PR for this story.

PR #831, opened 2015-07-14
[SD-2687] - End point for searching subscribers by filter condition

akintolga:


server/apps/publish/publish_filters/filter_condition.py
```diff
@@ -106,6 +106,12 @@ def _check_parameters(self, docs):
                     .format(doc['name'], doc['operator']))
 
     def _check_equals(self, docs):
+        """
+        Checks if any of the filter conditions in the docs
+        already exists
+        :param docs: List of filter conditions to be tested
+        :return: Raises an exception if it exists
```
plamut: In Sphinx the `:raises ExcType:` directive should be used for describing exceptions potentially raised by a method.

server/apps/publish/publish_filters/filter_condition_tests.py
```diff
@@ -394,3 +414,23 @@ def test_if_fc_is_used(self):
...
+            cmd = VocabulariesPopulateCommand()
+            filename = os.path.join(os.path.abspath(
+                os.path.dirname("apps/prepopulate/data_initialization/vocabularies.json")), "vocabularies.json")
+            cmd.run(filename)
+            self.assertTrue(len(f._check_similar(filter_condition1)) == 2)
+            self.assertTrue(len(f._check_similar(filter_condition2)) == 1)
+            self.assertTrue(len(f._check_similar(filter_condition3)) == 0)
+            self.assertTrue(len(f._check_similar(filter_condition4)) == 3)
+            self.assertTrue(len(f._check_similar(filter_condition5)) == 1)
+            self.assertTrue(len(f._check_similar(filter_condition6)) == 1)
```
plamut: Perhaps it's a bit more natural to write these as `self.assertEqual(len(...), 123)`, but nothing significant really.

server/apps/publish/subscribers.py
```diff
@@ -123,6 +132,30 @@ def on_create(self, docs):
     def on_update(self, updates, original):
         self._validate_seq_num_settings(updates)
 
+    def _get_subscribers_by_filter_condition(self, filter_condition):
+        req = ParsedRequest()
+        all_subscribers = list(super().get(req=req, lookup=None))
```
plamut: Perhaps better to leave this as is since it's only used for iterating over it (unless it's not possible to iterate over whatever `get()` returns more than once).

plamut:
Looks like a well-rounded piece of work to me, with comments, tests and decent overall readability.

:+1:

sivakuna-aap:
:+1:

server/apps/publish/subscribers.py
akintolga: True Peter. I had to have turn it to a list as I need to iterate and then re-iterate in the same function.

server/apps/publish/publish_filters/filter_condition.py
akintolga: That's a good point. Thanks for this.

amagdas:
:+1:

mdhaman:
:+1:

PR #832, opened 2015-07-14
fix(spellcheck): reload dicts when editing

petrjasek:
instead of keeping it in memory it's reloaded when opening authoring.

### DIFF
--- a/client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
+++ b/client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
@@ -68,10 +68,7 @@ function removeClass(elem, className) {
 SpellcheckService.$inject = ['$q', 'api', 'dictionaries', 'editor'];
 function SpellcheckService($q, api, dictionaries, editor) {
     var lang,
-        promise,
-        userDict,
         dict = {},
-        loaded = false,
         numberOfErrors,
         COLOR = '#123456', // use some unlikely color for hilite, we will change these to class
         COLOR_RGB = rgb(COLOR);
@@ -84,10 +81,7 @@ function SpellcheckService($q, api, dictionaries, editor) {
     this.setLanguage = function(_lang) {
         if (lang !== _lang) {
             lang = _lang;
-            dict = {};
-            userDict = null;
-            loaded = false;
-            promise = null;
+            fetchDict();
         }
     };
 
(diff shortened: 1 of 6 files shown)


plamut:
The build fails because of the bug fixed in #833, should be fine once that is merged.

amagdas:
:+1:

petrjasek:
ping @superdesk/developers

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -98,36 +85,24 @@ function SpellcheckService($q, api, dictionaries, editor) {
...
+            dict = $q.reject();
+            return;
+        }
+
+        dict = dictionaries.getActive(lang).then(function(items) {
+            dict = {};
+            angular.forEach(items, addDict);
+            return dict;
+        });
+    }
```
plamut: `dict` is initialized to an empty object when the service is instantiated, but after this is called it becomes a promise. Should we perhaps initialize it to a promise object (or at least `null`) to avoid confusion about its type when glancing over the code?

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -197,31 +173,24 @@ function SpellcheckService($q, api, dictionaries, editor) {
...
+            span.classList.add(ERROR_CLASS);
+            replace.splitText(end.offset - start.offset);
+            span.textContent = replace.textContent;
+            replace.parentNode.replaceChild(span, replace);
+        } else {
+            range.setStart(start.node, start.offset);
+            range.setEnd(end.node, end.offset);
+            selection.removeAllRanges();
+            selection.addRange(range);
+            document.execCommand('hiliteColor', false, COLOR);
```
plamut: Hmmm... should this DOM changing logic really be contained in a service? We should probably refactor this (not in this PR though).

plamut:
Looks good in general. :+1:

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
petrjasek: I think it should, but not in this one.

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
petrjasek: fixed this, it's null or a promise now. thx

PR #837, opened 2015-07-15
Fix parsing 'from' field in e-mail ingest

actionless:
https://dev.sourcefabric.org/browse/VRFPXL-23

### DIFF
--- a/server/superdesk/io/rfc822.py
+++ b/server/superdesk/io/rfc822.py
@@ -34,8 +34,11 @@ class UserNotRegisteredException(Exception):
     pass
 
 
-def get_user_by_email(from_field):
-    email_address = re.compile('^.*<(.*)>$').findall(from_field)[0]
+email_regex = re.compile('^.*<(.*)>$')
+
+
+def get_user_by_email(field_from):
+    email_address = email_regex.findall(field_from)[0]
     lookup = superdesk.get_resource_service('users').find_one(
         req=ParsedRequest(), email=email_address)
     if not lookup:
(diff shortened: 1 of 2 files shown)


marwoodandrew:
:+1:  Looks good to me!

server/superdesk/io/rfc822.py
```diff
@@ -23,11 +24,28 @@
...
 
 
+class UserNotRegisteredException(Exception):
+    pass
+
+
+email_regex = re.compile('^.*<(.*)>$')
+
+
+def get_user_by_email(field_from):
```
sivakuna-aap: It would be better to move this function to `UsersService`

server/superdesk/io/rfc822.py
```diff
@@ -54,11 +72,18 @@ def parse_email(self, data, provider):
                 if isinstance(response_part, tuple):
                     msg = email.message_from_bytes(response_part[1])
                     item['headline'] = self.parse_header(msg['subject'])
-                    item['original_creator'] = self.parse_header(msg['from'])
+                    field_from = self.parse_header(msg['from'])
+                    item['original_source'] = field_from
```
sivakuna-aap: Although, this persists successfully in Mongo this will fail when the call is routed through eve. Please register this property in `content.py` module to avoid this.

server/superdesk/io/rfc822.py
actionless: nope, it have specific parsing thing what won't make sense outside of this module

server/superdesk/io/rfc822.py
actionless: which property? both `original_creator` and `original_source` are there

server/superdesk/io/rfc822.py
actionless: or you are suggesting to split it to two routines, to parse e-mail field, and two get a user by email?

server/superdesk/io/rfc822.py
sivakuna-aap: My bad, I didn't realize that they are already declared in content.py

server/superdesk/io/rfc822.py
sivakuna-aap: Yes

sivakuna-aap:
It looks good :+1:

amagdas:
:+1:

PR #838, opened 2015-07-15
Convert "versioncreated" fields in demo data to datetime

plamut:
Also make some configuration changes so that correct database is used.

Resolves [SD-2719](https://dev.sourcefabric.org/browse/SD-2719).


amagdas:
Looks good :+1:

petrjasek:
:+1:

PR #840, opened 2015-07-16
(feat) [SD-2260] Creating an Item or Package when user is on desk should go directly to desk instead of personal

sivakuna-aap:


amagdas:
:+1:

client/app/scripts/superdesk-archive/module.js
```diff
@@ -130,6 +130,18 @@ define([
         };
     }
 
+    ArchiveService.$inject = ['desks', 'storage'];
+    function ArchiveService(desks, storage) {
+        this.addTaskToArticle = function (item) {
+            if ((!item.task || !item.task.desk) && desks.activeDeskId && desks.userDesks) {
+                var user = storage.getItem('sess:user');
```
petrjasek: there is an api for reading user id via `session` service - it has  `session.identity._id` property

client/app/scripts/superdesk-archive/module.js
```diff
@@ -130,6 +130,18 @@ define([
...
+                var currentDesk = _.find(desks.userDesks._items, {_id: desks.activeDeskId});
```
petrjasek: for this you can use `desks.getCurrentDesk();`

sivakuna-aap: `desks.getCurrentDesk()` will always give some desk based on default desk rules and user will never be able to create items in Personal Workspace.

client/app/scripts/superdesk-archive/module.js
sivakuna-aap: Is `session.identity._id` equal to `user._id`?

petrjasek: yes

client/app/scripts/superdesk-archive/module.js
petrjasek: right my bad

amagdas:
:+1:

petrjasek:
:+1:

PR #841, opened 2015-07-16
[SD-2700] - Apply selector codes to the stories going to ipnews

akintolga:


server/apps/publish/formatters/aap_ipnews_formatter.py
```diff
@@ -122,3 +174,110 @@ def format(self, article, subscriber):
...
+                self._set_fdx_selector_codes(article, subscriber_name, odbc_item, category)
+
+    def _set_ndx_selector_codes(self, article, subscriber_name, odbc_item, category):
+        if 'MONITOR' in article['slugline'].upper():
+            odbc_item['selector_codes'] = self.SELECTOR_CODES['monitor'][subscriber_name]
+            return
+        if 'DIARY' in article['slugline'].upper():
+            odbc_item['selector_codes'] = self._join_selector_codes(subscriber_name, 'printsubs', 'cnewsd', 'rnewsd')
+            return
+        if 'FRONTERS' in article['slugline'].upper():
```
plamut: It kind of makes sense to extract `article['slugline'].upper()` into a variable instead of repeatedly calling `.upper()` on the same thing.

server/apps/publish/formatters/aap_ipnews_formatter.py
```diff
@@ -122,3 +174,110 @@ def format(self, article, subscriber):
...
+        codes = []
+        for arg in args:
+            codes.extend(self.SELECTOR_CODES[arg][subscriber_name.lower()].split())
+        return ' '.join(list(set(codes)))
+
+    def _is_in_subject(self, article, qcode):
+        def compare(code1, code2):
+            return code1[:len(code2)] == code2
+
+        return any([compare(s['qcode'], qcode) for s in article.get('subject', [])])
```
plamut: No real need for the (temporary) list here, a generator expression as a parameter is already enough.

server/apps/publish/formatters/aap_ipnews_formatter_test.py
```diff
@@ -127,3 +126,18 @@ def TestMultipleCategories(self):
...
+        f = AAPIpNewsFormatter()
+        self.assertTrue(f._is_in_subject(article, '150'))
+        self.assertFalse(f._is_in_subject(article, '151'))
+        self.assertTrue(f._is_in_subject(article, '04001'))
+
+    def Test_join_selector_codes(self):
+        f = AAPIpNewsFormatter()
+        result = f._join_selector_codes('ipnewS', 'newsi', 'cnewsi', 'cnewsi')
+        result_list = result.split()
+        self.assertEquals(len(result_list), 12)
```
plamut: `assertEquals()` is deprecated, prefer using `assertEqual()` instead.

server/apps/publish/formatters/aap_ipnews_formatter.py
```diff
@@ -122,3 +174,110 @@ def format(self, article, subscriber):
...
+    def _set_selector_codes(self, article, subscriber_name, odbc_item, category):
+        desk_id = article.get('task', {}).get('desk', None)
+        if desk_id:
+            desk = superdesk.get_resource_service('desks').find_one(req=None, _id=desk_id)
+            if desk['name'] == 'NDX':
+                self._set_ndx_selector_codes(article, subscriber_name, odbc_item, category)
+            elif desk['name'] == 'SDX':
+                self._set_sdx_selector_codes(article, subscriber_name, odbc_item, category)
+            elif desk['name'] == 'FDX':
+                self._set_fdx_selector_codes(article, subscriber_name, odbc_item, category)
```
plamut: This `if..elif` logic feels repetitive, it might make sense to make it more generic, for example:

```py
handlers = {
    'NDX': self._set_ndx_selector_codes,
    'SDX': self._set_sdx_selector_codes,
    'FDX': self._set_fdx_selector_codes,
}
...
handler = handlers.get(desk['name'],  lambda *args: None)
handler(article, subscriber_name, odbc_item, category)
```
Although in this particular example the gains are small (if any), since we only have three cases to check. But I would definitely consider it with 4+ cases or if it is likely that more selector codes are added.

Just a suggestion though, no need to change it as is.

plamut:
Except for a few minor remarks it looks good to me :+1:

amagdas:
Looks good :+1:

ioanpocol:
:+1:

PR #842, opened 2015-07-16
(fix) [SD-2633] [SD-2730]

sivakuna-aap:
[SD-2633] ANPA Category and Targeted For was overwritten. ANPA Category should be multiple.
[SD-2730] Problems with correcting published items that have non-empty "Targeted For" field.
production-test-view.html was referring to anpa-category which is renamed to anpa_category


petrjasek:
might be useful to translate those tooltips, rest looks good :+1:

amagdas:
:+1:

PR #843, opened 2015-07-16
WIP: client fixes needed to run as dependency 

thnkloud9:
wrap templateUrls in assetProvider function
initial commit for grunt-angular-templates - although only for generation initially

### DIFF
--- a/client/app/scripts/superdesk-desks/desks.js
+++ b/client/app/scripts/superdesk-desks/desks.js
@@ -73,10 +73,10 @@
         });
     }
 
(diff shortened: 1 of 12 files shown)


client/app/scripts/superdesk-desks/desks.js
```diff
@@ -611,13 +610,12 @@
...
-                        if (
-                            !this.userDesks || !this.userDesks._items ||
-                            this.userDesks._items.length === 0
-                        ) {
-                            return null;
+                        if (this.activeDeskId === 'personal' ||
+                            !this.userDesks ||
+                            !this.userDesks._items ||
+                            !this.userDesks._items.length) {
+                            return 'personal';
```
plamut: Seems like this PR contains some old code, the above lines, for instance, have been changed just recently. Seems like rebase is needed.

PR #845, opened 2015-07-16
chore(ci): add eslint config for client

petrjasek:
so far there are too many errors to enable it,
but at some point it could replace jshint+jscs

### DIFF
--- a/client/.eslintignore
+++ b/client/.eslintignore
@@ -1,0 +1,10 @@
+.tmp/*
+dist/*
+bower/*
+report/*
+coverage/*
+node_modules/*
+app/scripts/bower_components/*
+
+po/*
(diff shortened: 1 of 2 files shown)


plamut:
What advantages does it offer compared to the existing linting tools we use? Could you perhaps summarize it in a few bullet points? Thanks!

petrjasek:
it has es6 support and it can handle both style and linting checks which makes it faster than jshint + jscs

sivakuna-aap:
:+1:

plamut:
:+1: 

BTW, should we create a ticket for cleaning up the Javascript codebase? It doesn't make much sense for config files to lie around when nobody uses the new tool.

PR #846, opened 2015-07-16
Fix fetching packages from public API using demo data

plamut:
Resolves [SD-2720](https://dev.sourcefabric.org/browse/SD-2720).

Things changed:
* Change demo data so that composite items only contain references to associated objects and not the objects themselves.
* Fix an issue with `uri` field generation for associated objects.
* Add/update docstrings, add quite a few tests that were missing.

### DIFF
--- a/server/publicapi/items/service.py
+++ b/server/publicapi/items/service.py
@@ -101,34 +101,27 @@ class ItemsService(BaseService):
     def on_fetched_item(self, document):
         """Event handler when a single item is retrieved from database.
 
-        It sets the item's `uri` field and removes all the fields added by the
-        `Eve` framework that are not part of the NINJS standard (except for
-        the HATEOAS `_links` object).
+        It triggers the post-processing of the fetched item.
 
         :param dict document: fetched MongoDB document representing the item
         """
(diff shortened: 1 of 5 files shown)


server/publicapi/items/service.py
```diff
@@ -137,14 +130,39 @@ def on_fetched(self, result):
...
         """
+        if document.get('type') == 'composite':
+            endpoint_name = 'packages'
+        else:
+            endpoint_name = 'items'
+
         resource_url = '{api_url}/{endpoint}/'.format(
             api_url=app.config['PUBLICAPI_URL'],
-            endpoint=app.config['URLS'][self.datasource]
+            endpoint=app.config['URLS'][endpoint_name]
```
plamut: **Explanation:** `self.datasource` is not used anymore, because `PackagesService` might use this method to generate links for the _items_ it references, resulting in invalid URLs (i.e. with correct item ID, but using the `/packages` endpoint instead of  `/items`).

amagdas:
:+1:

petrjasek:
:+1:

PR #849, opened 2015-07-20
(fix) [SD-2626] 

sivakuna-aap:
1. Checking for Multiple Categories in Publish Filters. 
2. Allowing None when user deselects publish filter in Subscribers.


akintolga:
:+1:

marwoodandrew:
:+1:

PR #852, opened 2015-07-20
(fix) [SD-2573] 

sivakuna-aap:
Fails to filter Subscribers when "Geographical Restrictions" and "Target For" values are different. For ex, a article which has "Victoria" as "Targeted For" is being transmitted to Subscribers with Geographical Restrictions as "Queensland", which shouldn't be case as the article is targeted for Victoria Subscribers only.

### DIFF
--- a/server/apps/publish/archive_publish.py
+++ b/server/apps/publish/archive_publish.py
@@ -466,15 +466,18 @@ class BasePublishService(BaseService):
                     continue
 
             if doc.get('targeted_for'):
-                matching_target = [t for t in doc.get('targeted_for')
-                                   if t['name'] == subscriber.get('subscriber_type', '') or
-                                   t['name'] == subscriber.get('geo_restrictions', '')]
+                found_match = [t for t in doc['targeted_for'] if t['name'] == subscriber.get('subscriber_type', '')]
 
-                if len(matching_target) > 0 and matching_target[0]['allow'] is False:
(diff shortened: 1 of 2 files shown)


akintolga:
:+1:

ioanpocol:
:+1:

server/apps/publish/archive_publish.py
```diff
@@ -466,15 +466,18 @@ def filter_subscribers(self, doc, subscribers, target_media_type):
...
             if doc.get('targeted_for'):
-                matching_target = [t for t in doc.get('targeted_for')
-                                   if t['name'] == subscriber.get('subscriber_type', '') or
-                                   t['name'] == subscriber.get('geo_restrictions', '')]
+                found_match = [t for t in doc['targeted_for'] if t['name'] == subscriber.get('subscriber_type', '')]
 
-                if len(matching_target) > 0 and matching_target[0]['allow'] is False:
+                if len(found_match) == 0 and subscriber.get('geo_restrictions'):
+                    found_match = [t for t in doc['targeted_for'] if t['name'] == subscriber['geo_restrictions']]
+                    if len(found_match) == 0 or (len(found_match) > 0 and found_match[0]['allow'] is False):
```
plamut: The `len(found_match) > 0` check is redundant as it always evaluates to `True` if checking comes to this point.

plamut:
Apart from the minor remark it looks good to me. :+1:

PR #855, opened 2015-07-21
[sd-2688]- Filter Search

sjunaid:


client/app/scripts/superdesk-publish/filters/publish-filters.js
```diff
@@ -483,9 +490,141 @@ function ProductionTestController($scope, filters, notify, $location, $window) {
...
+
+    populateData();
+    fetchPublishFilters();
+    $scope.isListValue = function() {
+        if ($scope.filterCondition != null) {
+            return _.contains(['in', 'nin'], $scope.filterCondition.operator) && $scope.valueLookup[$scope.filterCondition.field];
+        }
+    };
+
+    $scope.hideList = function () {
```
akintolga: Is anyone using this function?

sivakuna-aap: Yes, couple of html's are using. But the function doesn't have any logic. We can do the same without a function.

akintolga:
Overall looks good. :+1:

client/app/scripts/superdesk-publish/filters/publish-filters.js
```diff
@@ -483,9 +490,141 @@ function ProductionTestController($scope, filters, notify, $location, $window) {
...
+            notify.error(gettext('single value is required'));
+            event.preventDefault();
+            event.stopPropagation();
+            return false;
+        }
+    };
+
+    $scope.resetValues = function() {
+        $scope.searchResult = null;
+        $scope.filterCondition.values.length = 0;
```
sivakuna-aap: This should be `$scope.filterCondition.values = []` instead of zeroing the length.

client/app/scripts/superdesk-publish/filters/view/filter-search.html
```diff
@@ -0,0 +1,28 @@
...
+            <div class="field">
+                <label translate>Field</label>
+                <select class="input" ng-change="resetValues()" required ng-model="filterCondition.field">
+                    <option value="{{parameter.field}}" ng-selected="parameter.field === filterCondition.field" ng-repeat="parameter in filterConditionParameters">{{:: parameter.field}}</option>
+                </select>
+            </div>
+            <div class="field">
+                <label translate>Operator</label>
+                <select class="input" ng-change="resetValues()" required ng-model="filterCondition.operator" >
+                    <option value="{{operator}}" ng-selected="operator === filterCondition.operator" ng-repeat="operator in operatorLookup[filterCondition.field]">{{operator}}</option>
```
sivakuna-aap: One time binding?

sivakuna-aap:
Apart from minor comments it looks good :+1:

client/app/scripts/superdesk-publish/filters/publish-filters.js
plamut: Just curious - why do we need a new (empty) Array here, isn't emptying the existing one enough?

(also, if there perhaps exist any other references to this array, modifying the existing Array in place is probably what we want to do IMO)

client/app/scripts/superdesk-publish/filters/view/filter-search.html
plamut: Same thought (at two places).

We should probably add `track by` and `translate`?

Also, could the same be achieved with `ng-options` perhaps?

client/app/scripts/superdesk-publish/filters/publish-filters.js
sjunaid: True, either way we can but i choose to modify the existing one, 
http://www.2ality.com/2012/12/clear-array.html

akintolga: :+1:

PR #856, opened 2015-07-21
fix(authoring): show widgets in embedded authoring

petrjasek:
but only on the right side.

+ re-enable e2e tests and fix a few issues there

### DIFF
--- a/client/spec/search_spec.js
+++ b/client/spec/search_spec.js
@@ -29,7 +29,7 @@ describe('Search', function() {
         workspace.switchToDesk('SPORTS DESK').then(content.setListView);
         expect(element.all(by.repeater('items._items')).count()).toBe(2);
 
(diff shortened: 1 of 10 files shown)


client/spec/search_spec.js
```diff
@@ -40,11 +40,11 @@ describe('Search', function() {
         expect(element.all(by.repeater('parameter in tags.selectedKeywords')).count()).toBe(1);
     });
 
-    it('can search by subject codes field', function () {
+    xit('can search by subject codes field', function () {
```
akintolga: is xit used for disabling or making them look like pending?

ioanpocol:
:+1:

client/spec/search_spec.js
petrjasek: it says pending in logs but it's actually disabled

akintolga:
:+1:

PR #857, opened 2015-07-21
fix(ws): If ws is not available when the app server si started, the c…

ioanpocol:
…onnection with ws is not established.

In this case try to establish the connection when a notification is sent
.


petrjasek:
:+1:

plamut:
:+1:

PR #861, opened 2015-07-22
fix(users): fix user list performance

petrjasek:
there was recursive watch on each user in list plus it was fetching
users multiple times at the same time.


ioanpocol:
:+1:

Aca-jov:
:+1:

PR #864, opened 2015-07-23
SD-2336 Fetch news items via Content API by querying for individual terms

mugurrus:


ioanpocol:
please rebase

server/publicapi/tests/items_service_test.py
```diff
@@ -187,7 +187,7 @@ def test_sets_fields_filter_on_request_object(self, fake_set_fields_filter):
         instance.get(fake_request, lookup)
 
         self.assertTrue(fake_set_fields_filter.called)
-        args, kwargs = fake_set_fields_filter.call_args
+        args, _kwargs = fake_set_fields_filter.call_args
```
plamut: If the intent of the renaming is to explicitly mark the `kwargs` variable as unused, it's probably better to simply use a single underscore for that (an often used convention in Python).

server/publicapi/tests/items_service_test.py
```diff
@@ -213,24 +213,24 @@ def test_provides_request_object_to_superclass_if_not_given(self):
...
 
         self.assertTrue(fake_super_get.called)
-        args, kwargs = fake_super_get.call_args
+        args, _kwargs = fake_super_get.call_args
         self.assertGreater(len(args), 0)
 
-        query_filter = json.loads(args[0].where)
-        self.assertEqual(query_filter.get('language'), 'de')
+        query_filter = args[0].args['filters'][0]
+        self.assertEqual(query_filter.get('term').get('language'), 'de')
```
plamut: Probably better to set the default value of the first `.get()` to an empty dictionary so that a missing `'term'` would result in a failed assertion instead of an error.

server/publicapi/items/service.py
```diff
@@ -73,27 +73,27 @@ def get(self, req, lookup):
...
         request_params = req.args or {}
-        query_filter = {}
+        req.args = {}
 
+        # set the search query
         if 'q' in request_params:
-            # TODO: add validation for the "q" parameter when we define its
-            # format and implement the corresponding actions
-            query_filter = json.loads(request_params['q'])
+            req.args['q'] = request_params['q']
```
plamut: Is there any validation performed on the provided value of the parameter?

server/publicapi/items/service.py
```diff
@@ -73,27 +73,27 @@ def get(self, req, lookup):
...
+            req.args['default_operator'] = 'OR'
+
+        if 'filter' in request_params:
+            req.args['filters'] = [{'term': json.loads(request_params['filter'])}]
```
plamut: Same question for the `filter` parameter....

Also, IIRC we should demand JSON from the clients - @hlmnrmr ?

server/publicapi/items/service.py
mugurrus: The query parameter is a string that can have any value

server/publicapi/items/service.py
mugurrus: But the old version of q you implemented demanded json too:
query_filter = json.loads(request_params['q']).
Did the same rules apply?

server/apps/publicapi_publish/packages.py
```diff
@@ -12,7 +12,10 @@
 
 
 class PublicPackagesResource(PackagesResource):
-    datasource = {'source': 'items'}
+    datasource = {
+        'search_backend': 'elastic',
+        'source': 'items'
+    }
     item_methods = ['DELETE', 'PATCH', 'GET', 'DELETE']
```
mdhaman: Why we have DELETE and PATCH? My impression was that public api will be readonly.
There are 2 DELETE in the item_methods.

mugurrus: This is the superdesk public API publish service, not the public API delivery service.

server/publicapi/items/service.py
plamut: So it's probably the backend storage that will throw an error in case of a malformed value? And that would most likely result in some kind of an internal server error? I think we should make sure that InvalidParameterValueError is reported back in such cases.

Can you check if this is currently the case?

server/apps/publicapi_publish/items.py
```diff
@@ -12,8 +12,11 @@
 
 
 class PublicItemsResource(ItemsResource):
-    datasource = {'source': 'items'}
-    item_methods = ['DELETE', 'PATCH', 'GET', 'DELETE']
+    datasource = {
+        'search_backend': 'elastic',
+        'source': 'items'
+    }
+    item_methods = ['DELETE', 'PATCH', 'GET']
```
mdhaman: Thanks for the clarification.

server/publicapi/tests/items_service_test.py
```diff
@@ -187,7 +187,7 @@ def test_sets_fields_filter_on_request_object(self, fake_set_fields_filter):
         instance.get(fake_request, lookup)
 
         self.assertTrue(fake_set_fields_filter.called)
-        args, kwargs = fake_set_fields_filter.call_args
+        args, _ = fake_set_fields_filter.call_args
```
ioanpocol: There are multiple places where the same change should be performed

mugurrus: Yep, but doesn't affect the code clarity

server/publicapi/items/service.py
ioanpocol: You have to document in code the format of q.
Also q should be validated, otherwise it is an open door for injection type attack.

server/publicapi/tests/items_service_test.py
ioanpocol: So why did you changed here? The idea is to be on the same way in all places.

server/publicapi/items/service.py
plamut: Yes, initially we planned for JSON, but then steered away from it.... well, actually we haven't finalized the format for the `q` parameter initially, thus we left it as a dummy parameter until we decide how it was going  to work.

server/publicapi/items/service.py
mugurrus: JSON encoder will encode this string and it should do it safely (that's it's job) so that any injection attempt would be encoded as a JSON string.

server/publicapi/tests/items_service_test.py
mugurrus: Replaced in all places

server/publicapi/items/service.py
plamut: I'm not quite sure if we should leave the formatting errors to be handled by the backend - could you check if we return 422 on invalid parameter value/format?

mugurrus: What would constitute an invalid value for a string that can contain any text/characters?

plamut: Oh, sorry, I've re-checked and you're right, we decided not to support custom logical expressions in version 1, the query is just a flat comma-separated string of search terms, and any input is valid (provided we escape it of course, which we do).

server/publicapi/items/service.py
```diff
@@ -146,7 +148,7 @@ def _process_fetched_object(self, document):
             document.pop(field_name, None)
 
         if 'renditions' in document:
-            for k, v in document['renditions'].items():
+            for _k, v in document['renditions'].items():
```
plamut: Should probably be named simply `_`, denoting an unused variable (in contrast to a variable name prefixed with an underscore, denoting a "private" variable).

plamut:
Please check just in case if the coverage has not been decreased, but otherwise it should be good.

:+1:

ioanpocol:
:+1:

mugurrus:
Coverage is not 100% but the code that I wrote is covered.

PR #866, opened 2015-07-23
Fix (dactionary copy/paste)

Aca-jov:


ioanpocol:
:+1:

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
```diff
@@ -118,21 +119,27 @@ function SpellcheckService($q, api, dictionaries, editor) {
     /**
      * Find errors in given text
      *
-     * @param {string} text
```
petrjasek: i think you can keep the param just change it to `{Node} node`

petrjasek:
:+1:

client/app/scripts/superdesk/editor/spellcheck/spellcheck.js
Aca-jov: Yes, I deleted it, forgot to return.. I returned it now..

PR #867, opened 2015-07-24
[SD-2750] Takes Publishing

mdhaman:
- [x] Refactor the code for publishing so that subscribers are retrieved based publish action.
- [x] For takes:
      - If wire: the subsequent publishes should go to the same subscribers
      - If digital: the subsequent publishes should go at least to the same subscribers


mdhaman:
Added Content_Type named tuple. At a later stage, I will submit another PR to remove string references for content type for the whole project.

sivakuna-aap:
:+1:

akintolga:
:+1:

PR #870, opened 2015-07-24
[sd-2688-2]-(fix/improvement) - Filter search

sjunaid:


client/app/scripts/superdesk-publish/filters/publish-filters.js
```diff
@@ -580,21 +576,29 @@ function FilterSearchController($scope, filters, notify) {
...
+                'value': $scope.filterCondition.value
+            };
+            $scope.loading = true;
+            filters.getFilterSearchResults(inputs).then(function(result) {
+                $scope.searchResult = result;
+                if (result.length === 0) {
+                    notify.error(gettext('no results found'));
+                }
+                $scope.filterCondition.value = null;
+                $scope.loading = false;
```
akintolga: I think it is good practice to have this line "scope.loading = false" in a finally block so if will invalidate even if there's an error.

akintolga:
:+1:

marwoodandrew:
:+1:

PR #872, opened 2015-07-24
refactor(workspace): refactor list item action

petrjasek:
it is a single dropdown instead of 3 buttons + dropdown

SD-2715


client/app/scripts/superdesk-archive/module.js
```diff
@@ -275,10 +275,11 @@ define([
                                 });
                         }]
                 })
-                .activity('Re-write', {
+                .activity('rewrite', {
                     label: gettext('Re-write'),
                     icon: 'multi-star-color',
                     filters: [{action: 'list', type: 'archive'}],
+                    group: 'corrections',
```
mdhaman: Is the `group: 'corrections'` for grouping post publish action such as correction, kill, re-write and re-open (which is new take on the published item)?  If yes, then we would need `group: 'corrections'` for other actions as well.

petrjasek: right it is. will update

akintolga:
:+1:

ioanpocol:
:+1:

PR #875, opened 2015-07-27
fix(auth): use asset for reset password template

petrjasek:
seems working with asset now @vied12


petrjasek:
replaces #742

vied12:
:+1:  !

ioanpocol:
:+1:

PR #876, opened 2015-07-27
fix(menu): fix case when privileges are not in storage

petrjasek:
it stops using storage as user preferences cache altogether,
because after making a change storage prefs were not complete
and on page reload it would result in empty privileges.


plamut:
Docstrings need to be added for the modified functions, but otherwise it looks good.

:+1:

sivakuna-aap:
:+1:

akintolga:
:+1:

PR #878, opened 2015-07-28
[SD-2795] Failure to re-login in case of LDAP authentication.

mdhaman:


sivakuna-aap:
:+1:

akintolga:
:+1:

mugurrus:
:+1:

server/superdesk/tests.py
```diff
@@ -146,12 +146,16 @@ def get_prefixed_url(current_app, endpoint):
...
+    :param context: test context
+    :param dict user: user
+    """
     user = user or test_user
     with context.app.test_request_context(context.app.config['URL_PREFIX']):
         original_password = user['password']
 
-        if user.get('user_type') is None:
-            user['user_type'] = 'administrator'
+        user.setdefault('user_type', 'administrator')
```
plamut: In general I like this change, very Pythonic. However, it might not exactly equivalent to the old code - can it happen that the `user_type` key exists but is set to `None`? If yes, then the user will not be set to administrator anymore as it was in the old version.

server/superdesk/tests.py
```diff
@@ -176,10 +186,26 @@ def setup_ad_user(context, user):
     ad_user = ad_user.copy()
     ad_user['email'] = 'mock@mail.com.au'
 
-    if ad_user.get('user_type') is None:
-        ad_user['user_type'] = 'administrator'
+    ad_user.setdefault('user_type', 'administrator')
```
plamut: Same remark as above, the old and new code are not exactly the same.

plamut:
Looks good in general, but please double check that we haven't accidentally introduced a bug here.

server/superdesk/tests.py
mdhaman: `user_type` is not provided. There are some test where `user_type` is provided via the step definition. It should be fine. merging the code for now as it is blocker for our ldap configuration.

plamut: Perfectly fine with me, you probably know the possible edge cases better than me. :wink:

PR #879, opened 2015-07-28
[SD-2740],[SD-2753] Simple package publishing to the public API

marwoodandrew:
Publish a simple package of an article and a picture to the API.

There is a known issue in that the package reference is to the actual text article, it should be to the digital package.

### DIFF
--- a/server/apps/publish/archive_publish.py
+++ b/server/apps/publish/archive_publish.py
@@ -198,8 +198,7 @@ class BasePublishService(BaseService):
                 try:
                     if doc[ITEM_TYPE] == CONTENT_TYPE.COMPOSITE:
                         self._publish_package_items(doc)
-
-                    archive_publish.patch(id=doc['_id'], updates=doc)
+                    archive_publish.patch(id=doc.pop('_id'), updates=doc)
                 except KeyError:
                     raise SuperdeskApiError.badRequestError("A non-existent content id is requested to publish")
             self.publish(package, updates, target_media_type=DIGITAL)
--- a/server/apps/publish/archive_publish.py
+++ b/server/apps/publish/archive_publish.py
@@ -108,7 +108,7 @@ class BasePublishService(BaseService):
             last_updated = updates.get(config.LAST_UPDATED, utcnow())
 
             if original[ITEM_TYPE] == CONTENT_TYPE.COMPOSITE:
-                self._publish_package_items(original, last_updated)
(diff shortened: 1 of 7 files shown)


server/apps/publish/archive_publish.py
```diff
@@ -124,19 +124,22 @@ def update(self, id, updates, original):
...
-                            updated = copy(original)
-                            updated.update(updates)
-                            # create a takes package
-                            package_id = TakesPackageService().package_story_as_a_take(updated, {}, None)
-                            original = get_resource_service('archive').find_one(req=None, _id=original['_id'])
-                            queued_digital, takes_package = self._publish_takes_package(package_id, updates,
-                                                                                        original, last_updated)
+                        # if text or preformatted item is going to be sent to digital subscribers, package it as a take
+                        if self.sending_to_digital_subscribers(original):
+                            if (original['type'] == 'text' or original['type'] == 'preformatted'):
```
sivakuna-aap: Please use constants which @mdhaman has put for content type.

server/apps/publish/archive_publish.py
```diff
@@ -124,19 +124,23 @@ def update(self, id, updates, original):
...
-                            updated.update(updates)
-                            # create a takes package
-                            package_id = TakesPackageService().package_story_as_a_take(updated, {}, None)
-                            original = get_resource_service('archive').find_one(req=None, _id=original['_id'])
-                            queued_digital, takes_package = self._publish_takes_package(package_id, updates,
-                                                                                        original, last_updated)
+                        # if text or preformatted item is going to be sent to digital subscribers, package it as a take
+                        if self.sending_to_digital_subscribers(original):
+                            # takes packages are only created for these types
+                            if (original['type'] == CONTENT_TYPE.TEXT or original['type'] == CONTENT_TYPE.PREFORMATTED):
```
plamut: If you feel this line is a bit lengthy, you can shorten it to the following (esp. useful if there are 3 or more checks):
```py
original['type'] in (CONTENT_TYPE.TEXT, CONTENT_TYPE.PREFORMATTED)
```
No need to change it in this particular case though, just a tip.

plamut:
Looks good so far IMO, apart from a few failed tests that need to be fixed.

ioanpocol:
:+1:  (except the 2 failing behave tests)

mugurrus:
:+1: (just fix tests)

PR #881, opened 2015-07-28
chore(docker): use 'extends'

actionless:
now to use development docker-compose config do:
```sh
docker-compose -f docker/docker-compose-dev.yml up
docker-compose -f docker/docker-compose-dev.yml ps
```
and so on


sivakuna-aap:
:+1:

plamut:
:+1:

PR #882, opened 2015-07-28
Fix package preview in the sidebar.

plamut:
Resolves [SD-2581](https://dev.sourcefabric.org/browse/SD-2581).

The preview template seemed to display an incorrect object - some `item` way up the scope hierarchy, instead of `data`, the actual item contained in a package fetched from the server.

I also removed one-time bindings at a few places that were causing incorrect behavior, e.g. picture preview not shown etc.

### DIFF
--- a/client/app/scripts/superdesk-packaging/packaging.js
+++ b/client/app/scripts/superdesk-packaging/packaging.js
@@ -467,7 +467,7 @@
             link: function(scope) {
                 scope.data = null;
                 scope.error = null;
-                scope.type = scope.item.type || scope.item.itemClass.split(':')[1];
+
                 if (scope.item.location) {
                     api[scope.item.location].getById(scope.item.residRef)
                     .then(function(result) {
--- a/client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
+++ b/client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
@@ -1,28 +1,32 @@
-<div class="item {{ :: type }}" ng-class="{locked: isLocked, published: isPublished}">
+<div class="item {{ :: data.type }}" ng-class="{locked: isLocked, published: isPublished}">
 	<div class="info">
-	    <figure class="icons-holder" ng-if=":: type === 'composite' || type === 'text'">
-	    	<i class="filetype-icon-large-{{ :: type }}"></i>
+        <figure class="icons-holder" ng-if="data.type === 'composite' || data.type === 'text'">
+            <i class="filetype-icon-large-{{ :: data.type }}"></i>
 	    </figure>
 
-		<div class="holder" ng-if=":: type !== 'composite' && type !== 'text'">
+		<div class="holder" ng-if="data.type !== 'composite' && data.type !== 'text'">
 			<div class="loading" ng-show="data == null && error == null"></div>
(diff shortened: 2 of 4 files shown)


client/app/scripts/superdesk-packaging/packaging.js
```diff
@@ -467,7 +467,7 @@
             link: function(scope) {
                 scope.data = null;
                 scope.error = null;
-                scope.type = scope.item.type || scope.item.itemClass.split(':')[1];
```
plamut: This line was causing an error, `item` did not have neither `type` nor the `itemClass` property - for the packages created from RSS feeds at least. But does it behave differently for Reuters packages?

client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
```diff
@@ -1,28 +1,32 @@
...
-            <div ng-show=":: item.headline || item.slugline || item.description">{{ :: item.headline || item.slugline || item.description}}</div>
-            <div ng-hide=":: item.headline || item.slugline || item.description" translate>Blank headline received</div>
+            <div ng-show=":: data.headline || data.slugline || data.description">{{ :: data.headline || data.slugline || data.description}}</div>
+            <div ng-hide=":: data.headline || data.slugline || data.description" translate>Blank headline received</div>
         </div>
 
-        <a ng-href="#/{{ :: type !== 'composite' ? 'authoring' : 'packaging'}}/{{ :: item.residRef }}/view?_id={{ :: item.guid }}" 
-        	title="{{ :: 'Open item' | translate }}" ng-if="item.guid && !error" ng-click="$event.stopPropagation();" class="open-item">
+        <a
+            ng-href="#/{{ data.type !== 'composite' ? 'authoring' : 'packaging'}}/{{ data.residRef }}/view?_id={{ data.guid }}"
```
plamut: Is this URL correct? Data (an object fetched from the server) doesn't seem to have the `residRef` property.

client/app/scripts/superdesk-packaging/packaging.js
petrjasek: for reuters it's like `icls:picture`, `icls:text` etc

client/app/scripts/superdesk-packaging/views/sd-package-item-preview.html
petrjasek: looks ok

client/app/scripts/superdesk-packaging/packaging.js
plamut: So Reuters objects have this property? Is it better to then leave this line as is (with additional checks for undefined) and update the `type` that is used in the template when we fetch data from server?
(if `item.location` is defined)

petrjasek: they have it, but there is no need to use it because it's fetching the item anyhow.
so you can move this `scope.type = ` part after `scope.data = result;` or you can change template to use `data.type`

plamut: Yes, the latter is exactly what I have done, more clear IMO where the type comes from.

I won't do any changes then, thanks.

petrjasek:
:+1:

ioanpocol:
:+1:

PR #877, opened 2015-07-27
feat (header): Improve the header information

pavlovicnemanja92:
SD-2742

PR #884, opened 2015-07-28
fix (slider replace): Removed replace from sd slider, because it is deprecated

pavlovicnemanja92:
https://github.com/superdesk/superdesk/pull/877

According to @petrjasek comment, updated sdSlider and removed `replace`.


petrjasek:
:+1:

ioanpocol:
:+1:

PR #885, opened 2015-07-28
fix(preferences): make sure we have identity before loading

petrjasek:
it should keep promises around until user authenticates. and only
reset preferences if user was already authenticated and then lost
token.

SD-2768


ioanpocol:
:+1:

sivakuna-aap:
:+1:

PR #887, opened 2015-07-28
fix(workspace): move links from top bar to side bar

petrjasek:
there are no icons for ingest/content, it's rotated text atm.
also changed css so when there is route with sidebar it handles
main container position instead of positioning ng-view contents.

SD-2788


ioanpocol:
:+1:

plamut:
:+1:

sivakuna-aap:
:+1: Can we add tooltips to the icons in sidenav? QA has raised an issues about the same. SD-2789

petrjasek:
@sivakuna-aap will add it in a different PR, working on something related

sivakuna-aap:
ok :+1:

PR #888, opened 2015-07-29
[SD-2754] Take sequence field used for the headline

akintolga:


server/apps/packages/takes_package_service.py
```diff
@@ -140,6 +141,9 @@ def link_as_next_take(self, target, link):
             resolve_document_version(takes_package, ARCHIVE, 'PATCH', takes_package)
             archive_service.patch(takes_package_id, takes_package)
 
+        if link.get('sequence'):
```
sivakuna-aap: Better to use constant `SEQUENCE`

server/apps/packages/takes_package_service.py
```diff
@@ -140,6 +141,9 @@ def link_as_next_take(self, target, link):
...
+            archive_service.patch(link['_id'], {'sequence': link['sequence']})
```
sivakuna-aap: Same as above for `_id` we can use `config.ID_FIELD`

server/apps/archive/common.py
```diff
@@ -378,6 +378,9 @@ def item_schema(extra=None):
             'type': 'string',
             'mapping': not_analyzed,
             'nullable': True
+        },
+        SEQUENCE: {
```
mdhaman: Maybe change the field name to `take_sequence'.

akintolga: Discussed with Mayur, this will affect existing packages so we will handle this during a refactor where we need to wipe the db.

mdhaman:
:+1:

sivakuna-aap:
:+1:

PR #890, opened 2015-07-29
workspace updates

petrjasek:
- [x] make multi actions work in monitoring
- [x] add titles to sidebar icons
- [x] make opened articles visible everywhere in workspace

### DIFF
--- a/client/app/scripts/superdesk-dashboard/views/workspace-sidenav.html
+++ b/client/app/scripts/superdesk-dashboard/views/workspace-sidenav.html
@@ -1,26 +1,33 @@
 <ul>
     <li ng-class="{active: currentRoute.href == '/workspace'}">
-        <a href="#/workspace" class="big-icon-dashboard"></a>
+        <a href="#/workspace" class="big-icon-dashboard"
+            title="{{ :: 'Dashboard' | translate }}"></a>
     </li>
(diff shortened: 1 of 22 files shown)


ioanpocol:
:+1:

client/app/scripts/superdesk-dashboard/views/workspace-sidenav.html
```diff
@@ -1,26 +1,33 @@
 <ul>
     <li ng-class="{active: currentRoute.href == '/workspace'}">
-        <a href="#/workspace" class="big-icon-dashboard" translate></a>
+        <a href="#/workspace" class="big-icon-dashboard"
+            title="{{ :: 'Dashboard' | translate }}"></a>
```
pavlovicnemanja92: Maybe to add `tooltip` here as we have it on right side bar.

petrjasek: I'm not using tooltips here because need to check it's performance implications. it's doing some dom manipulation every digest cycle

pavlovicnemanja92: Yes, that is the downside of this fancy tooltips ...

pavlovicnemanja92:
:+1:

PR #891, opened 2015-07-29
feat (desk name): Added desk name in top of authoring screen

pavlovicnemanja92:
SD-2716


ioanpocol:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -721,6 +721,9 @@
                     $scope.itemActions = null;
                     if (new_value) {
                         $scope.itemActions = authoring.itemActions(new_value);
+                        desks.fetchDeskById(new_value.task.desk).then(function (desk) {
```
petrjasek: imo it would be better if it's somewhere nearby stage handling. and probably both could be in a directive

PR #902, opened 2015-08-03
[SD-2741] - Headline prefix mapper for aap ipnews formatter

akintolga:


server/apps/publish/formatters/aap_ipnews_formatter.py
```diff
@@ -175,115 +128,3 @@ def format(self, article, subscriber):
 
     def can_format(self, format_type, article):
         return format_type == 'AAP IPNEWS' and article['type'] in ['text', 'preformatted']
```
sivakuna-aap: Can we replace with constants?

server/apps/publish/formatters/field_mappers/locator_mapper.py
```diff
@@ -0,0 +1,104 @@
...
+        '15064000': 'TAE',
+        '15065000': 'TEN',
+        '15063000': 'TTEN',
+        '15067000': 'VOL',
+        '15070000': 'WGHT',
+        '15068000': 'WPOL',
+        '15072000': 'WRES'
+    }
+
+    def map(self, article, category):
```
sivakuna-aap: If this is overriding the base class method then the method definition in base and sub classes are not matching.

server/apps/publish/formatters/field_mappers/locator_mapper_tests.py
```diff
@@ -0,0 +1,51 @@
...
+             {'_id': 2, 'name': 'Sports'},
+             {'_id': 3, 'name': 'Finance'}]
+
+    def setUp(self):
+        super().setUp()
+        with self.app.app_context():
+            self.app.data.insert('desks', self.desks)
+            init_app(self.app)
+
+    def Test_map_subject_code(self):
```
sivakuna-aap: should start with small case.

server/apps/publish/formatters/field_mappers/selectorcode_mapper_tests.py
```diff
@@ -0,0 +1,55 @@
...
+    desks = [{'_id': 1, 'name': 'National'},
+             {'_id': 2, 'name': 'Sports'},
+             {'_id': 3, 'name': 'Finance'}]
+
+    def setUp(self):
+        super().setUp()
+        with self.app.app_context():
+            self.app.data.insert('desks', self.desks)
+            init_app(self.app)
+
```
sivakuna-aap: Should start with small case

server/apps/publish/formatters/aap_ipnews_formatter.py
akintolga: All formatters are based on the string literals at the moment. Out of curiosity I did a (global) search say 'preformatted' I've got 33 results so it is used in literal form pretty much every where. I suggest we discuss this approach and create a task to refactor the literals globally if we all agree.

sivakuna-aap: Agreed that there are modules which have string literals. My comment was just related to this module only. 

I also agree that we need a backlog task for replacing these string literals globally with constant.

mdhaman: @akintolga I have created this content type named tuple. I think @sivakuna-aap referring to that. I will do that refactor this sprint.

akintolga: Created the task for the AU backlog.
https://dev.sourcefabric.org/browse/SD-2856

server/apps/publish/formatters/field_mappers/selectorcode_mapper_tests.py
akintolga: Have a look at the file and let me know if you see any other issues. Your comment missed the line I guess.

server/apps/publish/formatters/aap_ipnews_formatter.py
plamut: BTW, in general it is better to use `sets` for containment testing, i.e. `if x in {'value1', 'value2'}`, because of O(1) lookup (compared to O(n) for lists/sets), but it's hardly of any significance here in this particular case.

PR #905, opened 2015-08-03
Monitoring filter

Aca-jov:
SD-2713 SD-2820 SD-2831

### DIFF
--- a/client/app/scripts/superdesk-desks/aggregate.js
+++ b/client/app/scripts/superdesk-desks/aggregate.js
@@ -28,6 +28,7 @@
         this.stageLookup = {};
         this.fileTypes = ['all', 'text', 'photo', 'composite', 'video', 'audio'];
         this.selectedFileType = 'all';
+        this.monitoringSearch = false;
 
         desks.initialize()
         .then(angular.bind(this, function() {
--- a/client/app/scripts/superdesk-desks/aggregate.js
+++ b/client/app/scripts/superdesk-desks/aggregate.js
@@ -83,7 +83,6 @@
          * param {string} fileType
          */
         this.setFileType = function(fileType) {
-            console.log('set filter: ', fileType);
             this.selectedFileType = fileType;
             var value = (fileType === 'all') ? null: JSON.stringify([fileType]);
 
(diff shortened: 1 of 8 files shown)


client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -66,6 +69,44 @@
...
+        this.hasFileType = function(fileType) {
+            return this.selectedFileType === fileType;
+        };
+
+        /**
+         * Set the current 'fileType' filter
+         * param {string} fileType
+         */
+        this.setFileType = function(fileType) {
+            console.log('set filter: ', fileType);
```
petrjasek: pls remove

client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -66,6 +69,44 @@
...
+         */
+        this.hasFileType = function(fileType) {
+            return this.selectedFileType === fileType;
+        };
+
+        /**
+         * Set the current 'fileType' filter
+         * param {string} fileType
+         */
+        this.setFileType = function(fileType) {
```
petrjasek: maybe we can put this filter into some `monitoring` service and make the groups watch that?
it would be simpler than updating each group. simpler to test (spoiler warning: there is no unittest for this method..)

client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -26,6 +26,9 @@
         this.searchLookup = {};
         this.deskLookup = {};
         this.stageLookup = {};
+        this.fileTypes = ['all', 'text', 'photo', 'composite', 'video', 'audio'];
```
sivakuna-aap: As mentioned by @amagdas we should consider preformatted also

client/app/scripts/superdesk-desks/aggregate.js
ioanpocol: I will refactor this as part of changing the settings at desk and workspace level.

client/app/scripts/superdesk-desks/aggregate.js
ioanpocol: Based on Andrew&Vince feedback this is not needed.

client/app/scripts/superdesk-desks/aggregate.js
petrjasek: :ok_hand:

pavlovicnemanja92:
:+1:

PR #907, opened 2015-08-03
[SD-2836] Takes cannot be scheduled.

mdhaman:


server/apps/archive/archive.py
```diff
@@ -196,12 +195,14 @@ def on_update(self, updates, original):
             return
 
         if updates.get('publish_schedule'):
+
             if datetime.datetime.fromtimestamp(False).date() == updates.get('publish_schedule').date():
```
plamut: While already changing this file, I would refactor this `False`to `0`, since semantically timestamps are integers.

server/apps/archive/archive_test.py
```diff
@@ -161,10 +161,8 @@ def test_validate_schedule(self):
...
-        with assert_raises(SuperdeskApiError):
-            ArchiveService().validate_schedule("2015-04-27T10:53:48+00:00")
-            ArchiveService().validate_schedule(date_without_time)
-            ArchiveService().validate_schedule(time_without_date)
-            ArchiveService().validate_schedule(date_in_past)
+        validate_schedule(date_in_future)
+        self.assertRaises(SuperdeskApiError, validate_schedule, "2015-04-27T10:53:48+00:00")
+        self.assertRaises(SuperdeskApiError, validate_schedule, date_without_time)
+        self.assertRaises(SuperdeskApiError, validate_schedule, time_without_date)
+        self.assertRaises(SuperdeskApiError, validate_schedule, date_in_past)
```
plamut: It seems that this method tries to test to cover to many test cases. Since being a unit test, I suggest splitting it into multiple test cases with each of them testing only a single edge case.

server/apps/archive/common.py
```diff
@@ -372,6 +372,29 @@ def handle_existing_data(doc, pub_status_value='usable', doc_type='archive'):
...
+    :param original: original item
+    :raises: SuperdeskApiError.badRequestError if following cases
+        - Not a valid date or Not a valid time
+        - Less than current time
+        - Takes cannot be scheduled.
+    """
+    if schedule:
+        if not isinstance(schedule, datetime):
+            raise SuperdeskApiError.badRequestError("Schedule date is not recognized")
+        if not schedule.date() or schedule.date().year <= 1970:
```
plamut: Should this be strict "less than"? After all 1970 is a valid year in epoch time.

server/apps/publish/archive_publish_tests.py
```diff
@@ -423,14 +422,17 @@ def test_queue_transmission_for_scheduled_publish(self):
...
-            self.assertEqual("2016-05-30T10:00:00+0000", queue_items[0]["publish_schedule"])
-            self.assertEqual("2016-05-30T10:00:00+0000", queue_items[1]["publish_schedule"])
-            self.assertEqual("2016-05-30T10:00:00+0000", queue_items[2]["publish_schedule"])
-            self.assertEqual("2016-05-30T10:00:00+0000", queue_items[3]["publish_schedule"])
-            self.assertEqual("2016-05-30T10:00:00+0000", queue_items[4]["publish_schedule"])
+            self.assertEqual(schedule_date, queue_items[0]["publish_schedule"])
+            self.assertEqual(schedule_date, queue_items[1]["publish_schedule"])
+            self.assertEqual(schedule_date, queue_items[2]["publish_schedule"])
+            self.assertEqual(schedule_date, queue_items[3]["publish_schedule"])
+            self.assertEqual(schedule_date, queue_items[4]["publish_schedule"])
```
plamut: IMO better to convert this checking to a loop for brevity.

plamut:
Looks good in general, just please see the comments if you think any of them is significant enough to warrant a change.

:+1:

server/apps/archive/common.py
```diff
@@ -372,6 +372,29 @@ def handle_existing_data(doc, pub_status_value='usable', doc_type='archive'):
...
+            raise SuperdeskApiError.badRequestError("Schedule date is not recognized")
+        if not schedule.time():
+            raise SuperdeskApiError.badRequestError("Schedule time is not recognized")
```
sivakuna-aap: I think `schedule.time()` check is not needed as we are already checking if `schedule` is of type `datetime` or not. Even if user doesn't specify anything for time by default python returns time `00:00:00`.

server/apps/archive/common.py
```diff
@@ -372,6 +372,29 @@ def handle_existing_data(doc, pub_status_value='usable', doc_type='archive'):
             doc['marked_for_not_publication'] = False
 
 
+def validate_schedule(schedule, package_sequence=1):
+    """
+    Validates the publish schedule.
+    :param updates: updates to the item
```
akintolga: Function comments need a rejig

sivakuna-aap:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1335,6 +1335,12 @@
                 };
 
                 scope.sendAndContinue = function () {
```
plamut: Don't forget the docstring here. :wink:

mdhaman:
@plamut done. thanks

PR #908, opened 2015-08-03
(blocker) Fix locators reading on ubuntu

amagdas:
SD-2858 sd-test server is not working

### DIFF
--- a/server/superdesk/locators/__init__.py
+++ b/server/superdesk/locators/__init__.py
@@ -19,7 +19,7 @@ def _load_json(file_path):
     :param file_path: path of the file having JSON string.
     :return: JSON Object
     """
-    with open(file_path, 'r') as f:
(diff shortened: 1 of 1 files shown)


plamut:
Test is missing for the change, but good otherwise.

amagdas:
There are a bunch of tests for this feature, this failed at runtime with celery, I don't really see the value in trying to add a test for this.

plamut:
Hm, so there is no good way to reproduce this error in a test environment that would prove the issue (and the fix for it), if I understood correctly?

amagdas:
Yup, that's the gist of it. Should I mention that I tested the fix directly on the test server? :imp:

plamut:
Meh, then we'll have to live with it I guess...

:+1:

sivakuna-aap:
:+1:

PR #910, opened 2015-08-04
fix(schema): use elastic that can handle dict mapping

petrjasek:


sivakuna-aap:
We had an issue with elastic serializer. When null/none value is passed to a referenced property the serializer serializes it to some ObjectId value and checks if a document with that ObjectId exists in the referenced collection and eventually fails as there won't be any document with that object id. Not sure if this will reproduce that issue!!! We might need a test to check if null values gets converted to ObjectId or not.

ioanpocol:
:+1:

petrjasek:
@sivakuna-aap added a test, but not sure it matches the issue, what do you think?

amagdas:
:+1:

sivakuna-aap:
@petrjasek I think the test you added should cover

PR #911, opened 2015-08-04
[sd-2744]- (& sd-2635) - Highlight lists on the new workspace

sjunaid:


client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -250,23 +250,50 @@
...
+                    scope.selected = active;
+                    highlightsService.get(desks.getCurrentDeskId()).then(function(result) {
+                        scope.highlights = result._items;
+                        scope.hasHighlights = _.size(scope.highlights) > 0;
+                    });
+                });
+
+                scope.listHighlight = function(highlight) {
+                    var data = {highlightId: highlight._id, name: highlight.name};
+                    $location.url('workspace/highlights-view?data=' + encodeURIComponent(JSON.stringify(data)));
```
petrjasek: it would be easier to use this from other places if the url is like `workspace/highlights/_id` or `workspace/highlights?highlight=<id>`, name can be fetched in activity controller.

client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -36,6 +37,18 @@
             });
     }
 
+    configureHighlightsView.$inject = ['superdeskProvider'];
```
petrjasek: this should be in `highlights.js`

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -808,6 +811,30 @@
...
+                            .then(function() {
+                                _exportHighlight(item._id);
+                            }
+                        );
+                    } else {
+                        _exportHighlight(item._id);
+                    }
+                };
+
+                function _exportHighlight(_id) {
```
plamut: Docstring is missing... we write them for private functions, too, AFAIK (but please correct me if I'm wrong).

client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -250,27 +250,67 @@
...
         return {
             templateUrl: 'scripts/superdesk-highlights/views/package_highlights_dropdown_directive.html',
             link: function(scope) {
-
-                scope.createHighlight = function(highlight) {
-                    highlightsService.createEmptyHighlight(highlight)
-                    .then(function(new_package) {
-                        superdesk.intent('author', 'package', new_package);
+                scope.$watch(function() {
+                    return desks.active;
```
plamut: Listeners on a service are not automatically removed when the directive gets `$destroy`-ed, thus this `$watch` needs to be deregistered manually to prevent a memory leak.

client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -250,27 +250,67 @@
...
+            link: function(scope, elem, attrs, ctrls) {
+                var authoring = ctrls[0];
+
+                scope.createHighlight = function(highlight) {
+                    highlightsService.get(desks.getCurrentDeskId()).then(function(result) {
+                    scope.highlights = _.find(result._items, {_id: scope.highlight_id});
+                    scope.hasHighlights = _.size(scope.highlight) > 0;
+                }).then(function() {
+                    highlightsService.createEmptyHighlight(scope.highlights)
+                    .then(function(new_package) {
```
plamut: It's a bit more readable if the promise chain is flat, rather than having nested `.then()` calls, IMO.

client/app/scripts/superdesk-highlights/views/package_highlights_dropdown_directive.html
```diff
@@ -1,8 +1,9 @@
-<div id="highlightPackage" ng-if="hasHighlights" tooltip="{{ :: 'Create highlight' | translate }}" tooltip-placement="left">
+<div id="highlightPackage" ng-if="hasHighlights" tooltip="{{ :: 'Highlights' | translate }}" tooltip-placement="right"> 
     <button class="dropdown-toggle" dropdown-toggle><i class="svg-icon-create-list"></i></button>
-    <ul class="dropdown-menu pull-right">
-		<li ng-repeat="h in highlights"> 
-		    <button option="{{ h.name | uppercase  }}" ng-click="createHighlight(h);">{{ h.name | uppercase  }}</button>
+    <ul class="dropdown-menu pull-left">
+		<li ng-repeat="h in highlights">
```
plamut: Do highlights have any unique attribute that can be used for `track by`?

client/app/scripts/superdesk-highlights/highlights.js
petrjasek: are you sure about this?
it's watching desks but listener itself is registered on scope which should get destroyed

plamut: Yes, I'm sure, although it's probably fine in this case. :)

The thing is that this directive does have its own scope, it uses its parent element's scope. In theory it's possible that the directive element gets removed (for whatever reason), while the parent element is left intact and thus its scope would still be alive, along with the directive's watcher.

But as I said, it's probably just a theoretical scenario in this particular case, right?

plamut:
Looks mostly good to me. Perhaps check the test coverage for the directives, I'm not sure if the tests cover them all, and my other comments are more or less optional, no showstoppers. :)

client/app/scripts/superdesk-highlights/highlights.js
petrjasek: I see, missed that the scope is not isolated

plamut: Now thinking... would it be better to use `scope: true` instead? The directive would still have access to parent scope's properties through the prototype chain, but would create $watches and other thing in its own scope.

ioanpocol:
:+1:

amagdas:
:+1:

PR #917, opened 2015-08-04
fix(widget): Updated the html file so now the data is showed on monit…

ioanpocol:
…oring widget

Some CSS changes are needed


client/app/scripts/superdesk-desks/aggregate-widget/aggregate-widget.html
```diff
@@ -6,23 +6,40 @@
                    ng-model-options="{debounce: 500}"
                    ng-change="agg.search(query)">
         </div>
+        
+        <div class="content-list" ng-if="!agg.state.solo">
+            <div class="desk" ng-repeat="group in agg.cards">
```
plamut: Can we add `track by` here?

mugurrus:
:+1:

amagdas:
:+1:

PR #920, opened 2015-08-05
[SD-2837] - Creating crops for picture stories

akintolga:
[SD-2838] - Validator for picture publishing
[SD-2839] - Public API: Copy all renditions to public API

### DIFF
--- a/server/apps/archive/__init__.py
+++ b/server/apps/archive/__init__.py
@@ -20,6 +20,7 @@ from .ingest import IngestResource, IngestService
 from .item_comments import ItemCommentsResource, ItemCommentsSubResource, ItemCommentsService, ItemCommentsSubService
 from .user_content import UserContentResource, UserContentService
 from .archive_lock import ArchiveLockResource, ArchiveUnlockResource, ArchiveLockService, ArchiveUnlockService
+from .archive_crop import ArchiveCropService, ArchiveCropResource
 from .archive_spike import ArchiveUnspikeResource, ArchiveSpikeService, ArchiveSpikeResource, ArchiveUnspikeService
 import superdesk
 from apps.common.components.utils import register_component
@@ -63,6 +64,10 @@ def init_app(app):
     endpoint_name = 'archive_lock'
     service = ArchiveLockService(endpoint_name, backend=superdesk.get_backend())
     ArchiveLockResource(endpoint_name, app=app, service=service)
+
+    endpoint_name = 'archive_crop'
+    service = ArchiveCropService(endpoint_name, backend=superdesk.get_backend())
+    ArchiveCropResource(endpoint_name, app=app, service=service)
 
     endpoint_name = 'archive_unlock'
     service = ArchiveUnlockService(endpoint_name, backend=superdesk.get_backend())
--- a/server/apps/archive/archive_crop.py
+++ b/server/apps/archive/archive_crop.py
@@ -82,7 +82,6 @@ class ArchiveCropService(BaseService):
         """
         width = doc['CropRight'] - doc['CropLeft']
         height = doc['CropBottom'] - doc['CropTop']
-        crop['height']
         doc_ratio = round(width / height, 1)
         spec_ratio = round(crop['width'] / crop['height'], 1)
         if doc_ratio != spec_ratio:
--- a/server/apps/archive/archive_crop.py
+++ b/server/apps/archive/archive_crop.py
@@ -17,6 +17,8 @@ from superdesk.services import BaseService
 from superdesk.errors import SuperdeskApiError
 from superdesk.media.media_operations import crop_image, process_file_from_stream
 from superdesk.upload import UploadService, url_for_media
+from superdesk.metadata.item import CONTENT_TYPE, ITEM_TYPE
+from apps.archive.archive import SOURCE as ARCHIVE
 
 
 class ArchiveCropResource(Resource):
@@ -29,6 +31,7 @@ class ArchiveCropResource(Resource):
         'CropBottom': {'type': 'integer'}
     }
     datasource = {'source': 'archive'}
+    item_methods = []
     resource_methods = ['GET', 'POST', 'DELETE']
     resource_title = endpoint_name
     privileges = {'POST': 'archive', 'DELETE': 'archive'}
@@ -39,7 +42,7 @@ class ArchiveCropService(BaseService):
     def create(self, docs, **kwargs):
         item_id = request.view_args['item_id']
         crop_name = request.view_args['crop_name'].lower()
-        archive_service = get_resource_service('archive')
+        archive_service = get_resource_service(ARCHIVE)
         original = archive_service.find_one(req=None, _id=item_id)
         self._validate_crop(original, crop_name, docs[0])
         self._add_crop(original, crop_name, docs[0])
@@ -48,13 +51,13 @@ class ArchiveCropService(BaseService):
     def delete(self, lookup):
         item_id = request.view_args['item_id']
         crop_name = request.view_args['crop_name'].lower()
-        archive_service = get_resource_service('archive')
+        archive_service = get_resource_service(ARCHIVE)
         original = archive_service.find_one(req=None, _id=item_id)
         self._delete_crop(original, crop_name)
 
     def _validate_crop(self, original, crop_name, doc):
         # Check if type is picture
-        if original.get('type') != 'picture':
+        if original[ITEM_TYPE] != CONTENT_TYPE.PICTURE:
             raise SuperdeskApiError.badRequestError(message='Only images can be cropped!')
 
         # Check if the renditions exists
@@ -66,7 +69,10 @@ class ArchiveCropService(BaseService):
             raise SuperdeskApiError.badRequestError(message='Missing original rendition!')
 
         # Check if the crop name is valid
-        crops = get_resource_service('vocabularies').find_one(req=None, _id='crop_sizes')['items']
+        crops = get_resource_service('vocabularies').find_one(req=None, _id='crop_sizes').get('items')
+        if not crops:
+            raise SuperdeskApiError.badRequestError(message='Crops couldn\'t be loaded!')
+
         crop = self._get_crop_by_name(crops, crop_name)
         if not crop:
             raise SuperdeskApiError.badRequestError(message='Unknown crop name!')
@@ -94,19 +100,13 @@ class ArchiveCropService(BaseService):
         :param crop_name: Crop name
         :return: Matching crop or None
         """
-        try:
-            matching_crops = list(filter(lambda crop: crop.get('name', '').lower() == crop_name.lower(), crops))
-            if len(matching_crops) > 0:
-                return matching_crops[0]
-            return None
-        except:
-            return None
+        return next((c for c in crops if c.get('name', '').lower() == crop_name.lower()), None)
 
     def _delete_crop(self, original, crop_name):
         """ Deletes an existing crop with the given name """
         renditions = original.get('renditions')
         renditions.pop(crop_name, None)
-        get_resource_service('archive').patch(original['_id'], {'renditions': renditions})
+        get_resource_service(ARCHIVE).patch(original['_id'], {'renditions': renditions})
 
     def _add_crop(self, original, crop_name, doc):
         """
@@ -120,15 +120,15 @@ class ArchiveCropService(BaseService):
         renditions = original.get('renditions')
         crop_data = UploadService().get_cropping_data(doc)
         original_image = original.get('renditions').get('original')
-        file = superdesk.app.media.fs(None).find_one(original_image.get('media'))
-        if not file:
+        original_file = superdesk.app.media.get(original_image.get('media'), 'upload')
+        if not original_file:
             raise SuperdeskApiError.badRequestError('Original file couldn\'t be found')
 
         try:
-            cropped, out = crop_image(file, file.filename, crop_data)
+            cropped, out = crop_image(original_file, original_file.filename, crop_data)
             if cropped:
-                renditions[crop_name] = self._save_cropped_image(out, file)
-                get_resource_service('archive').patch(original['_id'], {'renditions': renditions})
+                renditions[crop_name] = self._save_cropped_image(out, original_file)
+                get_resource_service(ARCHIVE).patch(original['_id'], {'renditions': renditions})
             else:
                 raise SuperdeskApiError.badRequestError('Saving crop failed: {}'.format(str(out)))
         except SuperdeskApiError:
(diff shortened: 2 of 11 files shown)


server/apps/archive/__init__.py
```diff
@@ -64,6 +65,10 @@ def init_app(app):
     service = ArchiveLockService(endpoint_name, backend=superdesk.get_backend())
     ArchiveLockResource(endpoint_name, app=app, service=service)
 
+    endpoint_name = 'archive_crop'
```
sivakuna-aap: Can we move it to a different package/sub-package?

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+from superdesk import get_resource_service
+from superdesk.resource import Resource
+from .common import item_url
+from superdesk.services import BaseService
+from superdesk.errors import SuperdeskApiError
+from superdesk.media.media_operations import crop_image, process_file_from_stream
+from superdesk.upload import UploadService, url_for_media
+
+
+class ArchiveCropResource(Resource):
```
sivakuna-aap: Inheriting from Archive resource gives better control on versioning

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+    endpoint_name = 'archive_crop'
+    url = 'archive/<{0}:item_id>/crop/<{0}:crop_name>'.format(item_url)
+    schema = {
+        'CropLeft': {'type': 'integer'},
```
sivakuna-aap: Not a pythonic way. Starts with small case and has underscore in between words.

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+    def delete(self, lookup):
+        item_id = request.view_args['item_id']
+        crop_name = request.view_args['crop_name'].lower()
+        archive_service = get_resource_service('archive')
+        original = archive_service.find_one(req=None, _id=item_id)
+        self._delete_crop(original, crop_name)
+
+    def _validate_crop(self, original, crop_name, doc):
+        # Check if type is picture
+        if original.get('type') != 'picture':
```
sivakuna-aap: Please use named tuple in `content.py` module.

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+    endpoint_name = 'archive_crop'
+    url = 'archive/<{0}:item_id>/crop/<{0}:crop_name>'.format(item_url)
+    schema = {
+        'CropLeft': {'type': 'integer'},
+        'CropRight': {'type': 'integer'},
+        'CropTop': {'type': 'integer'},
+        'CropBottom': {'type': 'integer'}
+    }
+    datasource = {'source': 'archive'}
+    resource_methods = ['GET', 'POST', 'DELETE']
```
sivakuna-aap: are `item_methods` relevant here? If not please initialize it to `[]`.

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+    resource_title = endpoint_name
+    privileges = {'POST': 'archive', 'DELETE': 'archive'}
+
+
+class ArchiveCropService(BaseService):
+
+    def create(self, docs, **kwargs):
+        item_id = request.view_args['item_id']
+        crop_name = request.view_args['crop_name'].lower()
+        archive_service = get_resource_service('archive')
```
sivakuna-aap: Please use constant for `archive` string literal. Use `from apps.archive import SOURCE as ARCHIVE`

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+        original = archive_service.find_one(req=None, _id=item_id)
+        self._validate_crop(original, crop_name, docs[0])
+        self._add_crop(original, crop_name, docs[0])
+        return [original]
+
+    def delete(self, lookup):
+        item_id = request.view_args['item_id']
+        crop_name = request.view_args['crop_name'].lower()
+        archive_service = get_resource_service('archive')
```
sivakuna-aap: Same as above use constant

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+        # Check if the renditions exists
+        if not original.get('renditions'):
+            raise SuperdeskApiError.badRequestError(message='Missing renditions!')
+
+        # Check if the original rendition exists
+        if not original.get('renditions').get('original'):
+            raise SuperdeskApiError.badRequestError(message='Missing original rendition!')
+
+        # Check if the crop name is valid
+        crops = get_resource_service('vocabularies').find_one(req=None, _id='crop_sizes')['items']
```
sivakuna-aap: This will throw `KeyError` if vocabulary is not available. We need to throw `BadRequestError` in this case as well.

sivakuna-aap:
Overall it looks good. But have some minor improvements.

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+            return None
+        except:
+            return None
+
+    def _delete_crop(self, original, crop_name):
+        """ Deletes an existing crop with the given name """
+        renditions = original.get('renditions')
+        renditions.pop(crop_name, None)
+        get_resource_service('archive').patch(original['_id'], {'renditions': renditions})
+
```
sivakuna-aap: Shouldn't we increment the version for this operation? Probably the Operation?

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+        original_image = original.get('renditions').get('original')
+        file = superdesk.app.media.fs(None).find_one(original_image.get('media'))
+        if not file:
+            raise SuperdeskApiError.badRequestError('Original file couldn\'t be found')
+
+        try:
+            cropped, out = crop_image(file, file.filename, crop_data)
+            if cropped:
+                renditions[crop_name] = self._save_cropped_image(out, file)
+                get_resource_service('archive').patch(original['_id'], {'renditions': renditions})
```
sivakuna-aap: Same as above.

server/apps/archive/__init__.py
akintolga: Why?

server/apps/archive/archive_crop.py
akintolga: I'm not sure versioning needed for this.

server/apps/archive/archive_crop.py
akintolga: It has been done like this before. Again you need either consistency or much bigger change to fix the previously done stuff.

sivakuna-aap: We can create a backlog for the existing one. But we can avoid adding more to the technical debt.

server/apps/archive/archive_crop.py
sivakuna-aap: Am also not sure if adding a crop to an article should change the version or not? Better to check with PO?

akintolga: I have done already.

server/apps/archive/archive_crop.py
akintolga: Look, I don't think this creates a technical debt. But it is just me.

server/apps/archive/archive_crop.py
akintolga: Maybe not.

server/apps/archive/archive_crop.py
akintolga: If vocabulary is not available, I don't think we would worry about cropping.

server/apps/archive/archive_crop.py
sivakuna-aap: Whats the decision then? We need versions or no.

server/apps/archive/archive_crop.py
akintolga: Same as above.

server/apps/archive/archive_crop.py
sivakuna-aap: It's ok if we don't want to inform API Client on missing vocabulary. But if vocabulary isn't available that line will throw `KeyError`

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _get_crop_by_name(self, crops, crop_name):
+        """
+        Finds the crop in the list of crops by name
+        :param crops: List of crops
+        :param crop_name: Crop name
+        :return: Matching crop or None
+        """
+        try:
+            matching_crops = list(filter(lambda crop: crop.get('name', '').lower() == crop_name.lower(), crops))
```
amagdas: A list comprehension is  a lot more readable I think.

server/apps/archive/archive_crop.py
amagdas: Either way, we need versioning for all the content, images are content so changes to them should be versioned.

server/apps/archive/archive_crop.py
```diff
@@ -0,0 +1,164 @@
...
+        otherwise creates a new crop with the given parameters
+        :param original: Article to add the crop
+        :param crop_name: Name of the crop
+        :param doc: Crop details
+        :raises SuperdeskApiError.badRequestError
+        """
+        renditions = original.get('renditions')
+        crop_data = UploadService().get_cropping_data(doc)
+        original_image = original.get('renditions').get('original')
+        file = superdesk.app.media.fs(None).find_one(original_image.get('media'))
```
amagdas: superdesk.app.media.get is the proper way yo get the content.

plamut:
Re: POST or PATCH - I think PATCH is slightly better semantically. Cropping an image modifies it, but it's kind of still the same image... or put it differently, a different "version" of the same image. OTOH, my first association with POST-ing would be "hey server, we want to create a new image".

...unless of course each image _rendition_ is treated as an image on its own. But it's more natural to treat images as logical units of visual information, and different renditions are just modified versions of the same logical concept (IMO).

Just my 2 ¢.

server/apps/archive/archive_crop.py
plamut: Just wondering if it is better to use a different variable name than "file", despite this being Python 3.x? `file` is a built-in in Python 2 and that might be a bit confusing for random contributors that mostly work in Python 2.

server/superdesk/media/media_operations.py
```diff
@@ -119,7 +119,7 @@ def process_image(content, type):
 def crop_image(content, file_name, cropping_data):
     if cropping_data:
         file_ext = os.path.splitext(file_name)[1][1:]
-        if file_ext in ('JPG', 'jpg'):
+        if not file_ext or file_ext in ('JPG', 'jpg'):
```
plamut: Tip: in general `set` is a better data structure for checks like these, because of its O(1) time for containment checking.

(but not that it really matters in this case with such a short tuple...)

server/apps/archive/archive_crop.py
akintolga: In this case the crop addition/removal doesn't affect any content value of the image story because it doesn't change the image and it doesn't change any metadata. But the history of the action would definitely be needed.

server/apps/archive/archive_crop.py
akintolga: They are not relevant so they will go. Thanks.

server/apps/archive/archive_crop.py
akintolga: This looks like a good to have change as I don't see a problem in using get_resource_service directly. I am open to other developers' opinion on this as these comments dominates the PRs lately.

server/apps/archive/archive_crop.py
akintolga: Thanks guys, I'll be fixing these two points shortly.

server/apps/archive/archive_crop.py
sivakuna-aap: My point was, if we don't specify `item_methods` explicitly on a resource flask will initialize the resource with default `item_methods`.

akintolga:
I was wondering... Would this PR be the one with record number of comments? I am dazzled :)

amagdas:
Getting there :smiley_cat:

sivakuna-aap:
Might be. At one point I was happy when this PR was closed. But it didn't last long. :disappointed:

server/apps/archive/archive_crop.py
plamut: Which comments, I assume you mean the ones related to using constants instead of string literals?

For these I personally strongly support them, because using pre-defined constants is more robust, resistant to typos (one gets an error immediately), and makes it easier to make changes if needed (one central place vs. across many files).

akintolga: Yes that's what I meant. I do understand the reasoning behind it. But I find it confusing. I can use the enum here. But why comment here but not a line above? There's the same problem with 'crop_name'. Actually line above too: 'item_id'. I think unless we -as project team- come to an agreement to decide on what to do, I find these comments shadowing the actual functionalities posted with the PRs.

plamut: I see your point, thanks for elaborating.

There's no clear-cut answer to that, I personally use my best judgement when such situations arise. The constants and/or enum-ish objects should IMO typically be used in cases similar to the following:

* **application-wide settings**, such as API_URL, DB_STRING, TMP_DIR_NAME, etc.
* arbitrary values that need to be **consistent throughout the application**, e.g. status codes, generic error strings, workflow state names, internal pre-defined limited sets of category/label names (whatever concept a "category" represents), etc.
* values that are reasonably likely to change and appear in multiple places, and we want to make such changes easily at a **single central place**,
* **meaningful aliases** for fixed values, often defined in some standard, e.g. HTTP_OK, KEY_ESC, ERR_SOCKET_TIMEOUT

On the other hand, constants should probably _not_ be (over)used for the following:

* dictionary keys names of "data" objects (i.e. those that represent entities in a system) - the schema of an entity is known, and besides, an object of a custom type should probably be used for that aynway, e.g. Article, SpellChecker, User etc.
* key names for dictionaries holding the request parameters sent by the user - we know what we expect/demand, and since such data must be validated anyway, the validator can perform all the checks for unexpected and/or missing parameters received.

In the end the goal is always to have a more readable and robust code base, but sometimes the opinions vary on where to draw the line between the sensible refactoring and overengineering.

ioanpocol:
If the original is changed should we trigger the regeneration of renditions?

ioanpocol:
:+1:

server/apps/archive/archive_crop.py
akintolga: That's fine.

akintolga:
@ioanpocol Good point. I don't know if it is possible to change the original image or anything in the image? Afaik we can only spike the old image and re-upload/re-fetch a new image.

server/apps/archive/archive_crop.py
akintolga: I see you points. Thanks.

sivakuna-aap:
:+1:

amagdas:
:+1:

PR #922, opened 2015-08-05
chore(core) - move content metadata to superdesk

amagdas:


server/apps/archive/common.py
```diff
@@ -22,13 +22,12 @@
 from superdesk.utc import utcnow, get_expiry_date
 from settings import SERVER_DOMAIN, OrganizationNameAbbreviation
 from superdesk import get_resource_service
-from apps.content import metadata_schema
+from superdesk.metadata.item import metadata_schema
 from superdesk.workflow import set_default_state, is_workflow_state_transition_valid
 import superdesk
 from apps.archive.archive import SOURCE as ARCHIVE
-from apps.content import PACKAGE_TYPE, TAKES_PACKAGE
+from superdesk.metadata.item import PACKAGE_TYPE, TAKES_PACKAGE, not_analyzed
```
petrjasek: not sure takes should be there, any other options?

amagdas: Yup, will move some of the keys to apps and move some stuff from `apps.archive.common.py`

sivakuna-aap: Takes isn't specific to AAP. I think it can be part of chore

petrjasek: yep, but it could be in 'superdesk.metadata.takes'

sivakuna-aap: :+1:

mdhaman: :+1:

ioanpocol:
:+1:

petrjasek:
:+1:

PR #923, opened 2015-08-05
Remove the "filter" parameter from Content API

plamut:
Although useful, the "filter" parameter is not (yet) ready for a production use - we would need to implement at least the parameter syntax validation, make sure it is not tied to a specific backend and probably also limit its features (we might not want to give users the full filtering power the backend provides).

Since this filtering functionality is not required for the API v1 and would require non-trivial amount of additional work, we removed it for now.

Resolves [SD-2859](https://dev.sourcefabric.org/browse/SD-2859).

P.S.: I also did a minor optimization of checking the allowed parameters along the way (used `set` instead of a `tuple`).

### DIFF
--- a/server/publicapi/items/service.py
+++ b/server/publicapi/items/service.py
@@ -50,7 +50,7 @@ class ItemsService(BaseService):
         if req is None:
             req = ParsedRequest()
 
-        allowed_params = ('include_fields', 'exclude_fields')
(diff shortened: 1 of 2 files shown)


ioanpocol:
:+1:

petrjasek:
:+1:

PR #924, opened 2015-08-05
fix(activity): filter activity to not contain erros

petrjasek:
and other system messages. it will only show activity of users
related to content.


ioanpocol:
:+1:

mdhaman:
:+1:

amagdas:
:+1:

PR #930, opened 2015-08-06
[SD-2852] Takes correction

mdhaman:
- refactor for takes publishing.
- Fix the bug where correction of a takes combines the body html from all takes and uses the metadata from the last take.

### DIFF
--- a/server/apps/packages/takes_package_service.py
+++ b/server/apps/packages/takes_package_service.py
@@ -33,7 +33,7 @@ class TakesPackageService():
         takes_package = [package.get(PACKAGE) for package in item.get(LINKED_IN_PACKAGES, [])
                          if package.get(PACKAGE_TYPE)]
         if len(takes_package) > 1:
-            message = 'Multiple takes found for item: {0}'.format(item['_id'])
+            message = 'Multiple takes found for item: {0}'.format(item[config.ID_FIELD])
             logger.error(message)
             raise SuperdeskApiError.forbiddenError(message=message)
         return takes_package[0] if takes_package else None
@@ -58,7 +58,7 @@ class TakesPackageService():
             sequence = self.__next_sequence__(sequence)
             target_ref[SEQUENCE] = sequence
             takes_package[SEQUENCE] = target_ref[SEQUENCE]
-            takes_package[LAST_TAKE] = target['_id']
+            takes_package[LAST_TAKE] = target[config.ID_FIELD]
             main_group[ASSOCIATIONS].append(target_ref)
 
         if link is not None:
@@ -66,7 +66,7 @@ class TakesPackageService():
             link_ref[SEQUENCE] = self.__next_sequence__(sequence)
             main_group[ASSOCIATIONS].append(link_ref)
             takes_package[SEQUENCE] = link_ref[SEQUENCE]
-            takes_package[LAST_TAKE] = link['_id']
+            takes_package[LAST_TAKE] = link[config.ID_FIELD]
             link[SEQUENCE] = link_ref[SEQUENCE]
 
     def __next_sequence__(self, seq):
@@ -136,7 +136,7 @@ class TakesPackageService():
             resolve_document_version(updates, ARCHIVE, 'PATCH', target)
             archive_service.patch(target.get(config.ID_FIELD), updates)
 
-        if not link.get('_id'):
+        if not link.get(config.ID_FIELD):
             self.__copy_metadata__(target, link, takes_package)
             archive_service.post([link])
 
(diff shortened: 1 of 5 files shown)


server/apps/packages/takes_package_service.py
```diff
@@ -199,8 +204,7 @@ def get_first_take_in_takes_package(self, item):
         """
         package = self.get_take_package(item)
         if package:
-            groups = package.get('groups', [])
-            refs = next((group.get('refs') for group in groups if group['id'] == 'main'), None)
+            refs = self.__get_package_refs(package)
```
plamut: I would suggest using a single underscore here, it's already enough to mark a method as "not for public use", and the double leading underscore still doesn't make a method really private.

Double underscores with their implied name mangling are mostly meant to prevent subclasses from accidentally overriding things in a class.

server/apps/publish/archive_publish.py
```diff
@@ -301,33 +304,31 @@ def process_takes(self, updates_of_take_to_be_published, package_id, original_of
...
-                                previous_take['body_html'] + '<br>' + package_updates['body_html']
+            metadata_tobe_copied = ['headline', 'abstract', 'anpa_category', 'pubstatus', 'slugline', 'urgency',
+                                    'subject', 'byline', 'dateline', 'publish_schedule']
 
-                metadata_tobe_copied = ['headline', 'abstract', 'anpa_category', 'pubstatus', 'slugline', 'urgency',
-                                        'subject', 'byline', 'dateline', 'publish_schedule']
+            updated_take = original_of_take_to_be_published.copy()
+            updated_take.update(updates_of_take_to_be_published)
+            metadata_from = updated_take
+            if self.published_state == 'corrected':
```
plamut: Do we have a CONSTANT_ALIAS defined for the state name? If so, I would suggest using it.
(apologies to Tolga in advance, this is a legitimate case IMO :laughing: )

plamut:
Maybe consider adding a few unit tests for the new methods that cover all their edge cases if e2e tests do not do that already, but in general it looks good to me.

:+1:

server/apps/publish/archive_publish.py
sivakuna-aap: I've introduced constant for this. But my pr will take time. If someone wants to introduce then am happy to rebase my pr.

plamut: Would it be difficult for you to extract the relevant commits into a separate branch and submit these changes as a separate PR?
Though it's probably also fine to wait for your (rebased) PR, we can live without the aliases for a bit more I guess. :)

server/apps/packages/takes_package_service.py
```diff
@@ -208,3 +212,58 @@ def get_first_take_in_takes_package(self, item):
...
+            archive_service = get_resource_service(ARCHIVE)
+            query = {'query': {'filtered': {'filter': {'terms': {'_id': takes}}}}}
+            request = ParsedRequest()
+            request.args = {'source': json.dumps(query)}
+            items = archive_service.get(req=request, lookup=None)
+            return items.count() == 0
+
+        return True
+
+    def get_takes_in_take_package(self, takes_package):
```
sivakuna-aap: can we rename this method `get_published_takes()`? Looking at the method name gives an impression that it brings all the takes.

server/apps/publish/archive_publish.py
```diff
@@ -301,33 +304,31 @@ def process_takes(self, updates_of_take_to_be_published, package_id, original_of
...
-                                raise PublishQueueError.previous_take_not_published_error(
-                                    Exception("Take with id {} is not published in Takes Package with id {}"
-                                              .format(previous_take_ref[GUID_FIELD], package[config.ID_FIELD])))
+                package_updates['body_html'] = '<br>'.join(body_html_list)
+            else:
+                package_updates['body_html'] = body_html
 
-                            package_updates['body_html'] = \
-                                previous_take['body_html'] + '<br>' + package_updates['body_html']
+            metadata_tobe_copied = ['headline', 'abstract', 'anpa_category', 'pubstatus', 'slugline', 'urgency',
```
sivakuna-aap: I think this list is used in 2 or more places. If there's no change in both lists can we extract this to TakesPackage module?

sivakuna-aap:
:+1: apart from minor suggestions.

server/apps/publish/archive_publish.py
plamut: Ah, I now see this is used below to check which fields should be copied (`for metadata in metadata_tobe_copied...`). In that case `set` is a better data structure to use for this purpose, because the `in` operator takes a constant time, which is faster than the `list`'s linear time.

amagdas:
:+1:  :heart: :blue_heart: :yellow_heart: :purple_heart:

server/apps/publish/archive_publish.py
mdhaman: well aware of that. but missed refactor that in this iteration.

plamut: It happens, no worries. ;)

server/apps/packages/takes_package_service.py
mdhaman: That's it true. Nothing is private.

I considered single underscore but looking at the implementation I would never subclass it hence I used for double underscore. 

Sure will change to single underscore.

PR #933, opened 2015-08-06
Fix: text item is created when creating a new empty package

plamut:
Resolves [SD-2870](https://dev.sourcefabric.org/browse/SD-2870).

### DIFF
--- a/client/spec/authoring_spec.js
+++ b/client/spec/authoring_spec.js
@@ -2,7 +2,8 @@
 'use strict';
 
 var workspace = require('./helpers/workspace'),
-    authoring = require('./helpers/authoring');
+    authoring = require('./helpers/authoring'),  // TODO: still needed?
+    monitoring = require('./helpers/monitoring');
 
 describe('authoring', function() {
     it('can open item stage', function() {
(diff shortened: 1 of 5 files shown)


amagdas:
Tests :blush:

plamut:
LOL.

Will add an e2e test.

Aca-jov:
Looks fine, except for missing tests :)

pavlovicnemanja92:
Sorry, accidentally closed... Just add tests :+1:

sivakuna-aap:
:+1:

plamut:
I updated the commit with a test. I also added some new instructions to README, because it was not obvious how to run the tests in the first place.

amagdas:
:+1:

plamut:
@pavlovicnemanja92 We'll let you merge this PR to fulfil your long-lasting dream. :wink:

pavlovicnemanja92:
Oh @plamut , my dream come true the moment I write to you "please add some tests" :)

client/spec/authoring_spec.js
```diff
@@ -102,4 +102,29 @@ describe('authoring', function() {
...
+            // 2. Open the "plus icon" menu in subnav bar and click the
+            //    "Empty Package" link in the menu.
+            //
+            // 3. Get the newly created item's info box and check if it
+            //    contains an icon indicating that the item's type is indeed
+            //    "composite".
+            //
+            authoring.openAuthoringHome();
+
+            $('.navbtn .sd-create-btn').click();
```
ioanpocol: Better to user PageObjects style to write the tests, see:
http://www.protractortest.org/#/page-objects

plamut: Makes a lot of sense, I'll refactor. Thanks for the suggestion.

plamut:
Test refactored to use PageObjects.

client/spec/authoring_spec.js
```diff
@@ -102,4 +102,30 @@ describe('authoring', function() {
...
+            //    contains an icon indicating that the item's type is indeed
+            //    "composite".
+            //
+            authoring.openAuthoringHome();
+
+            authoring.navbarMenuBtn.click();
+
+            authoring.newEmptyPackageLink.click().then(function () {
+                return authoring.infoIconsBox.all(
+                    by.css('.filetype-icon-composite')
```
ioanpocol: This selector by css is still on spec file

plamut: Yeah, that's on purpose. The package icon is not some common, always present page element, it's specific to this test case. It's a dynamic thing that might or might not exist as a result of previous actions, and since it also represents the meat of the test, it is referenced explicitly.

FWIW, I saw the same pattern in the existing tests...

ioanpocol: The basic idea is to move all selections on PageObjects. You can create a method that returns the list of elements filtered by the following css filter:  ".filetype-icon-composite".
And yes, some tests don't respect PageObjects pattern. We need to update them, but for sure those tests are not a good example to follow.

plamut: OK, agreed. Can you please also create a ticket to refactor those tests? Thanks!

ioanpocol:
:+1:

plamut:
Updated, the test now uses promises all the way down (the likely reason for element matching to fail?). I also rebased onto latest master, there was some unrelated unit test failing for some unknown reason.

plamut:
OK, so the thing is we don't navigate to the right page. We are supposed to go to `/authoring` but end up on `/workspace` as can be seen in debug output - https://travis-ci.org/superdesk/superdesk/jobs/74591008#L1266
(FWIW, I've tried both `nav()` and `open()` functions from utils, and they both behave the same)

Are there any redirects going on here or what?

Anyway, will continue looking at this on Monday.

client/spec/authoring_spec.js
```diff
@@ -102,4 +102,38 @@ describe('authoring', function() {
         });
 
     });
+
+    describe('subnavigation menu', function () {
+        fit('allows to create a new empty package', function () {
```
mugurrus: f needs to be removed

mugurrus: Sorry, saw later that you're still debugging this.

plamut:
Resolved in #945, a workaround was made to trigger the package creation action through the `/monitoring` view.

PR #935, opened 2015-08-06
fix(authoring): Removed monitoring, ingest and content icons from right sidebar 

pavlovicnemanja92:
Removed configurations for authoring widgest.

I left code, because the same code is used for widgets on dashboard.

SD-2826


amagdas:
:+1:

sivakuna-aap:
:+1:

PR #936, opened 2015-08-06
(imp) Extracted item state to namedtuple and constant

sivakuna-aap:
This improvement extracts the values of item state to a Named Tuple and moves `config.CONTENT_STATE` to `ITEM_STATE`.

Reason for moving `config.CONTENT_STATE` is, few modules doesn't need context and those modules fail to get the key from config.

This improvement doesn't replaces the string literals across all the modules. At the project level it will be implemented as part of [SD-2856](https://dev.sourcefabric.org/browse/SD-2856).

### DIFF
--- a/server/apps/archive/common.py
+++ b/server/apps/archive/common.py
@@ -21,7 +21,7 @@ from superdesk.celery_app import update_key
 from superdesk.utc import utcnow, get_expiry_date
 from settings import OrganizationNameAbbreviation
 from superdesk import get_resource_service
-from superdesk.metadata.item import metadata_schema
+from superdesk.metadata.item import metadata_schema, ITEM_STATE, CONTENT_STATE
 from superdesk.workflow import set_default_state, is_workflow_state_transition_valid
 import superdesk
 from apps.archive.archive import SOURCE as ARCHIVE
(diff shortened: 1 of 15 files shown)


akintolga:
:+1:

server/apps/archive/common.py
```diff
@@ -303,8 +303,8 @@ def update_state(original, updates):
     is changed to 'in-progress'.
     """
 
-    original_state = original.get(config.CONTENT_STATE)
-    if original_state not in ['ingested', 'in_progress', 'scheduled']:
+    original_state = original.get(ITEM_STATE)
+    if original_state not in [CONTENT_STATE.INGESTED, CONTENT_STATE.PROGRESS, CONTENT_STATE.SCHEDULED]:
```
plamut: (minor) Better to use `set()` here instead of a `list`.

server/apps/duplication/archive_move.py
```diff
@@ -73,15 +70,15 @@ def move_content(self, id, doc):
...
+        if not is_workflow_state_transition_valid('submit_to_desk', archived_doc[ITEM_STATE]):
             raise InvalidStateTransitionError()
 
         original = dict(archived_doc)
 
         send_to(doc=archived_doc, desk_id=doc.get('task', {}).get('desc'), stage_id=doc.get('task', {}).get('stage'))
 
-        if archived_doc[config.CONTENT_STATE] not in ['published', 'scheduled', 'killed']:
-            archived_doc[config.CONTENT_STATE] = 'submitted'
+        if archived_doc[ITEM_STATE] not in [CONTENT_STATE.PUBLISHED, CONTENT_STATE.SCHEDULED, CONTENT_STATE.KILLED]:
```
plamut: (minor) Same here, a set (`{first, second, third}`) is a more suitable data structure.

plamut:
:+1:

PR #945, opened 2015-08-10
Plamut sd 2870

ioanpocol:


mugurrus:
:+1:

sivakuna-aap:
:+1:

client/spec/authoring_spec.js
```diff
@@ -178,4 +179,12 @@ describe('authoring', function() {
         expect(authoring.getHistoryItems().count()).toBe(1);
         expect(authoring.getHistoryItem(0).getText()).toMatch(/Published by.*/);
     });
+
+    it('allows to create a new empty package', function () {
+        monitoring.openMonitoring();
+        browser.sleep(500);
```
plamut: (while we are waiting for the build to start...)

I have noticed this practice of using `browser.sleep()` in several tests, but I'm not sure this is the right approach. In general, shouldn't we use protractor's promises returned by triggered actions (`click()`, `setLocation()`, etc.) instead to wait for these actions to complete? Or am I missing something?

Thanks.

plamut:
Seems like the build is passing. I guess navigating to /monitoring instead of /authoring did the trick? I'll try in the other PR, too.

plamut:
A workaround through the monitoring view did the trick, for some reason it was not possible to navigate to `/authoring`. Thanks @ioanpocol

PR #946, opened 2015-08-10
[SD-2909] Can't spike or kill items in normal packages

marwoodandrew:


server/apps/archive/archive_spike.py
```diff
@@ -69,9 +70,17 @@ class ArchiveSpikeService(BaseService):
 
     def on_update(self, updates, original):
         updates[ITEM_OPERATION] = ITEM_SPIKE
+        self._validate_item(original)
         self._validate_take(original)
         self._update_rewrite(original)
 
+    def _validate_item(self, original):
+        # only allow an item to be spiked if it is not a member of a package
+        if original[ITEM_TYPE] != CONTENT_TYPE.COMPOSITE and original.get(LINKED_IN_PACKAGES, None) \
```
sivakuna-aap: To be consistent across its better to throw `SuperdeskApiError.badRequestError()` instead of `ValidationError`.

server/apps/publish/archive_publish.py
```diff
@@ -670,6 +671,11 @@ def __init__(self, datasource=None, backend=None):
         super().__init__(datasource=datasource, backend=backend)
 
     def on_update(self, updates, original):
+        # check if we are trying to kill and item that is contained in normal non takes package
+        if original[ITEM_TYPE] != CONTENT_TYPE.COMPOSITE and original.get(LINKED_IN_PACKAGES, None) \
+                and len([x for x in original.get(LINKED_IN_PACKAGES, []) if x.get(PACKAGE_TYPE, '') == '']):
+            raise ValidationError(['This item is in a package' +
```
sivakuna-aap: Same as above

client/app/scripts/superdesk-archive/module.js
```diff
@@ -86,6 +86,9 @@ define([
                     }
                 }, function(response) {
                     item.error = response;
+                    if (angular.isDefined(response.data._issues['validator exception'])) {
```
sivakuna-aap: Better to check if `response.data._issues` is defined or not as in some failure cases this wouldn't be defined.

server/apps/publish/archive_publish.py
```diff
@@ -670,6 +671,11 @@ def __init__(self, datasource=None, backend=None):
         super().__init__(datasource=datasource, backend=backend)
 
     def on_update(self, updates, original):
+        # check if we are trying to kill and item that is contained in normal non takes package
+        if original[ITEM_TYPE] != CONTENT_TYPE.COMPOSITE and original.get(LINKED_IN_PACKAGES, None) \
```
sivakuna-aap: Other than error message both seem to be doing the same thing. Can we extract it to a function?

server/apps/archive/archive_spike.py
```diff
@@ -69,9 +70,17 @@ class ArchiveSpikeService(BaseService):
...
+                and len([x for x in original.get(LINKED_IN_PACKAGES, []) if x.get(PACKAGE_TYPE, '') == '']):
```
plamut: How long is this list expected to be? Probably pretty short, since it only contains the linked packages?

If you think it's needed, we can get rid of the temporary list by saying `sum(1 for x in ...)`.

server/apps/publish/archive_publish.py
```diff
@@ -670,6 +671,11 @@ def __init__(self, datasource=None, backend=None):
...
+                and len([x for x in original.get(LINKED_IN_PACKAGES, []) if x.get(PACKAGE_TYPE, '') == '']):
```
plamut: Same as above (the temporary list comment).

plamut:
Maybe add docstrings for the changed methods, but it seems good nevertheless. :+1: 

(just please check @sivakuna-aap's comments and make changes if needed)

amagdas:
:+1:

server/apps/archive/archive_spike.py
marwoodandrew: I'd used SuperdeskApiError.badRequestError() but got bound up with trying to present a "nice" error notification, I'll play with it a little more.

server/apps/publish/archive_publish.py
marwoodandrew: Maybe a function in apps/archive/common.py?

sivakuna-aap: Yes

server/apps/archive/archive_spike.py
```diff
@@ -69,9 +69,21 @@ class ArchiveSpikeService(BaseService):
 
     def on_update(self, updates, original):
         updates[ITEM_OPERATION] = ITEM_SPIKE
+        self._validate_item(original)
         self._validate_take(original)
         self._update_rewrite(original)
 
+    def _validate_item(self, original):
+        '''
```
sivakuna-aap: doc string uses double quotes not single

server/apps/archive/archive_spike.py
```diff
@@ -69,9 +69,21 @@ class ArchiveSpikeService(BaseService):
...
+        Raises an exception if the item is linked in a non-take package, the idea being that you don't whant to
+        inadvertently remove thing from packages, this force that to be done as a conscious action.
+        :param original:
+        :return: An excetption or nothing
```
sivakuna-aap: Better to use `raises` instead of `return`

server/apps/archive/common.py
```diff
@@ -415,3 +416,13 @@ def item_schema(extra=None):
     if extra:
         schema.update(extra)
     return schema
+
+
+def is_item_in_package(item):
+    '''
```
sivakuna-aap: Same as above

server/apps/publish/archive_publish.py
```diff
@@ -670,6 +670,10 @@ def __init__(self, datasource=None, backend=None):
         super().__init__(datasource=datasource, backend=backend)
 
     def on_update(self, updates, original):
+        # check if we are trying to kill and item that is contained in normal non takes package
+        if is_item_in_package(original):
+            raise ValidationError(['This item is in a package' +
```
sivakuna-aap: Why this is missed out from being `SuperdeskApiError.badRequestError()`?

marwoodandrew: This flavour of exception is displayed nicely by the UI, badRequestError is not so nice but will change it.

sivakuna-aap:
:+1:

amagdas:
:+1:

server/apps/archive/archive_spike.py
plamut: Technically speaking it can be any string literal, but yes, the convention is to use triple double quotes indeed.

PR #947, opened 2015-08-10
fix(templates): Fix edit templates

ioanpocol:
The are no e2e tests at all for templates so they will come in a future
PR. For now just edit fix.


amagdas:
:+1:

PR #950, opened 2015-08-10
Fix (edit mode on double click)

Aca-jov:
SD-2653


plamut:
Looks good (except for the missing test :laughing: )

ioanpocol:
:+1:

ioanpocol:
The following line on e2e test should be updated to check if the editor is on edit mode:
https://github.com/superdesk/superdesk/blob/master/client/spec/monitoring_spec.js#L176

mugurrus:
:+1:

PR #951, opened 2015-08-10
SD-2682 Fetch news items via Content API by querying for phrases

mugurrus:


server/requirements.txt
```diff
@@ -1,6 +1,6 @@
 Cerberus==0.8.1
 Events==0.2.1
-Eve-Elastic==0.2.16
+Eve-Elastic==0.2.17
```
sivakuna-aap: Is `eve-elastic 0.2.17` released? I didn't find it in the github

plamut:
Uhm, where's the actual code?

server/requirements.txt
petrjasek: it is on [pypi](https://libraries.io/pypi/Eve-Elastic)

mugurrus:
The code is in the eve-elastic datalayer

plamut:
OK, that's good.

Have you checked, though, if the API returns a correct error if more than one phrase is submitted?  Also, how are the queries like `q=term1,"unclosed phrase` treated?

ioanpocol:
:+1:

amagdas:
:+1:

server/publicapi/items/service.py
```diff
@@ -93,7 +94,10 @@ def get(self, req, lookup):
         req.args['filter'] = json.dumps(date_filter)
         self._set_fields_filter(req)  # Eve's "projection"
 
-        return super().get(req, lookup)
+        try:
+            return super().get(req, lookup)
+        except InvalidSearchString:
+            raise BadParameterValueError('invalid search text')
```
plamut: Details on _why_ the search text was invalid would be helpful to the user, please check if we can use something meaningful from the `InvalidSearchString`'s error message.

mugurrus:
Queries like queries like q=term1,"unclosed phrase are treated, the client receives 10002 (BadParameterValueError) and a message showing with parameter caused the error.

plamut:
Ok, that's good then.

(there should still be a test for this, though, IMO this has been merged prematurely - no offence, please, meant constructively).

PR #956, opened 2015-08-11
fix(templates): Fix the creation of content item from a template

ioanpocol:


mugurrus:
:+1:

plamut:
I would add docstrings to the modified methods, but good otherwise. :+1:

amagdas:
:+1:

PR #958, opened 2015-08-11
[SD-2800] Legal archive takes and packages update

sivakuna-aap:
With this pull request the **Text Archive** is replaced with **Published**. 
**After merging we need to remove the data from Mongo and Elastic.**


client/app/scripts/superdesk-archive/directives.js
```diff
@@ -349,6 +349,24 @@
...
+                     * @returns {string}
+                     */
+                    scope.getType = function(item) {
+                        if (item.state === 'spiked') {
+                            return 'spike';
+                        } else if (item._type === 'published' && item.allow_post_publish_actions === true) {
+                            return 'archive';
+                        } else if (item._type === 'published' && item.allow_post_publish_actions === false) {
+                            return 'archived';
+                        }
```
plamut: Just an idea - do we have some kind of an enum-ish object containing all possible item types defined in the client application as well?

client/app/scripts/superdesk-archive/views/media-box-grid.html
```diff
@@ -55,7 +55,7 @@ <h5 ng-class="{'shifted' : item.highlights}">
     <ul ng-if="extras.activity" class="item-functions"
         sd-activity-list
         data-item="item"
-        data-type="{{ spike ? 'spike' : ((type || item._type) === 'published'? 'archive': (type || item._type))}}"
+        data-type="{{getType(item)}}"
```
plamut: I really like this improvement, moving the logic out of the template.

client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -459,13 +459,17 @@
...
+                    } else if (item.state === 'ingested') {
                         return 'ingest';
-                    }
+                    } else {
+                        var isPublished = _.contains(['published', 'killed', 'scheduled', 'corrected'], item.state);
 
-                    return 'archive';
+                        if (isPublished === false || (isPublished === true && item.allow_post_publish_actions === true)) {
+                            return 'archive';
+                        } else if (isPublished === true && item.allow_post_publish_actions === false) {
```
plamut: The explicit  comparisons to `true` and `false` are not really needed, because _.contains already returns a boolean, and we can get rid of those (using `isPublished` and `!isPublished` directly).

server/apps/archived/archived.py
```diff
@@ -0,0 +1,70 @@
...
+
+        query = {'$and': [{'item_id': {'$in': item_ids}}, query_filters[0], query_filters[1]]}
+        return super().get_from_mongo(req=None, lookup=query)
+
+    def on_delete(self, doc):
+        """
+        This method throws exception when invoked on PublishedItemService. Overriding to avoid that.
+        """
+
+        return
```
plamut: Or maybe use `pass`, making the intent even more explicit? When quickly glancing over, this `return` looks like a forgotten fragment of some deleted code (to me at least). Good job with the informative docstring, though!

If preferred, the function body can even be left out entirely, since there is already a docstring in place, making the parser happy.

server/apps/publish/archive_publish_tests.py
```diff
@@ -811,40 +773,95 @@ def get_publish_items(item_id, last_version):
...
             last_published = get_publish_items(published_doc['item_id'], True)
             self.assertEqual(1, last_published.count())
 
-    def test_versions_across_collections_after_publish(self):
-        with self.app.app_context():
-            self.app.data.insert('archive_versions', self.article_versions)
+    def _move_to_archived_and_assert_can_remove_from_production(self, item_id, assert_function, item_to_assert=None):
+        published_service = get_resource_service(PUBLISHED)
+        published_item = list(published_service.get_from_mongo(req=None, lookup={'item_id': item_id}))
+        self.assertEquals(len(published_item), 1)
```
plamut: `assertEquals` --> `assertEqual` (the former is deprecated)

server/apps/publish/archive_publish_tests.py
```diff
@@ -811,40 +773,95 @@ def get_publish_items(item_id, last_version):
...
+        assert_function(published_service.can_remove_from_production(
+            item_to_assert if item_to_assert else published_item))
+
+        return published_item
+
+    def test_can_remove_from_production_first_rule(self):
+        """
+        Test for can_remove_production() for the below:
+            1. When item is published only once. In this case can_remove_production() should return true
+            2. When item is published more than once. In this case can_remove_production() should return false
```
plamut: I would suggest splitting this into two separate tests, each of them testing only a single case.

server/apps/publish/published_item.py
```diff
@@ -472,17 +492,52 @@ def __get_user_name(self, user_id):
...
+        :param package:
+        :return: True if item can be removed from production, False otherwise.
+        """
+
+        item_refs = PackageService().get_residrefs(package)
+        archived_items = list(get_resource_service('archived').find_by_item_ids(item_refs))
+        is_removable = (len(item_refs) == len(archived_items))
+
+        if is_removable:
+            packages_in_archived_items = [p for p in archived_items if p[ITEM_TYPE] == CONTENT_TYPE.COMPOSITE]
```
plamut: Constructing a list is not really needed here, I'd suggest converting it to a generator expression.

server/apps/publish/published_item.py
```diff
@@ -472,17 +492,52 @@ def __get_user_name(self, user_id):
...
+
+            for package in packages_in_archived_items:
+                is_removable = self._can_remove_package_from_production(package)
+                if is_removable is False:
```
plamut: Probably more Pythonic to say `if not is_removable`, `is` is more for comparing object references (although it works here, too).

plamut:
:+1: (the remarks are not show stoppers)

Re: enum-ish object for item types in the client - it probably makes more sense to do that in a separate PR and apply it application-wide.

server/apps/search/__init__.py
```diff
@@ -8,34 +8,41 @@
...
+        super().__init__(datasource=datasource, backend=backend)
+        self._default_repos = ['ingest', 'archive', 'published', 'archived']
 
-    def get_private_filter(self):
-        pass
+        self._private_filters = [{
+            'or': [
+                {'and': [{'term': {'_type': 'ingest'}}]},
+                {'and': [{'exists': {'field': 'task.desk'}},
+                         {'terms': {ITEM_STATE: [CONTENT_STATE.FETCHED, CONTENT_STATE.ROUTED, CONTENT_STATE.DRAFT,
```
akintolga: Does this return the Draft stories from users private workspace as well or does it return only those on the desks?

sivakuna-aap: It returns stories in desk only.

akintolga: I think the search should return the stories from that user's workspace as well. This was how it was before I guess...

sivakuna-aap: At present, the global search doesn't return stories in loggedInUser's workspace. We can take a opinion from PO.

akintolga:
:+1:

server/apps/search/__init__.py
mdhaman: previously, global search used to show from workspace. Did we get an opinion from POs?

amagdas:
:+1: The only complaint is that the task number/description talks about legal archive but this work is related to Text Archive.
Also, most probably we will have to improve/speedup the expired content removal, there was a reason  why we used batch delete, I understand that the flow is more robust/correct now but we will still have to find ways to improve it.

:heart: :blue_heart: :yellow_heart:

sivakuna-aap:
With recent changes for takes and publish workflow the Legal Archive wasn't working as expected. Although the description talks more about text archive both legal and text archive is done in this ticket. I will add more details to either top level story or to the subtask.

sivakuna-aap:
@amagdas @petrjasek @ioanpocol @actionless 
As this PR is merged now, we need to remove the data from Mongo and Elastic. Reason being this PR replaces the Text Archive.

PR #959, opened 2015-08-11
fix(notifications): filter notifications for non-admin users

petrjasek:
for non admin users it will only get content notifications related
to current user, admins will get everything based on settings.


petrjasek:
@ioanpocol

ioanpocol:
:+1:

amagdas:
:+1:

plamut:
:+1:

PR #969, opened 2015-08-13
refactor(editor): integrate findreplace and spellcheck

petrjasek:
instead of kind of stand alone those functions are now integrated
in the editor, so you can switch between those.
you can trigger those via editor service on all registered scopes.

instead of using input event it's now using keydown/up to detect
changes on items which can be only triggered by user.

SD-2793

### DIFF
--- a/client/app/scripts/superdesk-authoring/editor/find-replace.js
+++ b/client/app/scripts/superdesk-authoring/editor/find-replace.js
@@ -21,33 +21,49 @@ function FindReplaceDirective($timeout, $rootScope, editor) {
             scope.to = '';
             scope.from = '';
 
+            /**
+             * Highlight next matching string
+             */
             scope.next = function() {
-                editor.command.next();
+                editor.selectNext();
             };
 
+            /**
+             * Highlight previous matching string
+             */
             scope.prev = function() {
-                editor.command.prev();
+                editor.selectPrev();
             };
 
+            /**
+             * Replace currently highlighted matching string with text
+             */
             scope.replace = function() {
-                editor.command.replace(scope.to || '');
+                editor.replace(scope.to || '');
+                editor.selectNext();
             };
 
+            /**
+             * Replace all matching string with text
+             */
             scope.replaceAll = function() {
-                editor.command.replaceAll(scope.to || '');
+                editor.replaceAll(scope.to || '');
             };
 
             scope.$watch('from', function(needle) {
-                var input = document.getElementById('find-replace-what'),
-                    selectionStart = input.selectionStart,
-                    selectionEnd = input.selectionEnd;
-                editor.command.find(needle);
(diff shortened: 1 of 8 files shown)


client/app/scripts/superdesk-authoring/editor/find-replace.js
```diff
@@ -22,32 +22,36 @@ function FindReplaceDirective($timeout, $rootScope, editor) {
...
             scope.replace = function() {
-                editor.command.replace(scope.to || '');
+                editor.replace(scope.to || '');
+                editor.selectNext();
             };
 
             scope.replaceAll = function() {
-                editor.command.replaceAll(scope.to || '');
+                editor.replaceAll(scope.to || '');
             };
```
plamut: I suggest adding docstrings to the modified methods.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -11,318 +11,485 @@
...
+ * @param {Node} node
+ * @param {numeric} offset
+ * @return {Object} {node: {Node}, offset: {numeric}}
+ */
+function findTextNode(node, offset) {
+    var tree = document.createTreeWalker(node, NodeFilter.SHOW_TEXT);
+    var currentLength;
+    var currentOffset = 0;
+    while (tree.nextNode()) {
+        tree.currentNode.textContent = tree.currentNode.textContent.replace(String.fromCharCode(65279), '');
```
plamut: It's not clear what character `String.fromCharCode(65279)` represents, I suggest creating an alias for it.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -400,48 +567,76 @@ angular.module('superdesk.editor', [])
...
+                editor.registerScope(scope);
+
+                var TYPING_CLASS = 'typing';
+
+                var ARROWS = {
+                    37: 1,
+                    38: 1,
+                    39: 1,
+                    40: 1
+                };
```
plamut: Perhaps it makes sense to freeze the object?

plamut:
Looks more or less good. :+1:

client/app/scripts/superdesk/editor/editor.js
petrjasek: it's not clear to me neither. anyone has an idea?

amagdas: http://stackoverflow.com/questions/6538203/how-to-avoid-echoing-character-65279-in-php-this-question-also-relates-to-java

plamut: I didn't know without googling for it, hence the remark. :)

amagdas:
:100: :trophy: :heart: :blue_heart: :yellow_heart: :green_heart:

PR #978, opened 2015-08-17
fix(packages) set roles when creating an empty package

marwoodandrew:


sivakuna-aap:
:+1: 
If usage of `role` literal is in multiple places then it's better to extract it also as a constant.

amagdas:
:+1:

mdhaman:
:+1:

PR #979, opened 2015-08-17
(fix) PATCH to /api/archive isn't invoked when user changes only Dateline and before autosave

sivakuna-aap:


amagdas:
:+1:

plamut:
Looks OK, although I would add a few tests and docstrings to the changed functions on top of it.

akintolga:
:+1:

PR #983, opened 2015-08-18
[SD-2910] - Package publishing

akintolga:
- [x] Package publishing for simple packages
- [x] Dynamic construction of packages per subscriber
- [x] Publishing non-published text, image and take stories with the package
- [x] Stop digital stories getting published to wire subscribers independent of the formatter
- [x] Stories will be regarded as "Published" even if nothing is saved to publish queue
- [x] Publishing a package with already published stories
- [x] Preventing empty packages getting formed and published
- [x] Publishing nested packages
- [x] Auto updating published packages if any of its published items are updated 
- [x] Validating the killed stories are not part of any package
- [x] Post-publish package actions: updates, kills


server/apps/packages/package_service.py
```diff
@@ -82,11 +82,11 @@ def on_update(self, updates, original):
             self.extract_default_association_data(original, assoc)
 
     def on_updated(self, updates, original):
-        toAdd = {assoc.get(ITEM_REF): assoc for assoc in self._get_associations(updates)}
-        toRemove = [assoc for assoc in self._get_associations(original) if assoc.get(ITEM_REF) not in toAdd]
-        for assoc in toRemove:
+        to_add = {assoc.get(ITEM_REF): assoc for assoc in self._get_associations(updates)}
+        to_remove = [assoc for assoc in self._get_associations(original) if assoc.get(ITEM_REF) not in to_add]
```
plamut: Since we are changing this, I would convert this `list` to a `set()` (faster checks for element presence).

server/apps/packages/package_service.py
```diff
@@ -222,6 +222,50 @@ def get_packages(self, doc_id):
...
+    def remove_ref_from_inmem_package(self, package, ref_id):
+        """
+        Removes the reference with ref_id from non-root groups. If there is nothing left
+        in that group then the group and its reference in root group is also removed.
+        If the removed item was the last item then returns
+        :param package: Package
+        :param ref_id: Id of the reference to be removed
+        :return: True if there are still references in the package, False otherwise
+        """
+        groups_to_be_removed = []
```
plamut: Probably better to use a `set()` instead, unless if the number of groups is small, then it doesn't really make a big difference.

plamut:
Looking good so far, keep going!

server/apps/packages/package_service.py
```diff
@@ -222,6 +229,50 @@ def get_packages(self, doc_id):
...
+        non_root_groups = [group for group in package.get('groups', []) if group.get('id') != 'root']
+        for non_rg in non_root_groups:
+            refs = [r for r in non_rg.get('refs', []) if r.get('residRef', '') != ref_id]
```
sivakuna-aap: Can replace `residRef` with constant `ITEM_REF`

server/apps/packages/package_service.py
```diff
@@ -222,6 +229,50 @@ def get_packages(self, doc_id):
...
+            if len(refs) == 0:
+                groups_to_be_removed.append(non_rg.get('id'))
+            non_rg['refs'] = refs
+
+        if len(groups_to_be_removed) > 0:
+            root_group = [group for group in package.get('groups', []) if group.get('id') == 'root'][0]
+            refs = [r for r in root_group.get('refs', []) if r.get('idRef') not in groups_to_be_removed]
+            root_group['refs'] = refs
+            removed_groups = [group for group in package.get('groups', [])
```
sivakuna-aap: name is misleading.

sivakuna-aap:
:+1: looks good. Use of constants in few places is missing though

server/apps/packages/package_service.py
akintolga: The number of items here won't be that many so I'll keep it like this for the time being, thanks for pointing out.

server/apps/packages/package_service.py
```diff
@@ -138,7 +138,7 @@ def check_that_all_groups_are_referenced_in_root(self, root, groups):
             raise SuperdeskApiError.forbiddenError(message=message)
 
     def check_package_associations(self, docs):
-        for (doc, group) in [(doc, group) for doc in docs for group in doc.get('groups', [])]:
+        for (doc, group) in [(doc, group) for doc in docs for group in doc.get(GROUPS, [])]:
```
plamut: This temporary list here can be turned into a generator expression.

server/apps/packages/package_service.py
```diff
@@ -206,6 +206,13 @@ def check_for_circular_reference(self, package, item_id):
             message = 'Trying to create a circular reference to: ' + item_id
             logger.error(message)
             raise ValidationError(message)
+        else:
+            # keep checking in the hierarchy
+            for d in package.get(LINKED_IN_PACKAGES, []):
+                if 'package' in d:
```
plamut: To save a level of indentation, this condition can be embedded into the `for` loop expression (the next line is a bit long and will benefit from that).

server/apps/packages/package_service.py
```diff
@@ -222,6 +229,50 @@ def get_packages(self, doc_id):
...
+        return True
+
+    def replace_ref_in_package(self, package, old_ref_id, new_ref_id):
+        """
+        Locates the reference with the old_ref_id and replaces with the new_ref_id
+        :param package: Package
+        :param old_ref_id: Old reference id
+        :param new_ref_id: New reference id
+        """
+        non_root_groups = [group for group in package.get(GROUPS, []) if group.get(GROUP_ID) != ROOT_GROUP]
```
plamut: list --> generator expression :)

server/apps/packages/package_service.py
```diff
@@ -266,8 +317,8 @@ def remove_refs_in_package(self, package, ref_id_to_remove, processed_packages=N
             else:
                 updates[SEQUENCE] = new_sequence
                 last_take_group = next(reference for reference in
-                                       next(new_group.get('refs') for new_group in new_groups if
-                                            new_group['id'] == 'main')
+                                       next(new_group.get(ASSOCIATIONS) for new_group in new_groups if
+                                            new_group[GROUP_ID] == MAIN_GROUP)
```
plamut: Just checking - a package always has `groups`, and thus it cannot happen that the `new_groups` list is empty? (which would result in a `StopIteration` error).

server/apps/packages/package_service.py
akintolga: The next line checks if the another item in the list: can we do it with the generators?

server/apps/packages/package_service.py
akintolga: I think a package has always groups. This is pretty much the main type:composite requirement afaik.

marwoodandrew:
:+1: Good work

server/apps/packages/package_service.py
plamut: Hm, what do you mean? Doesn't the next line simply iterate over the `non_root_groups` iterable once?

General answer: yes, it is possible to say `x in generator`, but only if you do it once, because the check exhausts the generator.

```py
>>> g = (a for a in range(5))
>>> g
<generator object <genexpr> at 0x7f909d158500>
>>> 3 in g
True
>>> 3 in g
False
>>> 3 in g
False
```

mdhaman:
:+1:   :blue_heart:  :green_heart:

plamut:
:+1:

PR #984, opened 2015-08-18
refactor(ci): shard e2e tests

petrjasek:


petrjasek:
testing if travis can use sharding at all

server/app.py
```diff
@@ -68,8 +79,6 @@ def get_app(config=None):
         json_encoder=MongoJSONEncoder,
         validator=SuperdeskValidator)
 
-    superdesk.app = app
```
ioanpocol: The superdesk.app package variable is used a few(34) times in code. Probably we have to change the code and use the current application instead.

petrjasek: yep right

client/protractor-conf.js
```diff
@@ -21,6 +23,7 @@ exports.config = {
     },
 
     specs: ['spec/**/*[Ss]pec.js'],
+    //exclude: ['spec/authoring*.js'],
```
plamut: Should be removed?

client/spec/helpers/auth.js
```diff
@@ -25,9 +25,9 @@ function getToken(callback) {
             }
             if (!json.token) {
                 console.log(json);
-                throw new Error('Auth failed');
+                //throw new Error('Auth failed');
```
plamut: Remove completely?

server/scripts/start_e2e_server.sh
```diff
@@ -7,6 +7,6 @@ ELASTICSEARCH_INDEX=superdesk_e2e
 REDIS_URL=redis://localhost:6379/2
 EOF
 
-honcho -e /tmp/testenv run python manage.py users:create -u admin -p admin -e admin@localhost --admin=true
-honcho -e /tmp/testenv start
+#honcho -e /tmp/testenv run python manage.py users:create -u admin -p admin -e admin@localhost --admin=true
```
plamut: Remove comment?

petrjasek:
will create a new PRs for both core/superdesk later

PR #985, opened 2015-08-18
(fix) [SD-2981] Dateline not saved when publishing without initial save

sivakuna-aap:


mdhaman:
Looks like we might need to fix this at the client level not server.

PR #1002, opened 2015-08-25
[SD-3014] Modify list of fields applied with templates

marwoodandrew:
This list of fields applied when creating the template has been unified with the list that is used when creating the template.

### DIFF
--- a/client/app/scripts/superdesk-templates/templates.js
+++ b/client/app/scripts/superdesk-templates/templates.js
@@ -44,8 +44,7 @@
                 'located',
                 'creditline',
                 'ednote',
-                'language',
-                'place'
+                'language'
             ];
 
         this.types = [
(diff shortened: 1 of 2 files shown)


client/app/scripts/superdesk-templates/templates.js
```diff
@@ -21,6 +21,33 @@
...
+                'pubstatus',
+                'description',
+                'body_html',
+                'body_text',
+                'place',
+                'located',
+                'creditline',
+                'ednote',
+                'language',
+                'place'
```
sivakuna-aap: Duplicate

sivakuna-aap:
:+1:

client/app/scripts/superdesk-templates/templates.js
marwoodandrew: Well spotted

akintolga:
:+1:

plamut:
:+1:

PR #1006, opened 2015-08-25
Fix links for viewing a contained item/package in the package preview side bar

plamut:
Resolves [SD-3029](https://dev.sourcefabric.org/browse/SD-3029).

I also replaced some tabs with spaces (to fix the indentation), sorry for a bit less readable diff.

**Note:** No tests for this one, because the European team will soon start refactoring this (as per @akintolga) and the template, links, etc. are very likely to change in the process. We thus viewed this more as a light fix.


akintolga:
:+1:

sivakuna-aap:
:+1:

PR #1011, opened 2015-08-25
SD-2462 As a user I want to see the publish history for an item as part of item history

mugurrus:
SD-2462
As a user I want to see the publish history for an item as part of item
history


eozan:
:+1:

pavlovicnemanja92:
:+1:

PR #1012, opened 2015-08-25
feat(workspace): add top level authoring workspace controller

petrjasek:
this contoller will keep state of authoring within workspace,
so if there is an item edited, if it's maximized, dirty etc.

via keeping this outside ng-view it keeps the state when switching
routes. but only as long as you stay within workspace activities.

SD-2885


plamut:
Looks good IMO, I would just add tests for the `AuthoringWorkspaceController` and docstrings for a few methods where missing.

ioanpocol:
:+1:

amagdas:
:+1:

PR #1018, opened 2015-08-26
[SD-3031] Removal of images when expiring from ingest

marwoodandrew:
It seems that the clean images routine was broken.

Also the images are not getting purged as the cursor had been traversed already when attempting to construct the list of file id's

### DIFF
--- a/server/superdesk/io/commands/update_ingest_tests.py
+++ b/server/superdesk/io/commands/update_ingest_tests.py
@@ -10,7 +10,7 @@
 
 
 import superdesk.io.commands.update_ingest as ingest
-
+import app
 from unittest import TestCase
 from datetime import timedelta
 from nose.tools import assert_raises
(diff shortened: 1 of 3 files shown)


sivakuna-aap:
:+1:

akintolga:
:+1:

server/superdesk/io/commands/update_ingest_tests.py
```diff
@@ -225,6 +225,26 @@ def test_query_getting_expired_content(self):
...
+            remove = RemoveExpiredContent()
+            remove.run(provider.get('type'))
+
+            # only one left in ingest
+            after = service.get(req=None, lookup={})
+            self.assertEqual(1, after.count())
+
+            # all gone
+            current_files = app.media.fs('upload').find()
+            self.assertEqual(0, current_files.count())
```
plamut: In general each unit test case should only test a single use case (i.e. one execution path), resulting in more fine-grained failure messages, pinpointing the exact execution path that doesn't work. I would thus suggest splitting this test into four separate test cases.

In general it's good though.

plamut:
:+1:

PR #1023, opened 2015-08-26
fix(monitoring): Fixed editing packages in monitoring.

eozan:
fix(authoring): Fixed close button for embedded mode.

SD-2889

### DIFF
--- a/client/app/scripts/superdesk-packaging/packaging.js
+++ b/client/app/scripts/superdesk-packaging/packaging.js
@@ -12,14 +12,24 @@
 
     'use strict';
 
-    PackagesService.$inject = ['api', '$q', 'archiveService'];
-    function PackagesService(api, $q, archiveService) {
+    PackagesService.$inject = ['api', '$q', 'archiveService', 'lock', 'autosave'];
+    function PackagesService(api, $q, archiveService, lock, autosave) {
 
         this.groupList = ['main', 'story', 'sidebars', 'fact box'];
 
         this.fetch = function fetch(_id) {
             return api.find('archive', _id).then(function(result) {
                 return result;
+            });
+        };
+
+        this.open = function open(_id, readOnly) {
+            return api.find('archive', _id, {embedded: {lock_user: 1}})
+            .then(function _lock(item) {
+                item._editable = !readOnly;
+                return lock.lock(item);
+            }).then(function _autosave(item) {
+                return autosave.open(item);
             });
         };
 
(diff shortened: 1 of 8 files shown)


ioanpocol:
:+1:

client/app/scripts/superdesk-packaging/packaging.js
```diff
@@ -601,6 +609,43 @@
...
+            scope: {
+                item: '='
+            },
+            link: function(scope, elem, attrs, authoringCtrl) {
+                scope.$watch('item', function(item) {
+                    if (item && item.lockIt){
+                        scope.lock();
+                    } else {
+                        scope.origItem = null;
+                        scope.$applyAsync(function() {
```
plamut: Isn't this call to `$applyAsync()` redundant? The code here is already wrapped in a `$watch`, meaning that it is executed during `$digest`, which runs at least twice anyway to check whether or not any of the `$watch` handlers made a yet another change to the model.

Or am I missing something? It's not really clear to me why we first set `origItem` to `null`, and soon after that to `item'. Thanks for any explanation in advance!

plamut:
There are a couple of docstrings and tests missing, they should be added IMO.

client/app/scripts/superdesk-packaging/packaging.js
eozan: PackagingEmbeddedDirective is slightly changed copy of AuthoringEmbeddedDirective, and that part comes from there. You might be right, I'm not sure why it's done like this, possibly to deal with an unrelated problem, though I can't imagine what at the moment. @petrjasek might know better as he worked on the original directive.

petrjasek: with `$applyAsync` it will only do this in next `$digest` cycle, so `origItem` will be `null` in current cycle and it will destroy `sd-authoring` directive which has `ng-if` on it. so it will effectively reset that directive inside.
not sure if it's needed for packaging, it was done for authoring.

eozan: I will check if it's needed for packaging to be sure.

mugurrus:
Agree with Peter, otherwise seems OK

client/app/scripts/superdesk-packaging/packaging.js
plamut: Ugh, this sounds quite hacky, although I now understand the reason. Is there perhaps a good way to achieve the same in a different way? For example, the `sd-authoring` directive could expose a controller with `reset()` method which can be used by other directives to reset it.

Just thinking aloud, I'm not familiar with all the implementation details...

eozan: If I understand correctly from Petr, fixing this properly would require refactoring sdAuthoring heavily. IMO, this should be handled separately from this task.

plamut: I agree, it should be a separate task - can you please open a ticket so that we don't forget about it? And we can then simply consider this one as "won't fix" in the context of this PR.

ioanpocol:
:+1:

PR #1024, opened 2015-08-26
fix(action dropdown): SD-3041 Added nested dropdown for highlights

pavlovicnemanja92:
and function for calculating future position of dropdown

SD-3041

### DIFF
--- a/client/app/scripts/superdesk/ui/ui.js
+++ b/client/app/scripts/superdesk/ui/ui.js
@@ -331,8 +331,23 @@ define([
                         css.right = Math.max(5, window.innerWidth - pos.left) + 'px';
 
                         dropdown.dropdownMenu.css({opacity: '0'}); // avoid flickering
-                        scope.$applyAsync(function() {
+
+                        scope.$applyAsync(function () {
                             dropdown.dropdownMenu.css(css);
+
+                            /*
+                             * Calculate if there is enough space for showing after the icon
+                             * if not, show it above the icon
+                             */
+                            var windowHeight = window.innerHeight - 30, //Subracting 30 is for submenu bar
+                                dropdownHeight = dropdown.dropdownMenu.outerHeight();
+
+                            if ((windowHeight - pos.top) < dropdownHeight) {
+                                dropdown.dropdownMenu.css({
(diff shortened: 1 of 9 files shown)


client/app/scripts/superdesk/ui/ui.js
```diff
@@ -331,8 +331,23 @@ define([
                         css.right = Math.max(5, window.innerWidth - pos.left) + 'px';
 
                         dropdown.dropdownMenu.css({opacity: '0'}); // avoid flickering
-                        scope.$applyAsync(function() {
+
+                        scope.$applyAsync(function () {
```
plamut: `$evalAsync()` is slightly better IMO, because it might execute in the same $digest cycle if one is already in progress (on the contrary, `$applyAsync()` always triggers a new cycle).

plamut:
Any tests for that, proving that the issue has indeed been fixed? :smile:

client/app/scripts/superdesk/ui/ui.js
pavlovicnemanja92: Yeah, but I need it to be executed in next $digest cycle. I need it to wait for all actions be finished, so it can calculate actual height of dropdown :) Test are comming :)

plamut: Well it does that already, the function is executed first... but yeah, never mind, better to focus on the tests :)

pavlovicnemanja92:
So, I asked @ioanpocol  and @petrjasek about test for this case, and we get to conclusion that here we can't get test work right, because of the screen size. This function use window height as base for calculation position for dropdown menu.

plamut:
Things can be mocked...

```js
>>> window.innerHeight
570
>>> window.innerHeight = 700
700
>>> window.innerHeight
700
```

Or are you talking about e2e tests? Those are indeed ungrateful, we had the issues with them because of the different viewport resolutions before...

If it's indeed to difficult to test this, then it can be skipped I guess (though I'm of course not that easily convinced, you know that :smiley: )

ioanpocol:
:+1:

mugurrus:
:+1:

PR #1026, opened 2015-08-26
Update dashboard.html

tictactatic:
The current text does not work (we don't mention "articles" anywhere else). Plus there could be all sorts of items here down the line. So I am making the first suggestion with this commit.


plamut:
Hm, seems like the templates are not in the template cache, resulting in unexpected GET requests being made, making the unit test to fail. Perhaps rebasing onto the latest code will solve the issue?

(the change itself is fine though :+1: )

PR #1028, opened 2015-08-27
[SD-3046] Change the label for slugline to slugline rather than keyword

marwoodandrew:
Think I got them all!


sivakuna-aap:
:+1:

plamut:
:+1:

PR #1030, opened 2015-08-27
fix(publish) - Saving the version of stories in the groups of packages

akintolga:


plamut:
:+1:

server/apps/publish/archive_publish.py
```diff
@@ -258,6 +259,8 @@ def _publish_package_items(self, package, updates):
                     package_item = super().find_one(req=None, _id=guid)
 
                 subscribers = self._get_subscribers_for_package_item(package_item)
+                PackageService().update_field_in_package(updates, package_item[config.ID_FIELD],
+                                                         'version', package_item['_current_version'])
```
sivakuna-aap: Shouldn't be `_current_version`? Please use the constant `config.VERSION` reason being if we change the name then this will fail.

server/apps/publish/archive_publish.py
```diff
@@ -247,6 +247,7 @@ def _publish_package_items(self, package, updates):
                         # if the item is a package do recursion to publish
                         wanted_keys = ['state', 'operation']
                         sub_updates = dict([(i, updates[i]) for i in wanted_keys if i in updates])
```
mugurrus: Dictionary comprehension would be better here.

mugurrus:
:+1:

server/apps/publish/archive_publish.py
```diff
@@ -258,6 +258,8 @@ def _publish_package_items(self, package, updates):
                     package_item = super().find_one(req=None, _id=guid)
 
                 subscribers = self._get_subscribers_for_package_item(package_item)
+                PackageService().update_field_in_package(updates, package_item[config.ID_FIELD],
```
sivakuna-aap: Sorry for the late notice. `PackageService` is already initialized in `BasePublishService` with variable name `package_service`. Apart from this it's been initialized in multiple places across the module. Can we replace the initialization with `self.package_service`?

sivakuna-aap:
Apart from the minor suggestion :+1:

ioanpocol:
:+1:

server/apps/packages/package_service.py
```diff
@@ -272,6 +272,19 @@ def replace_ref_in_package(self, package, old_ref_id, new_ref_id):
...
 
+    def update_field_in_package(self, package, ref_id, field, field_value):
+        """
+        Locates the reference with the ref_id and replaces field value
+        :param package: Package
+        :param ref_id: reference id
+        :param field: field to be replaced
+        :param field_value: value to be used
+        """
+        non_root_groups = (group for group in package.get(GROUPS, []) if group.get(GROUP_ID) != ROOT_GROUP)
```
ioanpocol: small improvement can be here, list comprehension support multi level for(s) with if statements

plamut: That's also an option indeed, although we need to be careful that readability does not suffer. The latter might happen, for example, when trying to pack a long long and/or complex comprehension, but it needs to be evaluated on a case to case basis. Using an intermediary variable with a meaningful name is a good strategy to fight that

I'm kind of more inclined to use the latter approach in this case, makes the intent more clear IMO.

akintolga: Thanks for your comment Ioan. I agree with Peter that in this instance it would be easier for me to read it if I keep the comprehension separate.

PR #1031, opened 2015-08-27
chore(preview features): Update the beta status for a few features

ioanpocol:
For the following features status has been changed:
    hamburger-> authoring - deleted
    hamburger -> settings -> data consistency - delete (we deleted only
the menu option because probably it will be needed to be reactivated)
    activity and workspace stream - stable
    user-> change avatar: upload from url and take a picture - stable


mugurrus:
:+1:

plamut:
:+1:

PR #1033, opened 2015-08-27
fix(workspace): make item actions menu use sidebar authoring

petrjasek:
refactor authoringWorkspace from controller into a service
so that it can be used also from activities and not only directives.

SD-2885

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -767,6 +767,7 @@
 
     AuthoringDirective.$inject = [
         'superdesk',
+        'authoringWorkspace',
         'notify',
         'gettext',
         'desks',
@@ -784,11 +785,10 @@
         '$window',
         'modal'
     ];
-    function AuthoringDirective(superdesk, notify, gettext, desks, authoring, api, session, lock, privileges,
-                                content, $location, referrer, macros, $timeout, $q, $window, modal) {
+    function AuthoringDirective(superdesk, authoringWorkspace, notify, gettext, desks, authoring, api, session, lock,
+                                privileges, content, $location, referrer, macros, $timeout, $q, $window, modal) {
         return {
-            require: '^sdAuthoringWorkspace',
-            link: function($scope, elem, attrs, workspaceCtrl) {
+            link: function($scope, elem, attrs) {
                 var _closing;
 
                 $scope.privileges = privileges.privileges;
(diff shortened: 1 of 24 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1798,11 +1799,10 @@
...
             templateUrl: 'scripts/superdesk-authoring/views/authoring-container.html',
-            require: '^sdAuthoringWorkspace',
-            scope: true,
-            link: function(scope, elem, attrs, workspaceCtrl) {
-                scope.$watch(workspaceCtrl.getItem, function(item) {
-                    scope.authoring.edit(item, true);
+            scope: {},
+            link: function(scope) {
+                scope.$watch(authoringWorkspace.getItem, function(item) {
+                    scope.authoring.edit(item, !authoringWorkspace.readonly);
```
plamut: We might want do skip this call when `$watcher` is invoked for the first time (due to initialization) and nothing has actually changed yet?

client/app/scripts/superdesk-authoring/tests/authoring_spec.js
```diff
@@ -1030,30 +1027,60 @@ describe('authoring actions', function() {
...
-        }));
-    });
+        authoringWorkspace.close();
+        expect(authoringWorkspace.item).toBe(null);
+        expect(authoringWorkspace.getItem()).toBe(null);
+        expect(superdeskFlags.flags.authoring).toBeFalsy();
+
+        authoringWorkspace.edit(item, true);
+        expect(authoringWorkspace.readonly).toBe(true);
+    }));
```
plamut: Seems like this test case is trying to test to much (`edit()`, `close()`, then `edit()` again), IMO it should probably be split to a couple of smaller independent test cases.

client/app/scripts/superdesk-authoring/tests/authoring_spec.js
```diff
@@ -1030,30 +1027,60 @@ describe('authoring actions', function() {
...
+
+        authoringWorkspace.close();
+        scope.$digest();
+        expect(iscope.authoring.item).toBe(null);
+        expect(iscope.authoring.state.opened).toBe(false);
+
+        authoringWorkspace.edit(item, true);
+        scope.$digest();
+        expect(iscope.authoring.item.lockIt).toBe(false);
+    }));
```
plamut: Same remark as above (splitting the test case).

plamut:
We need more refactorings like that, good work! :+1:

petrjasek:
now including #1023
let's see what travis thinks about it

ioanpocol:
:+1:

PR #1038, opened 2015-08-28
fix(monitoring): Monitoring show nothing after a saved search is adde…

ioanpocol:
…d to the monitoring view and the the monitoring view is deleted


pavlovicnemanja92:
:+1:

sivakuna-aap:
Instead of handling it on client side it would have been better if we can delete the saved search from user preferences when a saved search is deleted.

ioanpocol:
@sivakuna-aap Yes, we can do also the changes on saved search delete but I think those changes are needed as a fall off solution.

petrjasek:
:+1:

PR #1039, opened 2015-08-28
Fix (global search fixed)

Aca-jov:
SD-2987

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/metadata.js
+++ b/client/app/scripts/superdesk-authoring/metadata/metadata.js
@@ -180,8 +180,8 @@ function MetadataWordsListEditingDirective() {
  * @param {String} unique - specify the name of the field, in list item which is unique (qcode, value...)
  *
  */
-MetadataListEditingDirective.$inject = [];
-function MetadataListEditingDirective() {
+MetadataListEditingDirective.$inject = ['metadata'];
+function MetadataListEditingDirective(metadata) {
     return {
         scope: {
             item: '=',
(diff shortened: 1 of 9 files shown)


client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -1,41 +1,27 @@
...
-            </div>
+        <div class="field">
+            <label for="ingest-collection">
+                {{:: 'Ingest' | translate}}
+                <input type="checkbox" id="ingest-collection" ng-model="repo.ingest" ng-checked="repo.ingest">
+            </label>
         </div>
-    </fieldset>
-    <fieldset ng-if="context">
-        <div ng-repeat="provider in providers">
```
sivakuna-aap: Why this option is removed? We need this option to search for images in the external system.

Aca-jov: Go to http://master.dev.superdesk.org/ and in global search in parametres, theres only one radio button visible, the others are not, so we concluded that it can be moved, because one radio button is useeless and the other invinsible too... Pls check and if it is needed, I will return the previous version with radio buttons.

sivakuna-aap: The Image Search Provider isn't configured in master.dev and that's the reason it's not visible. But we need that option.

pavlovicnemanja92:
:+1:

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -195,6 +195,8 @@ function MetadataListEditingDirective() {
         },
         templateUrl: 'scripts/superdesk-authoring/metadata/views/metadata-terms.html',
         link: function(scope) {
+            metadata.subjectScope = scope;
```
petrjasek: this should be also unset on `scope.$destroy` event

ioanpocol:
:+1:

PR #1042, opened 2015-08-28
fix(moonitoring): Fix the issue when has been deleted a stage/desk th…

ioanpocol:
…at has been used on monitoring settings


pavlovicnemanja92:
:+1:

plamut:
Looks good, though the following should be added:

1. Docstrings should be added/updated for the changed functions.
2. A test that proves there is (was) a bug and that the changes have indeed made it pass.

sivakuna-aap:
Entity dependency checks are in place now in superdesk. It would have been better to check with POs once about this requirement. As far as I see there are 3 options:
1. System shouldn't allow to delete a desk/stage when it's used in monitoring
2. The way it's been implemented in this PR
3. System will also take care of deleting the desk from user preferences.

ioanpocol:
@sivakuna-aap 
First solution will not work because user can have private settings, on workspace and on monitoring view widgets.
I propose to do 2 & 3.

pavlovicnemanja92:
:+1:

Aca-jov:
:+1:

PR #1046, opened 2015-08-31
[SD-3032] Allow user to set his/her preferred categories in user profile

plamut:
Ticket [SD-3032](https://dev.sourcefabric.org/browse/SD-3032).

* [x] display a list of available categories in user's profile under Preferences tab with preferred categories' checkboxes selected
* [x] save any changes to selected categories list on the server
* [x] filter the categories offered in the menu in article's head in edit mode
* [x] add "check all", "check none" and "default" buttons
* [x] if no categories are selected, revert to the predefined default set of categories (and warn the user about it before saving)
* [x] UI polishing

Also add an end-to-end test for category filtering? Undecided here... it would effectively just replicate the corresponding unit tests on a higher (browser) level, and wouldn't test e.g. some complex collaboration between various system components.

**Update:**

* [x] Move default categories setting to vocabularies.
* [x] Add e2e tests.


client/app/scripts/superdesk-users/views/user-preferences.html
```diff
@@ -30,6 +33,34 @@
...
+                  <label translate>
+                    {{ :: preferences['categories:preferred'].label }}
+                  </label>
+                  <label translate>
+                    By checking the categories you are interested in, the system
+                    will only display them in a menu for setting the content
+                    item's categories (along with any of its existing categories).
+                  </label>
+
+                  <div ng-repeat="cat in categories track by cat.qcode">
```
sivakuna-aap: `sd-typeahead` directive would have been better instead of displaying all categories at once with check boxes.

plamut: Hm, the checkboxes were explicitly mentioned when discussing this feature, nevertheless we can of course change that.
There is a particular advantage with checkboxes, though, and that is the users immediately see which categories are not selected, allowing them to select them. With the sd-typeahead, on the other hand, they would have to remember the category name in order to type its name and select it.

I'm open to different arguments here, and can change the UI widget if deemed necessary.

sivakuna-aap: OK, sorry wasn't aware of discussion around checkboxes. The ticket doesn't mention about that. 
In any case user has to remember the category otherwise user can't map an article to category in Authoring screen. My point is if the list grows the preferences screen is completely occupied with categories checkboxes.

plamut: Actually, on the Authoring screen if you press the **(+)** button, you already see all the categories you can choose from in the menu - at least that's what I see?

The length argument on the settings is valid, I was thinking about in on my way to the office, and there are two points to consider:

* Preferred categories are the very last setting in the tab, thus it's not possible for a user to overlook a setting because it might be hidden too low. There are also not that many of them, it looks fine to me (on my laptop's resolution, though... and it's only a personal opinion).
* Is it possible to trivially implement "select all/none" buttons? I assume it is, it's only a matter of changing the array holding the choices.

Anyway, thanks for the comment, this kind of feedback is always appreciated. I suggest we discuss the matter with  others when seeing the settings tab in action on the screen.

akintolga:
:+1: Nice job...

client/app/scripts/superdesk-users/users.js
```diff
@@ -1034,6 +1049,19 @@
                         scope.datelinePreview = scope.preferences['dateline:located'].located;
                     });
 
+                    // A list of category codes that are considered preferred
+                    // by default, unless of course the user changes this
+                    // preference setting.
+                    scope.defaultCategories = Object.freeze({
```
sivakuna-aap: Can we drive this through `vocabularies` collection? If this list changes then we need to change the code and deploy.

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -3,8 +3,14 @@
...
-MetadataCtrl.$inject = ['$scope', 'desks', 'metadata', '$filter', 'privileges', 'datetimeHelper'];
-function MetadataCtrl($scope, desks, metadata, $filter, privileges, datetimeHelper) {
+MetadataCtrl.$inject = [
+    '$scope', 'desks', 'metadata', '$filter', 'privileges', 'datetimeHelper',
+    'preferencesService'
+];
+function MetadataCtrl(
+    $scope, desks, metadata, $filter, privileges, datetimeHelper,
+    preferencesService
+) {
```
amagdas: This is pretty weird identation...

client/app/scripts/superdesk-users/views/user-preferences.html
```diff
@@ -30,6 +34,48 @@
...
+                       translate>default</a>
+                  </div>
+
+                  <div ng-repeat="cat in categories track by cat.qcode">
+                    <input
+                        type="checkbox"
+                        name="categories" id="cat_{{::cat.qcode}}"
+                        value="{{cat.qcode}}" ng-model="cat.selected" />
+                     <label for="cat_{{::cat.qcode}}">
+                        {{:: cat.name | translate}}</label>
```
plamut: Hm... is this redundant? Do we translate category names or not?

ioanpocol:
:+1: 
I think that it will help if you add at least 2 e2e tests.

client/app/scripts/superdesk-users/users.js
amagdas: Yes, this needs to be read from vocabulary endpoint.

client/app/scripts/superdesk-authoring/metadata/metadata.js
plamut: What would be a good alternative here? The line was a bit too long and started wrapping already.
(the question also applies to other similar cases as well where lines are _definitely_ too long)

client/app/scripts/superdesk-users/users.js
plamut: Sure, it definitely makes sense. It was just that I didn't know if there exists a better place to hold this definition.

server/apps/preferences.py
```diff
@@ -120,6 +120,13 @@ class PreferencesResource(Resource):
         'category': 'dateline'
     })
 
+    superdesk.register_default_user_preference('categories:preferred', {
```
amagdas: Some behave test would be nice as well, at least to get a feeling how this new preference is supposed to work form API's point of view.

client/app/scripts/superdesk-authoring/metadata/metadata.js
amagdas: I refer to the paranthesis being on their own lines, didn't see this style in the code base till now.

amagdas:
Looks good, please add a couple of e2e.

plamut:
@ioanpocol What scenarios would you suggest? I basically want to avoid a test just for the sake of it that would only replicate what other tests are already testing, it needs to be something that tests non-trivial interaction between the various parts of the system.

server/apps/preferences.py
plamut: Sure, any concrete suggestions for the scenarios perhaps? If not, we'll just think and come up with something sensible tomorrow...

ioanpocol:
@plamut  I think that the happy flow it will be enough. 
Check if the default categories are available on authoring, then set some new categories and check if those new categories are available on authoring.

plamut:
Sound like a plan, thank you!

client/app/scripts/superdesk-authoring/metadata/metadata.js
plamut: I followed the pattern I sometimes see in `if` statements with longer conditionals, for example:
```js
if (foo.property.something || bar.baz.another.something ||
    (sky.is.blue && sun.is.shining) ||
    button.disabled || items.length > 0
) {
    // do something
    // and some more
    // ...
}
```

The goal was to keep the opening curly bracket indented to the left so that the "block head" is easily visually distinguished from the block body. It doesn't look ideal though, thus alternative suggestions welcome.

P.S.: Would it look better if the dependencies were spread more evenly across the lines, e.g. like this?
```js
function MetadataCtrl(
    $scope, desks, metadata, $filter, privileges,
    datetimeHelper, preferencesService
) {
    // ...
};
```

plamut:
The e2e tests have been added, and the default preferred categories setting has been moved to vocabularies.

amagdas:
:+1:

PR #1052, opened 2015-09-01
(fix) Rebuild ngtemplates on html change.

mdhaman:


sivakuna-aap:
:+1:

akintolga:
:+1:  :+1:

plamut:
:+1: :+1: :+1:

marwoodandrew:
:+1: :+1: :+1: :+1:

petrjasek:
@thnkloud9 does it still work for you without `templates.js` in repo? i guess we will have to push this into some extra bower repo

thnkloud9:
@petrjasek - thanks for this update and your help.  No, this would not work for us at the moment.  But, maybe we could generate for tags (releases), or if we are planning on generating builds for releases, we can (and probably should) update our client to use minified dist/ includes?  This would also be much cleaner for vpp and liveblog I think, no?

petrjasek:
@thnkloud9 check #1055

PR #1053, opened 2015-09-01
fix(publish) - Ability to remove from packages in corrections

akintolga:


plamut:
`for _ in range(10000000): print(`:+1: `)`

sivakuna-aap:
:+1:

PR #1055, opened 2015-09-01
chore(bower): put bower files into its own repo

petrjasek:
you can install it now via `bower install superdesk-client`

there is an extra repo https://github.com/superdesk/bower-superdesk-client

does current version work for you? @thnkloud9 @vied12 @nistormihai

if not I can put there files from bower folder (which are removed in this PR) and push this as a new version


amagdas:
:+1:

mugurrus:
:+1:

vied12:
I would prefer to have the distribution files in the same repository as the source, because we need sometimes to use sources (I think about less files that we would like to include)

petrjasek:
ok that's also possible. so `superdesk-templates` will be generated in dist folder instead of app and there will be a copy of it in bower folder, ok @thnkloud9 ?

thnkloud9:
yeah, that sounds ok to me.

petrjasek:
updated, what about now? @thnkloud9 @vied12
btw do you use this repo with `bower` ? should I register it there?

petrjasek:
ping @thnkloud9 @vied12

vied12:
Yes [we use it](https://github.com/superdesk/liveblog/blob/master/client/bower.json#L25) with bower and it would be wonderful to have it directly from bower with the release system !

petrjasek:
ok cool. will register it on bower then

amagdas:
:+1:

amagdas:
Should we merge this?

petrjasek:
I think so

PR #1059, opened 2015-09-01
Sync Eve to upstream

amagdas:


plamut:
This Eve installation issue was causing me problems until I completely deleted the local directory and cloned the repository from scratch. The fix looks good, though I'm not sure what else needs to be done, if anything (the WIP label)?

Giving :+1:  nevertheless.

amagdas:
It looks like some regressions got introduced in upstream, investigating.

PR #1060, opened 2015-09-01
fix(authoring): fix kill/correct to work within authoring sidebar

petrjasek:
authoringWorkspace will keep item id and action to be performed,
this will populate scope for authoring.

also authoring-container is now taking care of authoring opening
so authoring and packaging embedded are not duplicating the code.

SD-3099


client/app/scripts/superdesk-authoring/views/authoring-topbar.html
```diff
@@ -13,13 +13,13 @@
...
         <button id="Edit" class="btn btn-medium pull-left" ng-if="item._type !== 'legal_archive'"
             ng-show="action === 'view' && !item._locked && itemActions.edit"
-            ng-click="lock()"
+            ng-click="edit()"
             translate>Edit</button>
 
         <button class="btn btn-medium pull-left"
             ng-show="action === 'view' && !item._locked && itemActions.deschedule"
             ng-click="deschedule()"
-            translate>Deschedule</button>
```
sivakuna-aap: Any specific reason for removing button label?

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1885,18 +1838,24 @@
...
+            this.edit = function(item, action) {
+                self.item = null;
+                self.action = null;
+                self.state.opened = !!item;
+                if (item) {
+                    authoring.open(item, action === 'view')
+                        .then(function(origItem) {
+                            self.item = origItem;
+                            self.action = action;
+                        });
```
plamut: Docstring is missing for this method.

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -2035,10 +1973,30 @@
...
+
+        /**
+         * Correct an item
+         *
+         * @param {Object} item
+         */
+        this.correct = function(item) {
+            self.item = item._id || null;
+            self.action = 'correct';
+            saveState();
```
plamut: Seems like this action-setting logic repeats itself in several methods, with the only difference being the action name. It probably makes sense to refactor it to a common helper function?

plamut:
Apart from a few comments/suggestion it looks good to me. :+1:

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
petrjasek: my mistake, thx

amagdas:
:+1: :100: :heart: :blue_heart: :yellow_heart:

sivakuna-aap:
:+1:

PR #1063, opened 2015-09-02
(fix) Removed archive_crop endpoint and crop are created as part archive and archive/correct endpoint

mdhaman:


akintolga:
:+1:

amagdas:
:+1: :100:

PR #1064, opened 2015-09-02
feat(packaging): Added adding items to packages in monitoring.

eozan:
SD-2891

### DIFF
--- a/client/app/scripts/superdesk-packaging/packaging.js
+++ b/client/app/scripts/superdesk-packaging/packaging.js
@@ -98,14 +98,11 @@
         };
 
         this.isAdded = function(pkg, item) {
-            for (var i = 0; i < pkg.groups.length; i++) {
-                for (var j = 0; j < pkg.groups[i].refs.length; j++) {
-                    if (pkg.groups[i].refs[j].guid === item._id) {
-                        return true;
-                    }
-                }
-            }
-            return false;
+            return pkg.groups.some(function(group) {
+                return group.refs.some(function(ref) {
+                    return ref.guid === item._id;
+                });
+            });
         };
 
         this.fetchItem = function(packageItem) {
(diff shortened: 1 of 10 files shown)


ioanpocol:
:+1:  some tests are needed

eozan:
tests are WIP

client/app/scripts/superdesk-packaging/packaging.js
```diff
@@ -608,7 +629,7 @@
                             scope.action = 'view';
                         });
                     }
-                });
+                }, true);
```
petrjasek: not sure why you need this recursive watch, but it won't be available soon..
check #1060

client/app/scripts/superdesk-packaging/packaging.js
```diff
@@ -91,11 +92,22 @@
...
 
+        this.isAdded = function(pkg, item) {
+            for (var i = 0; i < pkg.groups.length; i++) {
+                for (var j = 0; j < pkg.groups[i].refs.length; j++) {
+                    if (pkg.groups[i].refs[j].guid === item._id) {
+                        return true;
+                    }
+                }
+            }
+            return false;
```
plamut: Instead of manually iterating over both lists, it's IMO better to simply use the built-in Array methods, making the code more concise and less error prone.

```js
return pkg.groups.some(function (group) {
    return group.refs.some(function (ref) {
        return ref.guid === item._id;
    });
});
```

plamut:
Docstrings should be added for the two modified methods, the rest looks OK. Maybe also check the test coverage if that single test indeed covers all execution paths.

eozan:
Added more tests, but we have some (probably unrelated) problems in monitoring, which causes opened items to be in view mode. When those problems are fixed, tests will pass.

petrjasek:
:+1: but would be good to squash some of those fix commits next time

PR #1065, opened 2015-09-03
Temporary fix chromium donwload URL to allow the e2e tests to run

plamut:


plamut:
FWIW, it probably shouldn't be considered a "temporary" fix, since simply reverting to the channel "stable" also worked, making it a proper fix.

akintolga:
:+1:

marwoodandrew:
:+1:

PR #1067, opened 2015-09-03
feat(search): SD-2986 Added infinite scroll for grid and list view

pavlovicnemanja92:
SD-2986

### DIFF
--- a/client/app/scripts/superdesk-archive/directives.js
+++ b/client/app/scripts/superdesk-archive/directives.js
@@ -512,6 +512,10 @@
             };
         })
 
+        /*
+         * This directive is only temporarly,
+         * it will be deleted with content and ingest
+         */
         .directive('sdContentResults', ['$location', 'preferencesService', 'packages', 'tags', 'asset',
             function ($location, preferencesService, packages, tags, asset) {
                 var update = {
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-search/search.js
```diff
@@ -675,6 +610,21 @@
...
+                    BUFFER: 8
+                },
+                mgrid: {
+                    ITEM_HEIGHT: 239,
+                    ITEM_WIDTH: 190,
+                    ITEMS_COUNT: 27,
+                    ITEMS_ROW: 9,
+                    BUFFER: 18
+                }
+            };
```
plamut: It might make sense to use freezed objects to make the intent (immutable objects containing config) even more explicit, but it's a very minor thing really...

client/app/scripts/superdesk-archive/directives.js
```diff
@@ -512,6 +512,85 @@
             };
         })
 
+        .directive('sdContentResults', ['$location', 'preferencesService', 'packages', 'tags', 'asset',
```
pavlovicnemanja92: This directive is created temporarily, when the development of monitoring will be finished, content and ingest stuffs will be deleted, also as this directive.

petrjasek: would be good to have this info in the source code..

ioanpocol:
:+1:

amagdas:
:+1:

PR #1071, opened 2015-09-04
[SD-3138] - User specific saved searched in monitoring

akintolga:


plamut:
Do we have a test for this, i.e. testing that he controller invokes the API with correct parameters on initialization?

The update looks good nevertheless.

ioanpocol:
The desk monitoring settings is not user related so a saved search should be visible for all users. This was the reason to add all_saved_searches. Please let me know what is the problem/bug that you want to fix with this PR.

akintolga:
This task is closed so this PR no longer needed.

PR #1072, opened 2015-09-04
fix(package publishing) newsML G2 formatter retrieves each package it…

marwoodandrew:
…em to get the metadata. In case the item metadata has changed since the item was put in the package.


akintolga:
:+1:

plamut:
:+1:

PR #1074, opened 2015-09-04
fix(client): only enable ng-view when we get user preferences

petrjasek:
this will prevent rendering incomplete pages to users

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -113,7 +113,6 @@
             'superdesk.archive',
             'superdesk.archive.directives',
             'superdesk.ingest',
-            //'superdesk.scratchpad',
             'superdesk.desks',
             'superdesk.groups',
             'superdesk.authoring',
@@ -133,6 +132,7 @@
             'superdesk.templates',
             'superdesk.monitoring',
 
+            'superdesk.loading',
             'templates'
         ];
 
--- a/client/app/scripts/main.js
+++ b/client/app/scripts/main.js
@@ -34,6 +34,7 @@ define('main', [
         superdesk.constant('config', config);
         superdesk.constant('lodash', _);
 
+        // setup default route for superdesk - set it here to avoid it being used in unit tests
         superdesk.config(['$routeProvider', function($routeProvider) {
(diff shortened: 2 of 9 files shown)


plamut:
Wouldn't it be better to resolve this in the router, so that loading the template would wait until user preferences have been fetched? It would be IMO a cleaner solution than communicating through the `$rootScope` and having the template checking for the (non-)existence of the `preferences` object.

(I'm not aware of all the details of the system though, so that might not be possible...)

petrjasek:
it might be, will check.

plamut:
That would be great. Defining a promise to wait for in the router is the way to go, and let's hope there are no big things that would break if we do this change.

petrjasek:
ping @superdesk/developers

amagdas:
:+1:

ioanpocol:
:+1:

plamut:
Wouldn't it be better to simply add a `resolve`  property?

petrjasek:
@plamut you mean in route registration? that wouldn't stop fetching data in other resolves before user authenticates.

plamut:
Yes, in the route registration, it's cleaner than manually managing and checking on a global flag in root scope. I mean it works, but it's not ideal IMO.

Would it make sense to add some kind of a "has logged in" promise to other resolves? Kind of makes sense if things are not supposed to be loaded before that. But I'm not sure exactly how complicated our routing logic is and how much refactoring would such an improvement require. :)

petrjasek:
problem with that is that it would be just eg:
```
resolve: {auth: auth.promise, users: function(user) { return users.query(); }}
```
but it would be like:
```
resolve: {users: function(auth, users) { return auth.promise.then(function() { return users.query(); }); }
```
so it would be possible, but we would have to update every single resolve.

bth the flag is set on root scope only to be able to display some kind of `loading` indicator that is not there yet, otherwise it would be just local variable in `run` func.

plamut:
I see, it would indeed be verbose, and probably not worth the effort at this point.
Thanks!

PR #1077, opened 2015-09-07
(feat) [SD-3036] Publishing embargo content

sivakuna-aap:
This PR depends on superdesk-core [PR#6](https://github.com/superdesk/superdesk-core/pull/6)

### DIFF
--- a/client/app/scripts/superdesk-archive/views/media-box-grid.html
+++ b/client/app/scripts/superdesk-archive/views/media-box-grid.html
@@ -47,8 +47,7 @@
 
     <span class="type-icon"><i class="filetype-icon-{{item.type}}"></i></span>
     <span class="urgency-label"><i class="urgency-icon-{{item.urgency}}"></i></span>
-
-    <div class="state-label state-{{item.state}}">{{item.state}}</div>
+    <div class="state-label state-{{item.state}}" ng-class="{state_embargo: item.embargo}">{{item.state}}</div>
 </div>
 
 <div class="action">
--- a/client/app/scripts/superdesk-archive/views/media-box-list.html
+++ b/client/app/scripts/superdesk-archive/views/media-box-list.html
@@ -24,7 +24,7 @@
     </div>
     <div class="item-info">
         <div class="line">
-            <div class="state-label state-{{item.state}}" ng-hide="viewType">{{item.state}}</div>
+            <div class="state-label state-{{item.state}}" ng-class="{state_embargo: item.embargo}" ng-hide="viewType">{{item.state}}</div>
             <div class="headline">
                 <span class="keyword" title="{{ item.slugline }}" ng-if="item.slugline">{{ item.slugline | limitTo: 40}}</span>
                 <span id="title" title="{{ item.headline }}">{{ item.headline | limitTo: 90 || item.type }}</span> 
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -29,6 +29,7 @@
         pubstatus: null,
         more_coming: false,
         targeted_for: [],
+        embargo: null,
         renditions: null
     });
 
@@ -128,10 +129,6 @@
             this.stop(item);
             timeouts[item._id] = $timeout(function() {
                 var diff = extendItem({_id: item._id}, item);
-
-                if (diff.publish_schedule === null) {
-                    delete diff.publish_schedule;
-                }
 
                 return api.save(RESOURCE, {}, diff).then(function(_autosave) {
                     extendItem(item, _autosave);
@@ -449,8 +446,9 @@
 
             var lockedByMe = !lock.isLocked(current_item);
 
-            // new take should be on the text item that are closed or last take but not killed.
+            // new take should be on the text item that are closed or last take but not killed and doesn't have embargo.
             action.new_take = !is_read_only_state && (current_item.type === 'text' || current_item.type === 'preformatted') &&
+                !current_item.embargo &&
                 (angular.isUndefined(current_item.takes) || current_item.takes.last_take === current_item._id) &&
                 (angular.isUndefined(current_item.more_coming) || !current_item.more_coming);
 
@@ -469,7 +467,7 @@
                 }
 
                 action.re_write = (_.contains(['published', 'corrected'], current_item.state) &&
-                    _.contains(['text', 'preformatted'], current_item.type) &&
+                    _.contains(['text', 'preformatted'], current_item.type) && !current_item.embargo &&
                     (angular.isUndefined(current_item.more_coming) || !current_item.more_coming));
 
             } else {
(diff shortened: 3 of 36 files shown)


client/app/scripts/superdesk-archive/views/media-box-grid.html
```diff
@@ -47,8 +47,7 @@ <h5 ng-class="{'shifted' : item.highlights}">
 
     <span class="type-icon"><i class="filetype-icon-{{item.type}}"></i></span>
     <span class="urgency-label"><i class="urgency-icon-{{item.urgency}}"></i></span>
-
-    <div class="state-label state-{{item.state}}">{{item.state}}</div>
+    <div class="state-label state-{{item.state}}" ng-class="{state_embargo: item.embargo !== undefined && item.embargo !== null}">{{item.state}}</div>
```
plamut: Would simply `{state_embargo: item.embargo}` also work? Or do we need to explicitly compare it to both `null` and `undefined`?

client/app/scripts/superdesk-archive/views/media-box-list.html
```diff
@@ -24,7 +24,7 @@
     </div>
     <div class="item-info">
         <div class="line">
-            <div class="state-label state-{{item.state}}" ng-hide="viewType">{{item.state}}</div>
+            <div class="state-label state-{{item.state}}" ng-class="{state_embargo: item.embargo !== undefined && item.embargo !== null}" ng-hide="viewType">{{item.state}}</div>
```
plamut: Same question regarding the `ng-class` condition as above.

client/app/scripts/superdesk-archive/views/media-box-grid.html
sivakuna-aap: Yes need both `null` and `undefined`

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -500,9 +499,11 @@
                 !is_read_only_state && current_item.package_type !== 'takes' &&
                  user_privileges.mark_for_highlights);
 
-            // allow all stories to be packaged
+            // allow all stories to be packaged if it doesn't have Embargo
             action.package_item = current_item.state !== 'spiked' && current_item.state !== 'scheduled' &&
-            current_item.package_type !== 'takes' && current_item.state !== 'killed';
+                (angular.isUndefined(current_item.embargo) || _.isNull(current_item.embargo)) &&
```
plamut: Same thought - could perhaps this "embargo condition" be simplified in this line and other similar lines in the code?

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1002,7 +1026,7 @@
                 }
 
                 $scope.publish = function() {
```
plamut: A few docstrings should be added for the modified functions/methods here...

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -120,7 +119,43 @@ function MetadataCtrl(
...
+    }
+
+    /**
+     * Publish Schedule and Embargo are saved as Timestamps in DB but each field has date and time as two different
+     * inputs in UI. This function breaks the timestamp fetched from API to Date and Time and assigns those values to
+     * the appropriate field.
+     */
+    function resolvePublishScheduleAndEmbargoTS() {
+        if ($scope.item.embargo) {
+            var embaroTS = new Date(Date.parse($scope.item.embargo));
```
plamut: Typo? (embaroTS)

Not that it affects the code itself in any way, though, it's more about about the variable naming...

client/app/scripts/superdesk-archive/views/media-box-grid.html
plamut: OK then. I was just curious why `item.embargo` would not work - it's probably a boolean, not an object, right?

server/apps/publish/archive_publish.py
```diff
@@ -828,8 +863,14 @@ def get_subscribers(self, doc, target_media_type):
...
         first_take = None
+
         # Step 1
-        subscribers = list(get_resource_service('subscribers').get(req=None, lookup={'is_active': True}))
+        if doc.get(EMBARGO) and doc.get(EMBARGO) > utcnow():
+            subscribers = list(get_resource_service('subscribers').get(req=None,
+                                                                       lookup={'subscriber_type': SUBSCRIBER_TYPES.WIRE,
+                                                                               'is_active': True}))
+        else:
+            subscribers = list(get_resource_service('subscribers').get(req=None, lookup={'is_active': True}))
```
plamut: It would probably be better to simply add the `subscriber_type` to the lookup dictionary if the condition holds, but otherwise keep the rest of the code in a single statement, without duplicating it across both `if` branches.

server/apps/publish/archive_publish.py
```diff
@@ -1032,7 +1074,12 @@ def get_subscribers(self, doc, target_media_type):
             # step 2
             if not self.takes_package_service.get_take_package_id(doc):
                 # Step 3
-                active_subscribers = get_resource_service('subscribers').get(req=None, lookup={'is_active': True})
+                if doc.get(EMBARGO) and doc.get(EMBARGO) > utcnow():
+                    active_subscribers = list(get_resource_service('subscribers').get(
+                        req=None, lookup={'subscriber_type': SUBSCRIBER_TYPES.WIRE, 'is_active': True}))
+                else:
+                    active_subscribers = get_resource_service('subscribers').get(req=None, lookup={'is_active': True})
```
plamut: Similar suggestion here - we could merge both branches and just add the `subscriber_type` condition to the lookup dict if necessary.

client/app/scripts/superdesk-archive/views/media-box-grid.html
sivakuna-aap: embargo is not a boolean. It's a timestamp

plamut: Hm... but then it is always considered to be a truthy value, unless it is `null` or `undefined`, thus the simplification can indeed be made?

plamut:
Maybe check if the changes on the client side code are sufficiently covered by tests, but in general it looks good (with some less significant comments/suggestions).

:+1:

client/app/scripts/superdesk-archive/views/media-box-grid.html
sivakuna-aap: Yes, the value will be null when user removes the embargo. That's the reason explicitly checking for both.

mdhaman:
:+1:  As discussed, change the expiry of the item to be embargo date + desk/content expiry

akintolga:
:+1:  Looks good

sivakuna-aap:
Resolved all github comments

amagdas:
:+1: Merged the core PR, please update requirements.txt file.

PR #1080, opened 2015-09-07
SD-3054 - add Http Push type of delivery to be used by Content API subscriber

mugurrus:


client/app/scripts/superdesk-publish/views/http-push-config.html
```diff
@@ -0,0 +1,8 @@
+<div class="field">
+    <label for="ftp_host" translate>Resource URL</label>
```
amagdas: 'http_host' ?

client/app/scripts/superdesk-publish/views/http-push-config.html
```diff
@@ -0,0 +1,8 @@
+<div class="field">
+    <label for="http_host" translate>Resource URL</label>
+    <input type="text" id="http_host" placeholder="{{ :: 'Resource URL'|translate }}" ng-model="destination.config.resource_url" required ng-change="$parent.setConfig(destination)">
+</div>
+<div class="field">
+    <label for="http_host" translate>Item URL</label>
+    <input type="text" id="http_host" placeholder="{{ :: 'Item URL'|translate }}" ng-model="destination.config.item_url" required ng-change="$parent.setConfig(destination)">
```
plamut: Would it make sense to defer the model update by a few hundred milliseconds, i.e. when the user has likely finished with typing?

ioanpocol:
It will help to have an e2e and/or behave test that proves that the http push delivery is working well.

plamut:
I agree with @ioanpocol on tests, they are needed.

mugurrus:
An e2e test for http push doesn't make much sense. HTTP push is run in a celery task and pushes content to an external API. The chain of involved components is long and there's not much info you get if the crash happens in the celery task. I think it's better to test for the item insertion in the publish queue.

petrjasek:
:+1:

amagdas:
:+1:

PR #1086, opened 2015-09-08
(fix) Fixing the problem of AWS Storage 'Metadata too Large' exception.

mdhaman:
https://github.com/superdesk/superdesk-core/pull/9

This only affects the AWS storage.  

Maybe change the code https://github.com/superdesk/superdesk/blob/master/server/apps/storage/amazon/amazon_media_storage.py#L118

`self.client.put_object(Key=filename, Body=content, Bucket=self.container_name,
                                   ContentType=content_type, Metadata=None)`


marwoodandrew:
:+1:

sivakuna-aap:
:+1: Need rebasing

PR #1091, opened 2015-09-08
feat(monitoring): Added group selection to "Add to package" action.

eozan:
SD-3172


eozan:
tests are WIP

petrjasek:
would it be possible to do this group picking in the actions dropdown? like marking for highlights?

eozan:
I didn't know it was possible with current activity implementation. I will change it if possible.

eozan:
Updated to use dropdown instead of modal.

client/app/scripts/superdesk-packaging/packaging.js
```diff
@@ -331,9 +331,13 @@
...
-                    for (var i = 0; i < data.length; i++) {
-                        scope.list[groupIndex].items.unshift(packages.getReferenceFor(data[i]));
+                    var groupIndex = _.findIndex(scope.list, {id: data.group});
+                    if (groupIndex === -1) {
+                        scope.list.push({id: data.group, items: []});
+                        groupIndex = scope.list.length - 1;
+                    }
+                    for (var i = 0; i < data.items.length; i++) {
+                        scope.list[groupIndex].items.unshift(packages.getReferenceFor(data.items[i]));
                     }
```
plamut: Maybe we should just use `forEach()` here, it's more expressive and concise?

```js
data.items.forEach(function (item) {
    scope.list[groupIndex].items.unshift(packages.getReferenceFor(item));
});
```

client/app/scripts/superdesk-packaging/views/add-to-package.html
```diff
@@ -0,0 +1 @@
+<div sd-add-package-dropdown></div>
```
plamut: It's maybe better to use `sdAddPackageDropdown` as an element instead of as an attribute - it makes it more obvious that the directive results in a new widget and doesn't just add some extra behavior to a `<div>` element (that's a recommended patterns for directives in general).
```html
<sd-add-package-dropdown />
```
On the other hand, we use attribute-level directives throughout the projects, even when those directives, in fact, create custom UI elements / widgets. If consistency is deemed more important, we can nevertheless leave this as is. 

Just an idea...

client/app/scripts/superdesk-packaging/views/sd-add-package-dropdown.html
```diff
@@ -0,0 +1,7 @@
+<ul>
+    <li ng-repeat="group in groupList">
```
plamut: `track by` ?

client/app/scripts/superdesk-packaging/views/sd-add-package-dropdown.html
```diff
@@ -0,0 +1,7 @@
...
+        <button option="{{group}}" ng-click="select(group);">
+                 <i class="icon-plus"></i>{{group}}
```
plamut: ... and maybe one-time binding, too?

plamut:
The comments on the `sd-add-package-dropdown.html` template should be resolved before merging IMO, while the others are more of a recommendation / idea.

:+1:

ioanpocol:
:+1:

client/app/scripts/superdesk-packaging/packaging.js
eozan: forEach is very slow, it shouldn't be used unless there are no other choices:
https://jsperf.com/array-foreach-vs-for-loop

client/app/scripts/superdesk-packaging/views/add-to-package.html
eozan: We had a decision to use directives as attributes. Unless it's changed, this should stay.

plamut: Ah thanks, didn't know about that. Leave it then as is, of course, if that is the agreed approach.

client/app/scripts/superdesk-packaging/packaging.js
plamut: I doubt this makes any noticeable difference, especially since there are other things whose impact on the performance is orders of magnitude bigger, e.g. communication with the server over a network. IMO readability should have a higher priority, although it's relatively minor and not really a must in this particular case.

client/app/scripts/superdesk-packaging/views/sd-add-package-dropdown.html
```diff
@@ -0,0 +1,7 @@
+<ul>
+    <li ng-repeat="group in groupList track by $index">
```
plamut: Erm, you usually want to track items by their IDs... :)

client/app/scripts/superdesk-packaging/views/add-to-package.html
petrjasek: we can revisit this, it would make things more explicit, and technical reasons for that are no longer valid (older ie support)

plamut: Thanks @petrjasek , good to know that, will keep in mind in the future when creating new directives.
(FWIW, this PR can of course stay as is)

client/app/scripts/superdesk-packaging/views/sd-add-package-dropdown.html
eozan: groupList is an array: ['story', 'sidebar', ...]

plamut: Well, in such case you can simply track by the item itself, provided that items are unique. ;)

But never mind, let's not spend too much time on this anymore, no need to update just because of it. Only some `fit` needs to be removed to make the build pass, and we're done.

client/app/scripts/superdesk-packaging/packaging.js
eozan: 26 million ops/sec vs 1 million ops/sec is a very noticeable difference, at least more than other nitpicking we are doing lately.

plamut: @eozan please, the comment was meant constructively, and was not a blocker for merging in any way, I tried my best to make that clear (please correct me if that was not the case). Using `track by` and one time bindings, on the other hand, _can_ have noticeable effects on the UI performance, and it's important that we pay attention to that (these two things still appear in PRs quite often), esp. because some more heavy parts of the UI sometimes already feel laggy.

26M ops/s vs. 1M ops/s does not make a noticeable difference for typical arrays that don't contain more than e.g. 300 elements (it's less than 0.3 ms), but having too many `$watches` on the page, on the other hand, hinders performance, and that is definitely not nitpicking.

No need for a harsh tone really, does not help with anything, but in case you feel disturbed by anything, I kindly invite you to send an email, so that we can sort it out.


(_"Always assume the best"_ - Micz)

eozan: I think we can agree that this place is for technical discussion, and arguments related to tone or anything else should be discussed elsewhere, and not posted here, at all.

plamut: That is correct, yes.
And just to make it clear, do performance discussions also fall under this, or should these be, generally speaking, discussed outside of PRs?

The rest does indeed not belong here, including using terms carrying a negative connotation (e.g. _nitpicking_) for what is a sincere attempt to make the code base better.

Over & out.

eozan: @plamut, as this PR is merged already, let me clear this up:
"nitpicking" was not directed to you (or anyone in particular, which is why I used "we"), and no specific common fixes (one-time bindings or track by) were named.

petrjasek: imo this part of code should be unit tested, and regarding performance - this iteration will run with 1 or 2 items most of the time, not sure it will ever get to 10. so no matter how slow `forEach` is, it makes no difference.
maybe we should use forEach/some/every/map/filter by default for readability, and in rare cases where it would make a difference we can optimize with for loop with some comment to make it explicit? what others think?
ping @superdesk/developers

amagdas: :+1:  on suing forEach/map/filter for readability, optimization should be handled only when needed

plamut: Readability counts and helps a lot with the general maintainability of the code base, thus it should be made a higher priority IMO.

Optimizations that hinder readability should generally only be used if profiling shows that a particular piece of code indeed represents a bottleneck.

PR #1094, opened 2015-09-08
fix(highlights): SD-3105 Fixed layout issue for highlights

pavlovicnemanja92:
And also added label when there are no items in monitoring group

SD-3105


sivakuna-aap:
:+1:

plamut:
:+1:

PR #1095, opened 2015-09-08
feat(monitoring): add upload btn to create content dropdown

petrjasek:
it was only available in content view, now it should be everywhere

SD-2940


sivakuna-aap:
:+1:

plamut:
:+1:

PR #1096, opened 2015-09-09
[SD-2707] Correct the display of the error reported by macro

marwoodandrew:


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1592,8 +1592,8 @@
                                 authoringWorkspace.close();
                             }
                         }, function(err) {
-                            if (err.data._issues['validator exception']) {
```
sivakuna-aap: We still need this. In other cases we receive error message in validator exception

marwoodandrew: According to Tolga the validation exception is not thrown any more!

sivakuna-aap:
:+1:

ioanpocol:
:+1:

plamut:
Is there a test for this? :smile_cat:

PR #1097, opened 2015-09-09
fix(loading): reload route in next digest

petrjasek:
once user logs in and preferences promise is resolved we're updating
routes to disable those that should not be available because of
permissions or beta flag. to make sure route reload is triggered only
after it's called via `$applyAsync`.


amagdas:
:+1:

ioanpocol:
:+1:

PR #1098, opened 2015-09-09
feat(monitoring): Add the option to show all users saved searches

ioanpocol:


client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -342,9 +341,19 @@
                     current: 'desks'
                 };
 
+                scope.$watch('step.current', function(step) {
```
plamut: Can it happen that `step` is equal to `'searches'` from the very beginning?

If it can and if it is undesired to save searches even when nothing has, in fact, changed, then an additional check is needed here (`$watcher` is automatically run when registered for initialization reasons, even if there is no change).

client/app/scripts/superdesk-desks/views/aggregate-settings.html
```diff
@@ -36,8 +36,11 @@ <h3 translate>Monitoring settings</h3>
 
     <div sd-wizard-step data-title="{{ :: 'Saved Searches' | translate }}" data-code="searches">
         <div class="content">
-            <div class="legend" translate>Select saved searches for view</div>
-            <div class="desk" ng-repeat="search in searches | filter: {'user': user} | orderBy: 'name'">
+            <div class="legend" translate>Select saved searches for view.  
+                <span sd-switch class="pull-right" ng-model="allSavedSearches" 
+                 ng-click="initSavedSearches(allSavedSearches)" title="Show all saved searches"></span>
```
plamut: Should the title be translated?

client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -468,7 +496,7 @@
                             });
                         }
                     });
-
+                    scope.allSavedSearches = false;
```
plamut: Maybe it would be better to rename this flag to `allSearchesSaved`, because the name `allSavedSearches` could be understood as _"list of all saved searches"_ (or maybe it's just me, a non-native speaker :smile: ).

client/app/scripts/superdesk-desks/aggregate.js
ioanpocol: It can't be 'searches' from beginning.

client/app/scripts/superdesk-desks/views/aggregate-settings.html
```diff
@@ -36,8 +36,11 @@ <h3 translate>Monitoring settings</h3>
...
+            </div>
+            <div class="desk" ng-repeat="search in currentSavedSearches | orderBy: 'name'">
```
plamut: Do `search` objects have any unique property to track them by it? Because the `currentSavedSearches` list tends to change. Just a thought....

plamut:
I think it's good, just please check if unnecessarily invoking `initSavedSearches()` might have any undesired effects.

:+1:

client/app/scripts/superdesk-desks/aggregate.js
```diff
@@ -398,6 +407,25 @@
                 };
 
                 /**
+                 * Init searches list with all searches if allSearches is true or
+                 * user saved searches and other user saved searches
+                 * if they are already selected
+                 */
+                scope.initSavedSearches = function(showAllSavedSearches) {
+                    var user = session.identity._id;
+                    scope.showAllSavedSearches = showAllSavedSearches;
```
akintolga: It looks like this assignment is not needed.

akintolga:
:+1:

client/app/scripts/superdesk-desks/aggregate.js
ioanpocol: Don't work well if I deleted it.

PR #1100, opened 2015-09-09
Use latest eve

amagdas:


sivakuna-aap:
:+1:  Any plans to upgrade other frameworks? Status shows they are outdated.

amagdas:
Yup, I'll look into it. Some of them can't be upgrade yet because of dependencies.

server/requirements.txt
```diff
@@ -3,7 +3,5 @@ honcho==0.6.6
 python3-ldap==0.9.8.4
 behave==1.2.5
 wooper==0.4.2
-boto3==1.0.0
 
-git+git://github.com/superdesk/eve@sync-develop-24-june#egg=Eve==0.6.0-dev
```
petrjasek: I think we still need this one, you can't set via setup in `superdesk-core` a git version of `superdesk/eve` can you?

amagdas: Hmm, works fine locally...

PR #1105, opened 2015-09-09
refactor(packaging): use authoring for package editing

petrjasek:
it switches to package edit within authoring, but from outside
there is single edit/view action for any type of item.

SD-3156


client/tasks/options/watch.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
     },
     assets: {
         options: {livereload: true},
-        tasks: ['ngtemplates'],
+        tasks: [''],
```
sivakuna-aap: With this change, does it build template.js in local when there is a change in html?

petrjasek: no, my bad, this should be part of different PR

ioanpocol:
:+1:

amagdas:
:+1:

PR #1116, opened 2015-09-10
fix(authoring): SD-3127 Added tooltips and improved icons and their behavior for theme chooser

pavlovicnemanja92:
SD-3127

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1374,22 +1374,43 @@
                 var DEFAULT_CLASS = 'main-article theme-container';
 
                 scope.themes = authThemes.availableThemes;
-                authThemes.get(scope.key).then(function(theme) {
+                authThemes.get('theme').then(function(theme) {
                     scope.theme = theme;
-                    applyTheme();
+                    if (scope.key === 'theme') {
+                        applyTheme(theme);
+                    }
+                });
+                authThemes.get('proofreadTheme').then(function(theme) {
+                    scope.proofreadTheme = theme;
+                    if (scope.key === 'proofreadTheme') {
+                        applyTheme(theme);
+                    }
                 });
 
-                scope.changeTheme = function(theme) {
-                    scope.theme = theme;
-                    authThemes.save(scope.key, theme);
-                    applyTheme();
+                /*
+                 * Changing predefined themes for proofread and normal mode
+                 *
+                 * @param {string} key Type of theme (proofread or normal)
+                 * @param {object} theme New theme
+                 */
+                scope.changeTheme = function(key, theme) {
+                    scope[key] = theme;
+                    authThemes.save(key, theme);
+                    if (scope.key === key) {
+                        applyTheme(theme);
+                    }
                 };
 
-                function applyTheme() {
+                /*
+                 * Applying a theme for currently selected mode
+                 *
+                 * @param {object} theme New theme
+                 */
+                function applyTheme(theme) {
                     elem.closest('.page-content-container')
                         .children('.theme-container')
                         .attr('class', DEFAULT_CLASS)
-                        .addClass(scope.theme && scope.theme.cssClass);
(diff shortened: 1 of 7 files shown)


client/app/scripts/superdesk-authoring/views/article-edit.html
```diff
@@ -40,13 +40,13 @@
 
 <div class="field">
     <label translate>Dateline</label>
-    <div sd-meta-locators class="dateline-info" ng-show="_editable && !proofread"
+    <div sd-meta-locators class="dateline-info" ng-show="_editable"
```
sivakuna-aap: Any specific reason for removing proofread here? In proofread mode we should show the dateline.text not the dateline fields.

client/app/scripts/superdesk-authoring/views/theme-select.html
```diff
@@ -1,19 +1,31 @@
...
             </li>
         </ul>
+        <ul class="proofread-theme-list">
+            <li>
+                <i class="theme-status" ng-class="{active: key === 'proofreadTheme'}"></i>
+                <div class="menu-label" translate>Choose Proofread Theme</div>
+            </li>
+            <li ng-repeat="t in themes">
+                <button ng-click="changeTheme('proofreadTheme', t)" class="{{t.cssClass}}">
+                    {{t.label}}
```
plamut: One time binding here if theme label is not editable (i.e. does not change)?

client/app/scripts/superdesk-authoring/views/theme-select.html
```diff
@@ -1,19 +1,31 @@
...
                     <i class="icon-ok" ng-if="theme === t"></i>
                 </button>
             </li>
         </ul>
+        <ul class="proofread-theme-list">
+            <li>
+                <i class="theme-status" ng-class="{active: key === 'proofreadTheme'}"></i>
+                <div class="menu-label" translate>Choose Proofread Theme</div>
+            </li>
+            <li ng-repeat="t in themes">
```
plamut: Maybe `track by` theme key in both ng-repeat loops... although it's probably not necessary after all, because the theme list does not tend to change and is predefined in the `AuthoringThemesService`, right?

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1374,22 +1374,32 @@
...
-                scope.changeTheme = function(theme) {
-                    scope.theme = theme;
-                    authThemes.save(scope.key, theme);
-                    applyTheme();
+                scope.changeTheme = function(key, theme) {
+                    scope[key] = theme;
+                    authThemes.save(key, theme);
+                    if (scope.key === key) {
+                        applyTheme(theme);
+                    }
```
plamut: I would also add docstring here for the method, while we are already changing it...

plamut:
What about tests? :)

Looks generally fine to me, though :+1:

client/app/scripts/superdesk-authoring/views/article-edit.html
fritzSF: Proofread mode is just for quickly toggling the theme. There should be no change in functionality.

sivakuna-aap: OK. If that's meant for theme toggling then can we have different name other than proof read? I thought proof read is meant for article proof reading.

fritzSF: It is meant for that, but i don't want to switch back to normal mode for correcting a mistake.

client/app/scripts/superdesk-authoring/views/theme-select.html
pavlovicnemanja92: Yep, themes are hard coded in `authoringThemeService`

client/app/scripts/superdesk-authoring/authoring.js
pavlovicnemanja92: Added :-)

PR #1117, opened 2015-09-10
fix(readme): update link to docker config

### LINKED ISSUE #1103: Set up instructions outdated
Hey guys! Really cool project! Can't wait to get my hands on a public demo! Which is why trying to set up myself :) I am using am Amazon EC2 server, thus I am not using 127.0.0.1, but instance's IP. The README.MD has links to lines in **superdesk/blob/master/docker/docker-compose.yml.sh**, but this file does not exist. Where should I make changes? **superdesk/blob/master/docker/common.yml**? **superdesk/blob/master/docker/run_e2e_tests.sh**? Both? Thanks for your answer!

petrjasek:
update info using `common.yml` file instead of missing `docker-compose.yml.sh`

fix #1103

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,6 @@ Open in a browser `http://localhost:80` after the server will be ready.
 
 If you can't log in then you probably need to run `./scripts/docker-local-create-user.sh`.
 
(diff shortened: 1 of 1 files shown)


amagdas:
:+1:

sivakuna-aap:
:+1:

plamut:
Nice little improvement :+1:

actionless:
it's ok by now

but later it would be better to recommend to create a copy of this file (https://github.com/superdesk/superdesk/blob/master/docker/docker-compose.yml), redefine needed env vars and use this file directly

so that definitely  out of the scope of this change

PR #1120, opened 2015-09-11
fix(dashboard):  Fix dashboard settings for desks(SD-3110)

ioanpocol:


sivakuna-aap:
@ioanpocol can you please mention the ticket number here?

ioanpocol:
SD-3110

client/app/scripts/superdesk-workspace/workspace.js
```diff
@@ -102,12 +102,12 @@
          */
         function getActiveWorkspace() {
             return desks.fetchCurrentDeskId()
-                .then(getActiveWorkspaceId)
+                .then(getActiveWorkspaceId())
```
plamut: Is this indeed correct? `then()` expects a callback that is invoked on success, and not a promise object itself (but a callback can return the latter).

mugurrus:
:+1:

akintolga:
:+1:

PR #1125, opened 2015-09-11
auto create content based on templates

petrjasek:
- [x] add `schedule` to template resource and maintain `next_run` and `last_run` fields
- [x] write a celery task that will create items based on `next_run` field
- [x] add schedule config to template edit modal

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1795,7 +1795,7 @@
                     if (angular.isDefined(item)) {
                         item._datelinedate = '';
 
-                        if (angular.isDefined(item.dateline.located) && !_.isNull(item.dateline.located)) {
+                        if (item.dateline && item.dateline.located != null) {
                             item._datelinedate = $filter('formatDatelinesDate')(item.dateline.located, item.dateline.date);
                         }
                     }
--- a/client/app/scripts/superdesk-templates/templates.js
+++ b/client/app/scripts/superdesk-templates/templates.js
@@ -22,30 +22,30 @@
         var PREFERENCES_KEY = 'templates:recent';
 
         this.TEMPLATE_METADATA = [
-                'headline',
-                'slugline',
-                'abstract',
-                'dateline',
-                'byline',
-                'subject',
-                'genre',
-                'type',
-                'language',
-                'anpa_category',
-                'anpa_take_key',
-                'keywords',
-                'priority',
-                'urgency',
-                'pubstatus',
-                'description',
-                'body_html',
-                'body_text',
-                'place',
-                'located',
-                'creditline',
-                'ednote',
-                'language'
-            ];
+            'headline',
+            'slugline',
+            'abstract',
+            'dateline',
+            'byline',
+            'subject',
+            'genre',
+            'type',
+            'language',
+            'anpa_category',
+            'anpa_take_key',
+            'keywords',
+            'priority',
+            'urgency',
+            'pubstatus',
+            'description',
+            'body_html',
+            'body_text',
+            'place',
+            'located',
+            'creditline',
+            'ednote',
+            'language'
+        ];
 
         this.types = [
             {_id: 'kill', label: gettext('Kill')},
@@ -182,9 +182,11 @@
                 $scope.edit = function(template) {
                     $scope.origTemplate = template || {'type': 'text'};
                     $scope.template = _.create($scope.origTemplate);
+                    $scope.template.schedule = $scope.origTemplate.schedule || {};
 
                     $scope.item = $scope.template;
                     $scope._editable = true;
+                    $scope.updateStages($scope.template.template_desk);
                 };
 
                 $scope.remove = function(template) {
@@ -208,6 +210,10 @@
                     $scope.origTemplate = null;
                     $scope.template = null;
                     $scope.vars = null;
+                };
+
+                $scope.updateStages = function(desk) {
+                    $scope.stages = desk ? desks.deskStages[desk] : null;
                 };
 
                 fetchTemplates();
--- a/client/app/scripts/superdesk/datetime/datetime.js
+++ b/client/app/scripts/superdesk/datetime/datetime.js
@@ -41,5 +41,19 @@ define([
             return function reldate(date) {
                 return moment(date).fromNow();
             };
-        });
+        })
+
+        .factory('weekdays', ['gettext', function(gettext) {
+            return Object.freeze({
+                MON: gettext('Monday'),
+                TUE: gettext('Tuesday'),
+                WED: gettext('Wednesday'),
+                THU: gettext('Thursday'),
+                FRI: gettext('Friday'),
+                SAT: gettext('Saturday'),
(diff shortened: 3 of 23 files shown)


server/apps/rules/routing_rules.py
```diff
@@ -21,6 +21,15 @@
 STATE_ROUTED = 'routed'
 
 
+WEEKDAYS = ['MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN']
+
+
+def parse_time(current_datetime, timestr=None):
+    if timestr is None:
+        timestr = '0000'
+    return current_datetime.replace(hour=int(timestr[:2]), minute=int(timestr[-2:]), second=0)
```
plamut: Would it be better to use `datetime.strptime` for parsing the string? It would, among other things, catch a few more potential format errors of the `timestr`. For example, passing in a time also containing the seconds part, e.g. `22:04:17` would slip through and be recognized as `22:17:00`.

A docstring would be highly useful as well, describing what an expected input is.

server/apps/rules/routing_rules.py
```diff
@@ -21,6 +21,15 @@
 STATE_ROUTED = 'routed'
 
 
+WEEKDAYS = ['MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT', 'SUN']
```
plamut: Would `Enum` be a more suitable data type here? If we decide for it, we can use it even in Python3.3, there exists a backport (called `enum34`).

server/apps/rules/routing_rules.py
petrjasek: thx, good point

server/apps/rules/routing_rules.py
petrjasek: will check the usage, it could be tuple for sure

plamut: A (named)tuple is also an option, I agree.

sivakuna-aap: We can use enum itself as we moved to Python 3.4 with superdesk-core

plamut: Thanks for the update, @sivakuna-aap, I was not aware of that.

(FWIW, this should have been reflected in the `superdesk/server`'s README...)

petrjasek:
used the enums, what do you think? @plamut @sivakuna-aap

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1768,7 +1768,7 @@
                     if (angular.isDefined(item)) {
                         item._datelinedate = '';
 
-                        if (angular.isDefined(item.dateline.located) && !_.isNull(item.dateline.located)) {
+                        if (item.dateline && item.dateline.located != null) {
```
sivakuna-aap: We should check for `undefined of item.dateline.located`. Sometimes it's undefined. Also it should be `item.dateline.located !== null`

plamut: I think the idea here is to compare the value both to `undefined` and `null` at the same time (i.e. in a single comparison), although relying on Javascript's type coercion can be tricky and should probably be avoided by default (not sure if our linter is configured to issue a warning on that, though).

```js
>>> x = undefined
>>> x == null
true
>>> x = null
>>> x == null
true
```
Would it perhaps also be fine to simply check if the value is truthy?

```js
if (... && item.dateline.located) { ...
```

client/app/scripts/superdesk-templates/templates.js
```diff
@@ -210,13 +212,23 @@
...
 
+                $scope.updateStages = function(desk) {
+                    $scope.template.template_stage = null;
+
+                    if (desk) {
+                        $scope.stages = desks.deskStages[desk];
+                    } else {
+                        $scope.stages = null;
+                    }
+                };
```
plamut: Maybe add docstring here?

client/app/scripts/superdesk-templates/templates.js
```diff
@@ -226,6 +238,8 @@
         this.types = templates.types;
         this.save = save;
 
+        console.log('in!');
```
plamut: Probably not needed anymore?

client/app/scripts/superdesk/datetime/datetime.js
```diff
@@ -41,5 +41,19 @@ define([
...
+        .factory('weekdays', ['gettext', function(gettext) {
+            return Object.freeze({
+                MON: gettext('Monday'),
+                TUE: gettext('Tuesday'),
+                WED: gettext('Wednesday'),
+                THU: gettext('Thursday'),
+                FRI: gettext('Friday'),
+                SAT: gettext('Saturday'),
+                SUN: gettext('Sunday')
+            });
```
plamut: :heart:

client/app/scripts/superdesk/ui/views/weekday-picker.html
```diff
@@ -0,0 +1,6 @@
+<button class="btn btn-light pull-left"
+    ng-repeat="day in weekdayList"
+    ng-click="toggleDay(day)"
+    ng-class="{active: isDayChecked(day)}">
+    {{ weekdays[day] }}
```
plamut: Does the name of the day already get translated when defining the `weekdays` factory and calling `gettext()`?

server/apps/rules/routing_rules.py
```diff
@@ -21,6 +22,49 @@
...
+    SAT = 5
+    SUN = 6
+
+    @classmethod
+    def is_valid_schedule(cls, list_of_days):
+        """Test if all days in list_of_days are valid day names.
+
+        :param list list_of_days eg. ['mon', 'tue', 'fri']
+        """
+        return all([day.upper() in cls.__members__ for day in list_of_days])
```
plamut: Maybe just convert the temporary list to a generator expression, but it's very minor.

server/apps/rules/routing_rules.py
```diff
@@ -21,6 +22,49 @@
...
+
+    @classmethod
+    def is_scheduled_day(cls, today, list_of_days):
+        """Test if today's weekday is in schedule.
+
+        :param datetime today
+        :param list list_of_days
+        """
+        return today.weekday() in [cls[day.upper()].value for day in list_of_days]
```
plamut: This `Weekdays` enum is _very_ Pythonic! :heart:

client/app/scripts/superdesk-authoring/authoring.js
petrjasek: exactly our linter only allows `!=` and `==` when comparing to `null`, which makes it work with `undefined` as well. otherwise it must be `===`

client/app/scripts/superdesk/ui/views/weekday-picker.html
petrjasek: yep

amagdas:
:+1:  :heart: :blue_heart: :yellow_heart:

amagdas:
Rebase needed.

ioanpocol:
:+1:

PR #1129, opened 2015-09-14
[sd-3226]-(fix) - After applying crops user cannot save item without …

sjunaid:
…making other changes

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1821,7 +1821,7 @@
                     });
 
                     superdesk.intent('edit', 'crop',  scope.item).then(function(data) {
-                        if (mainEditScope.hasOwnProperty('dirty') && !mainEditScope.dirty) {
+                        if (!mainEditScope.dirty) {
                             mainEditScope.dirty = data.isDirty;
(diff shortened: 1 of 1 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1820,8 +1821,8 @@
                     });
 
                     superdesk.intent('edit', 'crop',  scope.item).then(function(data) {
-                        if (!scope.$parent.dirty) {
-                            scope.$parent.dirty = data.isDirty;
+                        if (mainEditScope.hasOwnProperty('dirty') && !mainEditScope.dirty) {
```
plamut: The `hasOwnProperty()` should not be here. If we assume that we are using the correct scope, it will always evaluate to `true`), making it redundant, but if we cannot assume that, this check should result in an error IMO. As it currently stands, it would silently result in a bug (the `dirty` flag will never get set there).

ioanpocol:
:+1: 
At least one e2e test for cropping is needed.

mugurrus:
:+1:

mdhaman:
@sjunaid is working on another PR which include e2e test for cropping.

PR #1130, opened 2015-09-14
Unify ingest Routing Filters and Publish Filters into Content Filters

plamut:
Resolves [SD-3136](https://dev.sourcefabric.org/browse/SD-3136).

This pull requests creates a new content type called Content Filters, which can be used for, well, filtering content. The main change is that these filters are now decoupled from the actions the system can take based on them, meaning that the same content filter can be used for publish filtering and ingest routing, for instance.

The existing publish filters have been moved into its own app, renamed to Content Filters, and placed under a new tab in Settings.

**Refactoring steps:**
* [x] move ex- Publish Filters to its own content type Content Filters
* [x] move Content Filter setting under its own tab in Settings
* [x] refactor Publishing to use the new Content Filters instead of the old Publish Filters
* [x] refactor ingest routing to use the new Content Filters
* [x] misc. cleanup
* [x] NEW: Prevent deleting a content filter if it is referenced by at least one subscriber and/or routing rule.
* [x] NEW: Add additional checks (if needed) in routing and publishing when acting upon a content filter, because it might have been deleted during the operation.

**Sidenotes:**
* This will most likely break the existing publish filters and routing rules. Since no DB migration script is planned, care must be taken when merging this in order to not break any instances meant for e.g. showcasing.
* Content Filters are, for the most part, just the old Publish Filters code moved to a new place and adjusted a bit. A few minor bugs have been discovered and fixed along the way, but in general any bugs that migh have existed in the old code are probably still present in the new code - the test coverage wasn't that great and writing new tests to cover everything is outside the scope of this PR.
* For some reason, loading the `superdesk.content_filters` module in tests fails to load due to a _"Unknown provider: superdeskProvider"_ error. This happens if any taste case uses the `inject()` function. Seems like something needs to be fixed in config, but for now declaring the `superdesk.publish` module as a dependency is a reasonable workaround, even though it is not strictly necessary.


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -864,118 +897,95 @@ define([
...
+         *
+         * Taken from http://stackoverflow.com/a/3561711/5040035
+         *
+         * @method escape
+         * @param {string} s - the string to escape
+         * @return {string} - an escaped version of the given string
+         */
+        // XXX: should probably be moved into some utils module - but where?
+        RegExp.escape = function (s) {
+            return s.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
```
amagdas: Some usage examples an expected output will help in reviewing/validating the correctness of the regex.

amagdas:
There are still some TODO's lingering in the code, looks good until now.

client/app/scripts/superdesk-ingest/module.js
plamut: The author took the list of Regex special characters from MDN, and what this does is simply replacing all these characters with their escaped versions, e.g. `'+'` becomes `'\\+'`. I can update the docstring to explain the idea better, good point.

(and as for the correctness of the "special characters" set - I trust that the info on MDN is correct :) ).

plamut:
Yup, that's why this PR is still WIP. ;) They are mostly the notes for myself so that I don't forget things that still need to be done, but I deferred them until the final stage, so that they don't steal the focus from the main refactoring things that still need to be done.

ioanpocol:
:+1: 
At least one e2e test for content filters management is needed.
Is it an option to unify content filters with saved searches? And/or use content filters on monitoring view?

plamut:
@ioanpocol All the existing e2e tests for publish filters have been moved with the code as well, so the test coverage in this respect has remained the same (even the UI is almost exactly the same, with just some labels changed).

I am not sure how exactly the saved searches work, but it's IMO doable, since the saved search is basically just a content filter used in the search context. The main advantage this change will bring is that content filters will be decoupled from the actions based on these filters, so yes, with some refactoring they should be usable at various other places in the application (and interchangeable between them).

sivakuna-aap:
@ioanpocol at the moment, the content filters is not an option for saved searches/filters monitoring view. This is for unifying the filters being used in Routing and Publishing. That doesn't mean it can't be extended. @akintolga did I miss something?

plamut:
@sivakuna-aap That is correct, this PR will not refactor the saved searches/filters monitoring - but it will make refactoring them possible.

ioanpocol:
@plamut  I proposed to add a few e2e tests because on this PR I didn't seen any e2e tests.
Re: unifying content filters with saved search is just something that we have to evaluate if it is a good idea or not. For sure this potential refactoring is not in the scope of this PR.

plamut:
@ioanpocol Hm, the thing is that the test coverage of publish filters was already low (both for e2e and unit tests), and removing this technical debt might be a bit too much to include in this PR. The general strategy was to preserve the state of the original code that was just moved to another place, and add tests only for the new code and the code that was changed.

I can improve the coverage of the existing code, of course (needs to be done at some point anyway), I'm just not sure whether it should be done in this PR. If we decide to add additional tests, we should IMO do it properly and thoroughly - adding just a few tests could give us a false sense of accomplishment, even though those tests wouldn't cover much.

What do you other guys think? (ping @superdesk/superdesk-developers )

petrjasek:
for such refactoring I don't think there are new e2e required, it's better tested with unittests

akintolga:
For the saved searches, I think it is out of the scope for this task. So I suggest we try to complete this task with this scope first.
For the coverage, there are unit tests are in place for most of the scenarios. I think it may be good enough to add missing ones for this refactoring.

plamut:
Unfortunately there are not that many unit tests either (for the ex- publish filters at least), it would take quite some extra work to bring the coverage to sufficient levels. On the client side at least, server-side things are a bit better I think (and there weren't that many changes introduced).

ioanpocol:
Re: e2e tests, my point was that for this large functionality we don't have even one e2e tests and because of this I think that we have to start writing them. We can add them in this PR or in other, is not so important, but ASAP we need to add a few e2e tests on content filters.

plamut:
I think we all agree on that, tests are indeed needed, there is no doubt about that. The question, though, was more about whether to do it as a part of this PR or in a separate task.

What is the final consensus on this then, guys? Thanks!

amagdas:
Please update the reference to core.

plamut:
Yup, already planned. Thanks for the reminder nevertheless!

client/app/scripts/superdesk-ingest/module.js
plamut: Resolved (GitHub did not detect that, though, as the updates appeared a couple of lines above the original comment).

petrjasek:
:+1:

plamut:
The PR is now ready IMO. Here's an approximate list of changes since the last batch of comments:
* Fixed failing end-to-end tests,
* Resolved a couple of additional issues (mostly minor) found during refactoring,
* Resolved the remaining TODOs in code, improved commentfor the regex escape method

akintolga:
I suggest adding two extra behave tests:
- Deleting a content filter fails if it is referenced by a routing rule
- Deleting a content filter fails if it is referenced by a subscriber

Apart from that it looks pretty good... 
:+1:

plamut:
@akintolga It seems like it is actually possible to delete a referenced filter - I will fix that and add the suggested tests on top of it.
Please also update the ticket's acceptance criteria for the QA, thanks!

server/apps/prepopulate/app_initialization_test.py
```diff
@@ -47,8 +47,6 @@ def data_scaffolding_test(self):
         self.assertEqual(cursor.count(), existing_desks * stories_per_desk)
 
     def test_app_initialization_index_creation(self):
-        result = app.data.mongo.pymongo(resource='users').db['users'].index_information()
```
amagdas: Why is this removed?

plamut: This was causing the test to incorrectly fail in Mongo 3.0.6 (did not test other versions FWIW). After a discussion with @mdhaman (the author of this test) we determined that this is most likely due to a changed Mongo behavior. It used to return an empty dictionary if a collection did not exist (`'users'` in this case), but in 3.0.6 it returns `None` and calling a method on it fails of course.

@mdhaman then said that this line is not actually needed for the test and suggested to simply remove it, since the essential part of the test lies a few lines below.

amagdas: K

plamut:
Updates from yesterday:
* It is now also not possible to delete a content filter if it is referenced by a routing scheme or a subscriber (besides the existing check if a filter is referenced by other content filters).
* Added additional check for a non-existing filter in `FilterConditionService`'s `does_match()` method. The latter might happen if a routing scheme is fetched and its referenced filters retrieved, but somebody deletes a filter right between these two DB queries.

amagdas:
:+1:

PR #1132, opened 2015-09-14
fix(multipanel): SD-2979 Close authoring when opening multiedit items

pavlovicnemanja92:
SD-2979


ioanpocol:
:+1:

sivakuna-aap:
:+1: Do we have e2e test for this?

pavlovicnemanja92:
@sivakuna-aap I think no, because this was a bug :) Authoring screen was staying open after multiedit window is called .

PR #1135, opened 2015-09-15
feat(monitoring): SD-3123 Added fetch as panel on monitoring view

pavlovicnemanja92:
SD-3123

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1435,9 +1435,9 @@
             }
         };
     }
-    SendItem.$inject = ['$q', 'api', 'desks', 'notify', 'authoringWorkspace',
+    SendItem.$inject = ['$q', 'api', 'desks', 'notify', 'authoringWorkspace', 'superdeskFlags',
         '$location', 'macros', '$rootScope', 'authoring', 'send', 'spellcheck', 'confirm'];
-    function SendItem($q, api, desks, notify, authoringWorkspace,
+    function SendItem($q, api, desks, notify, authoringWorkspace, superdeskFlags,
         $location, macros, $rootScope, authoring, send, spellcheck, confirm) {
         return {
             scope: {
@@ -1474,6 +1474,10 @@
                 }
 
                 function activateItem(item) {
+                    if (scope.mode === 'monitoring') {
+                        superdeskFlags.flags.fetching = !!item;
+                    }
+
                     scope.isActive = !!item;
                     activate();
                 }
@@ -1489,6 +1493,10 @@
                 }
 
                 scope.close = function() {
+                    if (scope.mode === 'monitoring') {
+                        superdeskFlags.flags.fetching = false;
+                    }
+
                     if (scope.$parent.views) {
                         scope.$parent.views.send = false;
                     } else {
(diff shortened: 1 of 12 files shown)


akintolga:
:+1:

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1431,6 +1431,8 @@
             },
             templateUrl: 'scripts/superdesk-authoring/views/send-item.html',
             link: function sendItemLink(scope, elem, attrs) {
+                var parentScope = scope.$parent.$parent;
```
plamut: If I understand correctly, this grandparent scope is only referenced to obtain a reference to the `monitoring` controller used by the `MonitoringViewDirective`? If so, this pattern should be avoided, because it is fragile and sensitive to changes in the scope hierarchy.

The right way to do it is to declare that this directive depends on the parent `MonitoringViewDirective` (`require: '^sdMonitoringView'`), and then the `monitoring` controller will be passed to the `link()` function as its 4th argument:

```js
function sendItemLink(scope, elem, attrs, ctrl) {
    // ctrl === MonitoringController
```

mdhaman: agree with @plamut .

pavlovicnemanja92: Yeah, I know, and in first way I do it like this. But (always some but), this directive is also used for sending item on another desk in authoring view. As there is no `sdMonitoringView` directive found by require in authoring screen, it throws error. 

This way was my alternative.

plamut: Hm... so how does the `monitoring` controller end up in the parent scope then, what places it there? Would it be possible to somehow reference that thing and obtain the controller reference from it?

@petrjasek @amagdas Are there any other alternatives to obtaining the controller from the grandparent scope?

pavlovicnemanja92: `Monitoring Controller` is accessible with `require` on monitoring page. But it is not accessible on authoring page. So, as this two pages use same directive for sending item, on monitoring view works fine, but on authoring screen it throws an error. Because of that I added `if (parentScope.monitoring) {` .

plamut: Yeah, that's clear and perfectly fine... I'm just wondering if there exists a better way, that's all :)

petrjasek: seems like it's only setting some css classes - you can use `superdeskFlags` service for that:
```
superdeskFlags.flags.fetching = !!item;
```
it will add fetching class to `#main-container`

pavlovicnemanja92: Oh, great :) ... I didn't saw this service :)

client/app/scripts/superdesk-authoring/views/send-item.html
```diff
@@ -8,7 +8,7 @@ <h3 translate ng-show="mode === 'archive'">Duplicate</h3>
         <div class="content">
             <div class="publish-pane-box">
                 <div sd-toggle-box data-title="{{ 'Destination' | translate }}" data-open="true"
-                     ng-if="mode === 'ingest' || (mode === 'authoring' && $root.isActionAllowed(item, 'move') && itemActions.send)">
+                     ng-if="mode === 'ingest' || mode === 'monitoring' || (mode === 'authoring' && $root.isActionAllowed(item, 'move') && itemActions.send)">
```
amagdas: Please extract long conditionals in the controller with a proper function name.

ioanpocol:
:+1: 
One e2e test is needed

ioanpocol:
:+1:

amagdas:
:+1:

PR #1137, opened 2015-09-16
[SD-3229] - Real time currency changes

akintolga:


mdhaman:
:+1:

sivakuna-aap:
:+1:

server/macros/currency_base.py
```diff
@@ -0,0 +1,42 @@
...
+    return float(r.text.split(',')[1])
+
+
+def do_conversion(item, rate, prefix, search_param):
+    """
+    Performs the conversion
+    :param item: story
+    :param rate: exchange rate
+    :param prefix: to be used in the results
+    :param search_param: search parameter to locate the original value
```
plamut: IMO it should be made a bit more clear that `search_param` should, in fact, be a valid regular expression pattern, and not just an arbitrary string, for example.

amagdas:
:+1:

PR #1139, opened 2015-09-16
fix(packaging): fix open item from package button

petrjasek:
it should open item in view mode

SD-3167


ioanpocol:
:+1:

amagdas:
:+1: Rebase needed

PR #1144, opened 2015-09-16
fix(monitoring): On change current desk selection apply the current s…

ioanpocol:
…et search and file type filters to the new desk monitoring view


amagdas:
:+1:

mugurrus:
:+1:

PR #1151, opened 2015-09-17
fix(monitoring): refresh monitoring on content changes

petrjasek:
there is single event `content:update` with info about item, desk
and stage which client can use to figure out what to update in monitoring.

SD-3204 SD-3213

### DIFF
--- a/server/apps/content/__init__.py
+++ b/server/apps/content/__init__.py
@@ -1,0 +1,33 @@
+"""Content related helpers and utils.
+"""
+
+from superdesk.notification import push_notification
+from apps.auth import get_user
+
+
+def push_content_notification(items):
+    """Push content:update notification for multiple items.
+
+    It can be also 2 versions of same item in updated handler
+    so that we sent event with both old and new desk/stage.
+
+    :param list items: list of items
+    """
+    ids = {}
+    desks = {}
+    stages = {}
+    for item in items:
+        ids[str(item.get('_id', ''))] = 1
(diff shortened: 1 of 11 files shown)


server/apps/content/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+        ids[str(item.get('_id', ''))] = 1
+        task = item.get('task', {})
+        if task.get('desk'):
+            desks[str(task.get('desk', ''))] = 1
+        if task.get('stage'):
+            stages[str(task.get('stage', ''))] = 1
+    user = flask.g.get('user', {})
+    push_notification(
+        'content:update',
+        user=str(user.get('_id', '')),
```
sivakuna-aap: can we use `common.py.get_user()` here?

sivakuna-aap:
:+1:

server/apps/content/__init__.py
petrjasek: yep I just didn't want to import from archive here, so I will move the func here and import it into archive

petrjasek: actually I will move it into `apps.auth`, imo it should be there

sivakuna-aap: yeah makes sense

mdhaman:
:+1:

ioanpocol:
:+1:

amagdas:
:+1: Rebase needed

PR #1153, opened 2015-09-18
[SD-3209] - Display crops in the preview and edit

akintolga:


client/app/scripts/superdesk-archive/directives.js
```diff
@@ -384,6 +384,20 @@
...
+
+        .directive('sdItemCrops', ['metadata', function(metadata) {
+            return {
+                templateUrl: 'scripts/superdesk-archive/views/item-crops.html',
+                scope: {
+                    item: '='                
+                },
+                link: function(scope, elem) {
+                    metadata.initialize().then(function() {
+                        scope.metadata = metadata.values;
```
sivakuna-aap: Since we use only crops from metadata.values it would be good to assign only crops instead of whole metadata.values.

client/app/scripts/superdesk-archive/views/item-crops.html
```diff
@@ -0,0 +1,4 @@
+<div ng-repeat="crop in metadata.crop_sizes | orderBy:'name'" ng-if="item.renditions[crop.name]">
+    <span translate>{{crop.name}}<span>
+    <div class="rendition" sd-item-rendition data-item="item" data-ratio="ratio" data-rendition="{{crop.name}}"></div>
+</div>
```
sivakuna-aap: `track by` and one time binding

amagdas:
:+1:

mugurrus:
:+1:

client/app/scripts/superdesk-archive/views/item-crops.html
```diff
@@ -0,0 +1,4 @@
+<div ng-repeat="crop in crop_sizes | orderBy:'name' track by crop.name" ng-if="item.renditions[crop.name]">
+    <span translate>{{crop.name}}<span>
+    <div class="rendition" sd-item-rendition data-item="item" data-ratio="ratio" data-rendition="{{:: crop.name}}"></div>
```
plamut: Can `crop.name` change or not? In this line it is one-time bound, but in the line above it is not - which of the two should be updated?

plamut:
I assume tests are already in place?

:+1:

PR #1156, opened 2015-09-18
fix(templates): fix template auto creation

petrjasek:
which was failing after first run, because it was using id from
template for item. now this field is removed.

SD-2099

### DIFF
--- a/server/apps/templates/content_templates.py
+++ b/server/apps/templates/content_templates.py
@@ -11,17 +11,17 @@
 import re
 import datetime
 import superdesk
+from flask import current_app as app
 from superdesk import Resource, Service
 from superdesk.utc import utcnow
 from superdesk.errors import SuperdeskApiError
-from superdesk.metadata.item import metadata_schema
-from apps.rules.routing_rules import Weekdays, set_time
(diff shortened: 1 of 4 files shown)


server/apps/templates/content_templates.py
```diff
@@ -143,16 +143,15 @@ def set_template_timestamps(template, now):
...
 def create_item_from_template(template):
-    """Create item in production using data from given template.
+    """ Create item using data from template.
 
     :param dict template
     """
-    production = superdesk.get_resource_service('archive')
     item = {key: value for key, value in template.items() if key in metadata_schema}
+    item['template'] = item.pop('_id')
     item['state'] = SCHEDULED_ITEM_STATE
```
sivakuna-aap: Can we use constants for `state`, `archive`? Also, `SCHEDULED_ITEM_STATE` is defined in `item.py.CONTENT_STATE`

petrjasek: right updated it. it shouldn't be scheduled state after all, but probably submitted - it's a new item on a desk that needs some work, not scheduled for publishing.

sivakuna-aap:
:+1:

amagdas:
:+1:

PR #1159, opened 2015-09-18
feat: vocabularies' settings

actionless:
2015-09-18--1442609701_2125x1396_scrot
UPD:
2015-09-19--1442616395_1730x1261_scrot

### DIFF
--- a/client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
+++ b/client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
@@ -14,47 +14,70 @@ describe('vocabularies', function() {
         expect(api.query).toHaveBeenCalledWith('vocabularies');
     }));
 
-    describe('config modal directive', function() {
-        var scope;
-        var testItem;
+    describe('config modal', function() {
 
-        beforeEach(inject(function($rootScope, $controller) {
-            scope = $rootScope.$new();
-            testItem = {foo: 'flareon', bar: 'beedrill'};
-            scope.vocabulary = {items: [testItem]};
-            $controller('VocabularyEdit', {$scope: scope});
-        }));
+        describe('model', function() {
+            var scope;
 
-        it('can add items', function() {
-            scope.addItem();
-            expect(scope.vocabulary.items.length).toBe(2);
-            expect(scope.vocabulary.items[1]).toEqual({foo: null, bar: null});
-        });
-
-        it('can remove items', function() {
-            scope.addItem();
-            scope.removeItem(testItem);
-            expect(scope.vocabulary.items.length).toBe(1);
-            expect(scope.vocabulary.items[0]).toEqual({foo: null, bar: null});
-            scope.removeItem({foo: null, bar: null});
-            expect(scope.vocabulary.items.length).toBe(0);
-        });
-
-        it('can save vocabulary', inject(function(api, $q, $rootScope) {
-            scope.addItem();
-            scope.vocabulary.items[1].foo = 'feraligatr';
-            scope.vocabulary.items[1].bar = 'bayleef';
-            scope.closeVocabulary = function() {};
-            spyOn(api, 'save').and.returnValue($q.when());
-            spyOn(scope, 'closeVocabulary').and.returnValue($q.when());
-            scope.save();
-            $rootScope.$digest();
-            expect(api.save).toHaveBeenCalledWith('vocabularies', {
-                items: [
+            beforeEach(inject(function($rootScope, $controller) {
+                scope = $rootScope.$new();
+                scope.vocabulary = {items: [
                     {foo: 'flareon', bar: 'beedrill'},
-                    {foo: 'feraligatr', bar: 'bayleef'}
-                ]
+                    {bar: 'bellsprout', spam: 'sandslash'},
+                    {qux: 'quagsire', foo: 'frillish', corge: 'corfish'}
+                ]};
+                $controller('VocabularyEdit', {$scope: scope});
+            }));
+
+            it('being detected correctly', function() {
+                expect(scope.model).toEqual(
+                    {foo: null, bar: null, spam: null, qux: null, corge: null}
+                );
             });
-        }));
+        });
+
+        describe('controller', function() {
+            var scope;
+            var testItem;
+
+            beforeEach(inject(function($rootScope, $controller) {
+                scope = $rootScope.$new();
+                testItem = {foo: 'flareon', bar: 'beedrill'};
+                scope.vocabulary = {items: [testItem]};
+                $controller('VocabularyEdit', {$scope: scope});
+            }));
+
+            it('can add items', function() {
+                scope.addItem();
+                expect(scope.vocabulary.items.length).toBe(2);
+                expect(scope.vocabulary.items[1]).toEqual({foo: null, bar: null});
+            });
+
+            it('can remove items', function() {
+                scope.addItem();
+                scope.removeItem(testItem);
+                expect(scope.vocabulary.items.length).toBe(1);
+                expect(scope.vocabulary.items[0]).toEqual({foo: null, bar: null});
+                scope.removeItem({foo: null, bar: null});
+                expect(scope.vocabulary.items.length).toBe(0);
+            });
+
+            it('can save vocabulary', inject(function(api, $q, $rootScope) {
+                scope.addItem();
+                scope.vocabulary.items[1].foo = 'feraligatr';
+                scope.vocabulary.items[1].bar = 'bayleef';
+                scope.closeVocabulary = function() {};
+                spyOn(api, 'save').and.returnValue($q.when());
+                spyOn(scope, 'closeVocabulary').and.returnValue($q.when());
+                scope.save();
+                $rootScope.$digest();
+                expect(api.save).toHaveBeenCalledWith('vocabularies', {
+                    items: [
+                        {foo: 'flareon', bar: 'beedrill'},
+                        {foo: 'feraligatr', bar: 'bayleef'}
+                    ]
+                });
+            }));
+        });
     });
 });
--- a/client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
+++ b/client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
@@ -7,11 +7,18 @@ describe('vocabularies', function() {
     beforeEach(module('templates'));
 
     it('can fetch vocabularies', inject(function(api, vocabularies, $q, $rootScope) {
-        spyOn(api, 'query').and.returnValue($q.when());
-        var scope = $rootScope.$new();
-        vocabularies.getVocabularies(scope);
+        var fixture = {foo: 'bar'};
+        spyOn(api, 'query').and.returnValue($q.when(fixture));
+        var result;
+        vocabularies.getVocabularies().then(
+            function(vocabs) {
+                result = vocabs;
+            }
+        );
         $rootScope.$digest();
         expect(api.query).toHaveBeenCalledWith('vocabularies');
+        expect(result).toBe(fixture);
+        expect(vocabularies.vocabularies).toBe(fixture);
     }));
 
     describe('config modal', function() {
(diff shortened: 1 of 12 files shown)


client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,152 @@
...
+ * AUTHORS and LICENSE files distributed with this source code, or
+ * at https://www.sourcefabric.org/superdesk/license
+ */
+
+(function() {
+  'use strict';
+
+  VocabularyService
+      .$inject = [ 'api', 'urls', 'session', '$upload', '$q', 'metadata' ];
+  function VocabularyService(api, urls, session, $upload, $q, metadata) {
```
sivakuna-aap: Unused injections `$upload`, `metadata`

client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,152 @@
...
+    var service = this;
+
+    this.vocabularies = undefined;
+
+    this.getVocabularies = function(scope) {
```
sivakuna-aap: Vocabularies are already available in `metadata.values`. We can reuse.

actionless:
@sivakuna-aap see https://dev.sourcefabric.org/browse/SD-694

client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  VocabularyConfigController.$inject =
+      [ '$scope', 'gettext', 'session', 'modal', 'notify', 'vocabularies' ];
+  function VocabularyConfigController($scope, gettext, session, modal, notify,
+                                      vocabularies) {
+    $scope.vocabularies = null;
+    $scope.vocabulary = null;
+
+    $scope.createVocabulary = function() {
```
sivakuna-aap: Since `metadata.values` is cached we either need to update the `metadata.values` or invalidate the `metadata.values`.

client/app/scripts/superdesk-vocabularies/vocabularies.js
actionless: they regrouped into a different structure there

client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,152 @@
...
+ * This file is part of Superdesk.
+ *
+ * Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+ *
+ * For the full copyright and license information, please see the
+ * AUTHORS and LICENSE files distributed with this source code, or
+ * at https://www.sourcefabric.org/superdesk/license
+ */
+
+(function() {
```
sivakuna-aap: Doc Strings would have been helpful

sivakuna-aap:
@actionless after this ticket we had a discussion and finalized that vocabularies will not have UI. Not sure if this is been changed. 

@akintolga/@ioanpocol  any idea if you had discussions with Lasky or Vince about this and added this ticket to the current sprint?

client/app/scripts/superdesk-vocabularies/vocabularies.js
sivakuna-aap: Even though they are regrouped I think it's still reusable.

actionless: it's not very handy to update them after, ie they should regrouped back then

actionless:
@sivakuna-aap 
mb for keywords we can do separate controller and directive and just put them into settings.html here?

sivakuna-aap:
@actionless I had a discussion with @Andrew Lasky and @akintolga. The keywords management will fall under the same UI you put in this PR. I will close my PR and rework after this is merged.

However, this PR needs to address below things:
1. When a vocabulary is updated then respective `metadata.values` should also be updated
2. Deleting an item in a vocabulary shouldn't delete the item from `vocabulary.items`. Instead it should change the value of `is_active` to false.

client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
```diff
@@ -0,0 +1,83 @@
...
+            });
+
+            it('can remove items', function() {
+                scope.addItem();
+                scope.removeItem(testItem);
+                expect(scope.vocabulary.items.length).toBe(1);
+                expect(scope.vocabulary.items[0]).toEqual({foo: null, bar: null});
+                scope.removeItem({foo: null, bar: null});
+                expect(scope.vocabulary.items.length).toBe(0);
+            });
```
plamut: It would be better that the test for removing an item does not rely on the correct behavior of the `scope.addItem()` method - a bug in the latter could cause the former to fail as well, even though there would be nothing worng with the `scope.removeItem()` method.

A recommended approach is to directly create the desired fixture (i.e. setting the `scope.vocabulary.items` list), and run the method under tests (and nothing else) and observe the outcome.

client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
```diff
@@ -0,0 +1,83 @@
...
+                spyOn(api, 'save').and.returnValue($q.when());
+                spyOn(scope, 'closeVocabulary').and.returnValue($q.when());
+                scope.save();
+                $rootScope.$digest();
+                expect(api.save).toHaveBeenCalledWith('vocabularies', {
+                    items: [
+                        {foo: 'flareon', bar: 'beedrill'},
+                        {foo: 'feraligatr', bar: 'bayleef'}
+                    ]
+                });
```
plamut: Similar recommendation to above, there is a dependency on the behavior of the `scope.addItem()` method.

client/app/scripts/superdesk-vocabularies/views/vocabulary-config-modal.html
```diff
@@ -0,0 +1,36 @@
...
+
+        <fieldset>
+            <div ng-repeat="(key, value) in model" class="field vocabularyField">
+              <label translate>{{key}}</label>
+            </div>
+          <div ng-repeat="item in vocabulary.items track by $index">
+            <div ng-repeat="(key, value) in model" class="field vocabularyField">
+              <input type="text" ng-model="item[key]">
+            </div>
+            <button id="del-btn" class="btn del" ng-click="removeItem(item)" translate>x</button>
```
plamut: This probably doesn't need to be translated?

client/app/scripts/superdesk-vocabularies/views/vocabulary-config.html
```diff
@@ -0,0 +1,15 @@
+<div sd-shadow>
+  <ul class="pills-list">
+
+    <li ng-repeat="vocabulary in vocabularies._items">
```
plamut: Can the list of vocabularies change? If yes, it is recommended to add `track by` here
(and if not, `vocabularies._items` should probably be one-time bound)

client/app/scripts/superdesk-vocabularies/views/vocabulary-config.html
```diff
@@ -0,0 +1,15 @@
...
+      <h6>{{ vocabulary._id }}</h6>&nbsp;
```
plamut: Probably one-time binding here? The `_id` does not change?

client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,138 @@
...
+            if (typeof service.vocabularies === 'undefined') {
+                return api.query('vocabularies')
+                .then(function(result) {
+                    service.vocabularies = result;
+                    return $q.when(service.vocabularies);
+                });
+            } else {
+                return $q.when(service.vocabularies);
+            }
+        };
```
plamut: I think the tests for this method are missing?

plamut:
The code looks good IMO, it just needs a few extra tests, docstrings (as @sivakuna-aap said) and perhaps some minor changes in the templates.

client/app/scripts/superdesk-vocabularies/vocabularies.js
actionless: https://github.com/superdesk/superdesk/pull/1159/files#diff-5cfce4615a7541ba392d6011f12c29a1R9

client/app/scripts/superdesk-vocabularies/views/vocabulary-config.html
actionless: atm it can't

client/app/scripts/superdesk-vocabularies/views/vocabulary-config-modal.html
actionless: in japan it will be `o`/`0`

client/app/scripts/superdesk-vocabularies/vocabularies.js
plamut: Sorry, my bad, should have worded it more clearly - _some_ tests are missing. The linked test case tests that the method makes a correct API call, but there should also be other test cases, testing e.g. whether the returned promise is resolved with a correct value and that the list of vocabularies is stored in the service itself.

BTW, why is a scope object passed to the `getVocabularies()` method in the linked test, the method does not have any parameters?

client/app/scripts/superdesk-vocabularies/views/vocabulary-config-modal.html
plamut: Really?? They use a different symbol for closing something? Almost hard to believe that. :)

client/app/scripts/superdesk-vocabularies/tests/vocabularies.spec.js
```diff
@@ -0,0 +1,83 @@
...
+    'use strict';
+
+    beforeEach(module('superdesk.vocabularies'));
+    beforeEach(module('superdesk.authoring'));
+    beforeEach(module('templates'));
+
+    it('can fetch vocabularies', inject(function(api, vocabularies, $q, $rootScope) {
+        spyOn(api, 'query').and.returnValue($q.when());
+        var scope = $rootScope.$new();
+        vocabularies.getVocabularies(scope);
```
petrjasek: the scope param is not used

client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,138 @@
...
+        .$inject = [ 'api', 'urls', 'session', '$q'];
+    function VocabularyService(api, urls, session, $q) {
+        var service = this;
+
+        this.getVocabularies = function() {
+            if (typeof service.vocabularies === 'undefined') {
+                return api.query('vocabularies')
+                .then(function(result) {
+                    service.vocabularies = result;
+                    return $q.when(service.vocabularies);
```
petrjasek: there is no need to call `$q.when`, it's a promise anyhow

client/tasks/options/watch.js
```diff
@@ -17,6 +17,13 @@ module.exports = {
             '<%= appDir %>/scripts/superdesk*/**/*.js'
         ]
     },
+    ngtemplates: {
```
petrjasek: is there a need for that? it's watched already in assets config

client/app/scripts/superdesk-vocabularies/views/vocabulary-config-modal.html
actionless: it was a joke*, i did not changed it during development when label was `delete`

*https://www.reddit.com/r/Games/comments/1f3ron/why_is_o_yes_an_x_no_in_japanese_playstation/

client/app/scripts/superdesk-vocabularies/vocabularies.js
actionless: i did not updated the test after refactoring of that method, will commit it together with other fixes

client/tasks/options/watch.js
actionless: was not working for me unless i added this

client/app/scripts/superdesk-vocabularies/views/vocabulary-config.html
actionless: oops, i misplaced it with length

actionless:
@sivakuna-aap @plamut  @petrjasek resolved comments

client/app/scripts/superdesk-vocabularies/vocabularies.js
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        /**
+         * Save current edit modal contents on backend.
+         */
+        $scope.save = function() {
+            $scope._errorUniqueness = false;
+            api.save('vocabularies', $scope.vocabulary).then(onSuccess, onError);
+            // discard metadata cache:
+            metadata.initialize();
```
sivakuna-aap: This won't force the metadata to reinitialize. https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-authoring/metadata/metadata.js#L507

mdhaman:
@actionless it is good to have vocabulary management. 
imo, some of the vocabularies don't require management. For example, `type, signal, subscriber_type`. 

There are few question that needs to be discussed before we introduce vocab management. What should happen to the system if we mark the `type=text` as inactive or delete it? What should happen if I add a new subscriber_type or remove existing subscriber_type?

imo, `categories, default_categories, genre, rightsinfo, locators` will require management but not sure about others.

ioanpocol:
@actionless  Do you need any help on this PR?

actionless:
what else need to be done here?

amagdas:
We need to make the tests pass before merging.

actionless:
rebased on current master

plamut:
Looking at the failed test I think these code changes have not caused them (at least not all of them), it is just our e2e tests failing randomly (again). One or two protractor tests that failed in my PR (basically just a minor CSS change) were the same that failed here, too, and every time a build was restarted, a different test (or a combination of them) failed.

Restarting build jobs enough times will IMO make the build pass eventually, though this is of course not a definite solution of any kind...

ioanpocol:
@actionless 
Please rebase again, seams that some tests are old and don't contains the changes for desk management.

actionless:
rebased

ioanpocol:
The issue is still there. I will fix it.

actionless:
thanks, i have no idea how can it be related to these changes

ioanpocol:
@actionless 
Please disable "monitoring view configure a stage and then delete the stage" test until the issue SD-3371 will be fixed.

actionless:
now it's passing

ioanpocol:
cool, so we can merge it

amagdas:
:+1:

PR #1162, opened 2015-09-21
[SD-3236] - Scheduled items stay in "scheduled" state if they don't get queued

akintolga:
This issue happens when:
- There is no queue entries for an scheduled item so a transmission never takes place so the item always stays in "scheduled" state.
- A text story is scheduled but wire clients didn't get the text but digital clients did get the take pckg. So text version of the story always stays in "scheduled" state.

This is a clean-up task to update the scheduled stories to "Published" once the "publish_schedule" field is overdue and the story is still in scheduled state. Even if there are problems in transmitting the story the state needs to change to "Published" as per users publish queue state doesn't matter much.

### DIFF
--- a/server/apps/archive/commands.py
+++ b/server/apps/archive/commands.py
@@ -8,14 +8,18 @@
 # AUTHORS and LICENSE files distributed with this source code, or
 # at https://www.sourcefabric.org/superdesk/license
 
-
-from eve.utils import ParsedRequest, date_to_str
 import superdesk
+import logging
+from eve.utils import ParsedRequest, date_to_str, config
+from superdesk.celery_task_utils import is_task_running, mark_task_as_not_running
 from superdesk.utc import utcnow
 from .archive import SOURCE as ARCHIVE
-import logging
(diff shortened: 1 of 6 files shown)


server/apps/archive/commands.py
```diff
@@ -50,11 +61,68 @@ def _get_query_for_expired_items(self, now):
...
+    """
+
+    logger.info('Get overdue scheduled content from {}'.format(resource))
+    query = {'$and': [
+        {'publish_schedule': {'$lte': expired_date_time}},
+        {ITEM_STATE: CONTENT_STATE.SCHEDULED}
+    ]}
+
+    req = ParsedRequest()
+    req.sort = '_created'
```
sivakuna-aap: Sort should be on `_modified` aka the item which is scheduled first instead of `_created`.

sivakuna-aap:
Overall it looks good. In some places we can use constants for `archive`, `state`

mdhaman:
:+1:  looks good except for @sivakuna-aap comments.

PR #1164, opened 2015-09-21
feat(packages): Implemented combining an item with currently open ite…

eozan:
…m into a package.

SD-3124


petrjasek:
:+1: but put those commits together pls

amagdas:
:+1:

PR #1167, opened 2015-09-21
fix(lock): fix unlocking of auto generated items

petrjasek:
when item is saved it gets populated, but before it's not set
and unlocking it fails with KeyError. there is also some bc fix
for items using objectId and not uuid

### DIFF
--- a/server/apps/common/models/io/base_proxy.py
+++ b/server/apps/common/models/io/base_proxy.py
@@ -49,11 +49,10 @@ class BaseProxy(DataLayer):
         return self.data_layer.delete(resource, filter)
 
     def _update(self, resource, filter, doc, method='update'):
-        _id = doc.get(ID_FIELD, None)
-        if ID_FIELD in doc:
-            del doc[ID_FIELD]
+        _id = doc.pop(ID_FIELD, None)
         original = self.find_one(resource, filter, None)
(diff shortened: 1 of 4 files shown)


ioanpocol:
:+1:

amagdas:
:+1:

server/apps/common/models/io/base_proxy.py
```diff
@@ -49,11 +49,11 @@ def delete(self, resource, filter):
...
-        _id = doc.get(ID_FIELD, None)
-        if ID_FIELD in doc:
-            del doc[ID_FIELD]
+        _id = doc.pop(ID_FIELD, None)
         original = self.find_one(resource, filter, None)
-        res = getattr(self.data_layer, method)(resource, filter[ID_FIELD], doc, original)
-        if _id is not None:
-            doc[ID_FIELD] = _id
+        if filter[ID_FIELD] != original[ID_FIELD]:
+            filter[ID_FIELD] = original[ID_FIELD]  # make sure it's correct type
```
plamut: The `if`condition seems kind of redundant here.

plamut:
:+1:

server/apps/common/models/io/base_proxy.py
petrjasek: ah right

PR #1170, opened 2015-09-21
fix(packaging): fix open item link in package

petrjasek:
in case you have the lock for an item you can not open it readonly.
so in such case it will now try to open it in edit mode.

SD-3167


amagdas:
:+1:

ioanpocol:
:+1:

PR #1176, opened 2015-09-22
fix(desk dropdown): Added Untitled for workspace where name is not specified

pavlovicnemanja92:
There was some cases when workspace name is empty. This is a fix for it.

desk drodpown


Aca-jov:
:+1:

sivakuna-aap:
Isn't Workspace name mandatory? If not I would suggest it to make mandatory.

pavlovicnemanja92:
Yeah, it is mandatory... But there are some ways where workspace is created without name. I couldn't reproduce this, and because of that this is "just in case" fix :)

mugurrus:
:+1:

amagdas:
:+1:

PR #1182, opened 2015-09-23
[SD-3163] Dont show the publish button if the item state is draft

marwoodandrew:
Attempts to publish anything in a draft state fails the state transition checks, this should avoid that.


sivakuna-aap:
Can we drive it via authoring.itemActions() similar to other buttons?
https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-authoring/authoring.js#L501

marwoodandrew:
Other than I tried that and it failed, I look at it again!

sivakuna-aap:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

PR #1183, opened 2015-09-23
fix(chrome) - Default chrome build that works for e2e test

akintolga:


sivakuna-aap:
:+1: :100: :purple_heart: :yellow_heart: :heart:

plamut:
:+1: 


@superdesk/superdesk-developers:
I would also consider changing the current data provider (in another PR), since this is the second time it failed us in the last few weeks. While this fix will serve us well for quite some time, but a hard-coded value might prove problematic in the long run, because it will be reverting to a Chrome version that will be quite old at that time.

Probably something to discuss in one of the meetings...

PR #1184, opened 2015-09-23
Fixed Issues

sivakuna-aap:
1. [SD-3313](https://dev.sourcefabric.org/browse/SD-3313) Disable "send and continue" for embargo items. Also, modified send.fetchStage() to select the stage of the item instead of selecting Incoming Stage always.
2. [SD-3311](https://dev.sourcefabric.org/browse/SD-3311) Update function is not working
3. Fixed the button Label is being displayed as "Save and Edit" when user tries to publish an item without saving changes.


akintolga:
:+1:

mugurrus:
:+1:

PR #1185, opened 2015-09-23
[SD-3262] - Unlocking an item throws the other user out of monitoring

akintolga:
Navigation is not needed anymore...


sivakuna-aap:
:+1: Can we check if there are traces of `$location.url($scope.referrerUrl)` and if they are relevant?

akintolga:
These are the only one I could see...

mugurrus:
:+1:

PR #1191, opened 2015-09-24
feat(monitoring): Implemented "add to current package" multi action.

eozan:
SD-3308

### DIFF
--- a/client/app/scripts/superdesk-search/search.js
+++ b/client/app/scripts/superdesk-search/search.js
@@ -1377,8 +1377,8 @@
             };
         })
 
-        .directive('sdMultiActionBar', ['asset', 'multi',
-        function(asset, multi) {
+        .directive('sdMultiActionBar', ['asset', 'multi', 'authoringWorkspace',
+        function(asset, multi, authoringWorkspace) {
             return {
                 controller: 'MultiActionBar',
                 controllerAs: 'action',
@@ -1387,6 +1387,11 @@
                 link: function(scope) {
                     scope.multi = multi;
                     scope.$watch(multi.getItems, detectType);
+
+                    scope.isOpenItemType = function(type) {
+                        var openItem = authoringWorkspace.getItem();
+                        return openItem.type === type;
+                    };
 
                     /**
                      * Detects type of all selected items and assign it to scope,
@@ -1445,6 +1450,10 @@
                     notify.error(gettext(response.data._message), 3000);
                 }
             });
+        };
+
+        this.addToPackage = function() {
+            $rootScope.$broadcast('package:addItems', {items: multi.getItems(), group: 'main'});
         };
 
         /**
--- a/client/app/scripts/superdesk-search/views/multi-action-bar.html
+++ b/client/app/scripts/superdesk-search/views/multi-action-bar.html
@@ -9,6 +9,9 @@
         <button class="navbtn strict" ng-click="action.createPackage()" title="{{ :: 'Create Package' | translate }}">
             <i class="big-icon-package-create"></i>
         </button>
+        <button class="navbtn strict" ng-if="isOpenItemType('composite')" ng-click="action.addToPackage()" title="{{ :: 'Add to Current Package' | translate }}">
+            <i class="big-icon-package-add"></i>
+        </button>
     </div>
     
     <div class="dropdown strict highlights-dropdown pull-right" ng-if="type === 'archive'" title="{{ :: 'Mark item' | translate }}" dropdown>


plamut:
Tests and docstrings should be added for the modified methods, esp. tests for broadcasting the `package:addItems` event are important IMO.

The rest looks good to me.

mugurrus:
:+1:

ioanpocol:
I will update the e2e test for this task after this PR will be merged.
Usually a method with 1-3 lines of code don't need any documentation if if well written. If still needs usually something is wrong. Of course exceptions can exists, but are very low as percentage.

:+1:

plamut:
I tend to disagree here, since it has actually happened a couple of times that there was a bug in a simple method that caused quirky behavior in some seemingly unrelated feature in certain conditions. Locating and fixing an error took a couple of hours, while a corresponding test would take 10 minutes (give or take) to write.
It is also hard to draw a line when a method is "complex enough" to justify testing, making it more likely that we might skip tests even at places where they would actually be needed. On the other hand, accompanying code with tests by default avoids such risks.

With that said, I would not recommend skipping tests, unless if there is a very good reason to do so. Deviate from that on your own risk. ;)

ioanpocol:
@plamut I think you misunderstood my comment.

I am refactoring e2e tests re: packages and because of this I proposed to change the current test. So I think that your feedback is not for this case.

Also my point is that you don't have to document all methods. If a comment don't add any info it is useless. As I said, a very short method should self-document, if not, usually something is wrong.
Just to add a comment in order to check that the function has a comment is not a help for anyone.

plamut:
@ioanpocol Ah, OK, thanks.

Re: comments - I think they are useful even for shorter methods, because it is easier for the brain to parse a sentence in a natural language than having to parse the expressions in code. Forcing a reader to mentally parse the source code represents a rather unnecessary additional cognitive burden.

Maybe others that have been working on the project for much longer and work in the same room with others see this differently, but for those of us who can't simply turn to somebody sitting at the nearby desk, comments are invaluable and at times even the only source of information (considering the lack of project documentation). Judging from my personal experience, the code is not always as clear as it could ideally be.

I imagine that potential external contributors and newcomers would feel the same, and that is why I actively encourage writing comments and provide them myself as well.

Just my 2 ¢

ioanpocol:
@plamut 
My comment was about short methods well written that usually document them-self. Adding a comment to this kind of methods don't add any additional info. And by the way we are coders, so we read code every day :).
So again second part of your feedback don't apply here, I agree that more complex methods should be documented. And as a simple rule, if you have doubts that a method is complex or not, it is already complex.

plamut:
@ioanpocol generally the only "problem" is that there is no clear way of determining where to draw the line. :) What seems "simple enough" for somebody, might not be that simple for someone else. That is, it might still be _understandable_, but not without some extra effort, which can become quite a burden when digging deeply through the layers of code.

A missing docstring where needed is more harmful than a docstring that is at worst redundant for somebody, and that's why having a docstring should be a default IMO. If nothing else, it at least makes a good example for new developers, since they often just copy the patterns they see in the existing code. ;)

ioanpocol:
@plamut 
If a docstring just repeat the name of the method it is not useful at all.
This is my point. I hope you understood.

plamut:
@ioanpocol 
I understood that, yes, I call that kind of a method _"trivial"_. A _"simple"_ method (at least in the way I understand it) is a bit more complex than that, usually because it contains a non-trivial expression and/or accepts arguments.

These were the kinds of methods I was referring to in my replies, and my point was that for _"simple"_ methods a docstring already provides some value, especially when having parameters.

If missing docstrings for _"trivial"_ (and _"simple"_?) methods are acceptable, I suggest making this an "official" guidance to avoid unnecessary repeated discussions on this topic. To my best knowledge, we have agreed that a minimal viable PR should, among other things, include docstrings for all methods, but if this is no longer the case, I was not informed about it.

Sorry for the noise, if this is the case.

ioanpocol:
@plamut 
Sorry for my replies, indeed this is not the place to have this discussion.
I will open this discussion on next retrospective meeting.

plamut:
@ioanpocol 
No, no, please don't get me wrong - if you ask me, we should have _more_ discussions like this, and I really liked hearing your perspective. :) It's an excellent way to exchange knowledge and ideas between different dislocated groups, and if we identify a topic/problem worth discussing in a meeting, that can only be positive.

I agree to continue the discussion in a meeting, it has grown too big to be handled here.

ioanpocol:
@plamut 
I think that you misunderstood my message :). 
I didn't propose to close the discussion, but to continue on the right place.

plamut:
@ioanpocol 
Don't worry, I have understood you perfectly well. ;)

ioanpocol:
@plamut 
From your message seams not. But don't worry, it's ok.
I think we have to close this discussion since is not going in the right direction.

PR #1193, opened 2015-09-25
[SD-3349] (fix) Fixing the duplicate item ng-repeat error for published content.

mdhaman:


sivakuna-aap:
:+1: can we change the value of `_current_version` to `integer` instead of `string`?

akintolga:
:+1:

PR #1196, opened 2015-09-25
Fix (spiked list): On double click should not open spiked item on edit mode

Aca-jov:
SD-3093

### DIFF
--- a/client/app/scripts/superdesk-monitoring/monitoring.js
+++ b/client/app/scripts/superdesk-monitoring/monitoring.js
@@ -285,18 +285,20 @@
                     moveTimeout;
 
                 function edit(item, lock) {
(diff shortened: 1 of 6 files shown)


plamut:
IMO an extra test case should be added to verify that editing only happens for non-spiked items, especially since the `edit()` method is complex enough to justify this.

A docstring welcome as well, even though it is "just" a helper function (the `scheduleQuery()` function has it, for instance).

The commit message itself is descriptive and serves the purpose.

mugurrus:
Yep, needs a test. The edit function is pretty self-explanatory.

amagdas:
:+1:

mugurrus:
:+1:

PR #1197, opened 2015-09-25
(imp) [SD-3033] Dateline value should update with edits to the point of publishing

sivakuna-aap:
Requires [Superdesk-Core's PR#26](https://github.com/superdesk/superdesk-core/pull/26)

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -126,7 +126,7 @@
          * Autosave an item
          */
         this.save = function saveAutosave(item) {
-            if (item._editable && item._locked) {
+            if (item._editable && !item._locked) {
                 this.stop(item);
 
                 timeouts[item._id] = $timeout(function() {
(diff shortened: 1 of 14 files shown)


sivakuna-aap:
ping @superdesk/developers

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -126,7 +126,7 @@
          * Autosave an item
          */
         this.save = function saveAutosave(item) {
-            if (item._editable && item._locked) {
+            if (item._editable && !item._locked) {
```
ioanpocol: Autosave works when an item is edited but when an item it will be locked.
So why is this change required?

sivakuna-aap: That's not always true. There are some watches on `item` which sets field values using javascript. Once the values are set the `autosave.save()` is triggered even though the item isn't opened for edit. So this check prevents autosaving if the item isn't opened for edit and isn't locked by other users.

akintolga:
:+1:

mdhaman:
:+1:

PR #1198, opened 2015-09-28
Make the currently edited routing scheme rule highlighted

plamut:
Resolves [SD-2434](https://dev.sourcefabric.org/browse/SD-2434).

Please see the attached screenshot for an idea how the change looks like. The currently selected filter has a dark gray background.

routing_scheme_edit

P.S.: We keep `client/bower/styles/css/bootstrap.css` under source control, right?

### DIFF
--- a/client/app/scripts/superdesk-ingest/styles/settings.less
+++ b/client/app/scripts/superdesk-ingest/styles/settings.less
@@ -206,6 +206,12 @@
 		max-width: 100%;
 		li {
 			margin: 0 5px 5px 0;
+			&.highlight {
+				background-color: #5e5e5e;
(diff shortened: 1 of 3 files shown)


sivakuna-aap:
:+1: I think deletion of `bootstrap.css` isn't needed

plamut:
@sivakuna-aap It wasn't deleted, just recompiled and replaced with a new version. It seems that we have this file under source control, thus I didn't do any changes to `.gitignore`.

sivakuna-aap:
ok..

mdhaman:
:+1:

amagdas:
:+1:

PR #1201, opened 2015-09-28
Grid view for templates list

pavlovicnemanja:
- Changed markup for template list
- improved styling for action menu
- new icons


client/app/scripts/superdesk-templates/templates.js
```diff
@@ -156,8 +157,37 @@
...
+                    $scope.getTemplateDesk = function getTemplateDesk(template) {
+                        return _.find($scope.desks._items , {_id: template.template_desk});
+                    };
+
+                    /*
+                     * Returns stage name
+                     */
+                    $scope.getTemplateStage = function getTemplateStage(template) {
+                        return _.find(desks.stages._items , {_id: template.template_stage});
+                    };
```
plamut: Is there a specific reason that these two methods are defined inside the promise resolve handler? They can (and probably should) be simply defined outside of it, just like the rest.

client/app/scripts/superdesk-templates/templates.js
```diff
@@ -156,8 +157,37 @@
...
                 });
 
+                /*
+                 * Convert string to time object
+                 *
+                 * @param {String} time
```
plamut: I recommend describing the expected format of the input string, it is probably not apparent that it is `%M%S` (?).

client/app/scripts/superdesk-templates/views/templates.html
```diff
@@ -7,17 +7,43 @@ <h2 translate>Templates</h2>
...
+                          <h4 class="with-value">{{ :: 'Template type' | translate }} <span class="value-label">{{ template.template_type }}</span></h4>
+                      </li>
+                      <li>
+                          <h4>{{ :: 'Desk / Stage' | translate }}</h4>
+                          <span>{{getTemplateDesk(template).name}} / {{getTemplateStage(template).name}}</span>
+                      </li>
+                      <li>
+                          <h4>{{ :: 'Automated item creation' | translate }}</h4>
+                          <span ng-repeat="day in template.schedule.day_of_week">
+                              {{weekdays[day]}}
```
plamut: We can probably use one-time binding here, since we are dealing with fixed values.

plamut:
Looks good, nice and clean!

Consider adding unit tests at least for the `$scope.getTime()` method (it is always easy to get the indexes wrong...) if they don't exist yet, and see if the other minor comments are applicable, but all in all it is a good piece of work IMO!

:+1:

client/app/scripts/superdesk-templates/templates.js
pavlovicnemanja: They are defined inside promise because we need to wait for the desks to be initialize first, because we use that list to find current desk/stage...

plamut: Hey, good morning, you can sleep for a bit more, you know. ;)

Regarding the methods - in the promise success handler we simply define the methods, we don't invoke them. If the intent is to prevent the user from _using_ these methods before the data arrives (by no having the methods available yet), a better pattern IMO is to have some kind of a `hasLoaded` flag and disable the relevant UI elements based on it.

Would that do the trick, what do you think?

pavlovicnemanja: Night is deprecated for me :-) a lot of lost time! :-)

I will take a try, I think that this will do the trick! :-) Thanks! :-)

mugurrus:
:+1:

amagdas:
:+1:

PR #1202, opened 2015-09-28
Fix (multiedit disabled for spiked items)

Aca-jov:
SD-2094

And couple of small fixes that I saw and there is no issue for that:
- On monitoring sorting with searching looks is fixed
- User management looks fixed
- Settings dropdown shown only on monitoring, on other pages not visible (3 dots)
- On monitoring on filtering by filetype checked items are unchecked

### DIFF
--- a/client/app/scripts/superdesk-desks/aggregate.js
+++ b/client/app/scripts/superdesk-desks/aggregate.js
@@ -11,8 +11,8 @@
  (function() {
     'use strict';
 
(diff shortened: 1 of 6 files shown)


plamut:
Do we have a test in place for the fix (e.g. `setFileType()` indeed resets multi)?

Good work otherwise, especially with all the other small issues that you found and fixed along the way.

:+1:

ioanpocol:
:+1: 
Probably we can disable filter by file type when we are on multi mode.

PR #1204, opened 2015-09-29
[sd-3303]- UI: Make clients aware of socket disconnection and reconnect automatically

sjunaid:


client/app/scripts/superdesk/notification/notification.js
```diff
@@ -30,19 +34,67 @@
...
+     * Service for notifying user when websocket connection disconnected or connected.
+     */
+    NotifyConnectionService.$inject = ['$rootScope', 'notify', 'gettext', '$timeout'];
+    function NotifyConnectionService($rootScope, notify, gettext, $timeout) {
+        var successTimeout, alertTimeout;
+        var _this = this;
+        _this.message = null;
+
+        $rootScope.$on('disconnected', function(event) {
+            _this.message = 'websocket connection lost, attempting to reconnect ...';
```
sivakuna-aap: Users might not understand technical terms like websocket...user friendly message would be a better

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -30,19 +34,67 @@
...
+            $timeout.cancel(alertTimeout);
+            alertTimeout = $timeout(function() {
+                notify.error(gettext(_this.message));
+            }, 100);
+        });
+        $rootScope.$on('connected', function(event) {
+            _this.message = 'Connected!!!';
```
sivakuna-aap: Better to expand the message. Something like `Connected to Notification Server`

sivakuna-aap:
:+1: except the minor comments

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -30,19 +34,67 @@
...
+                clearInterval(connectTimer);
+                $rootScope.$broadcast('connected');
+            };
+
+            ws.onclose = function(event) {
+                $rootScope.$broadcast('disconnected');
+                clearInterval(connectTimer);
+                connectTimer = setInterval(function() {
+                    if (ws) {
+                        connect();  // Retry to connect for every TIMEOUT interval.
```
plamut: It it probably better to incrementally increase the timeout period. For example, if a disconnect happens due to a temporarily overloaded server, hammering it with reconnection attempts would do no good, especially if _all_ clients would be trying to do the same.

If reconnection has failed several times in a row already, a client can probably assume that something more serious is going on and the issues are not just momentary, hence less likely to be resolved in the very next attempt. It thus makes sense to wait for a bit longer. When a connection is finally restored, the timeout period can be again reset to the initial short value.

What do you guys think, should we add this smarter reconnecting logic?

client/app/scripts/superdesk/notification/tests/notification_spec.js
```diff
@@ -68,3 +68,27 @@ describe('Reload Service', function() {
...
+        msg = 'websocket connection lost, attempting to reconnect ...';
+        rootScope.$broadcast('disconnected');
+        expect(notifyConnectionService.message).toEqual(msg);
+    }));
+
+    it('can show success message when websocket connected', inject(function(notifyConnectionService) {
+        msg = 'Connected!!!';
+        rootScope.$broadcast('connected');
+        expect(notifyConnectionService.message).toEqual(msg);
+    }));
```
plamut: It's good to have the event handlers tested!

client/app/scripts/superdesk/notification/notification.js
sivakuna-aap: I agree with @plamut

plamut:
We might want to add a few unit tests for the `ws` event handlers, but other than that this feature is great. I would recommend adding a smarter reconnection logic, too, but it's not necessary to be done in the same PR.

:+1:

client/app/scripts/superdesk/notification/tests/notification_spec.js
sjunaid: It turned out that WebSocketProxy is not a service but function run in the `run()` method, making it difficult to obtain a reference to it and to test it. Do you think that only testing the broadcasted events is good enough?

plamut: As discussed in the office, yes, it is fine. Creating proper unit tests in this would require too many workarounds and code changes to justify the time investment.

client/app/scripts/superdesk/notification/notification.js
sjunaid: Nice! and agreed, that can be improved in separate PR.

ioanpocol:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -30,19 +34,67 @@
...
+
+            ws.onopen = function(event) {
+                clearInterval(connectTimer);
+                $rootScope.$broadcast('connected');
+            };
+
+            ws.onclose = function(event) {
+                $rootScope.$broadcast('disconnected');
+                clearInterval(connectTimer);
+                connectTimer = setInterval(function() {
```
petrjasek: just wondering - is there any reason to use `setInterval` and not `$interval` service from angular?

plamut: Could be a typo, or an intentional way of avoiding firing unnecessary $digest cycles, since all the event are propagated by `$broadcasting` them. But it is also true that this can be achieved with the `$interval` service as well, it's just a matter of passing the right parameter to it.

Without a specific reason for using `setInterval()`, it is probably slightly better to use `$interval` IMO, because it would represent a better (default) coding pattern example for future contributors. :)

sjunaid: Ok, will do with $interval, consider this in progress.

client/app/scripts/superdesk/notification/notification.js
```diff
@@ -30,19 +34,67 @@
...
+            };
+
+            ws.onclose = function(event) {
+                $rootScope.$broadcast('disconnected');
+                $interval.cancel(connectTimer);
+                connectTimer = $interval(function() {
+                    if (ws) {
+                        connect();  // Retry to connect for every TIMEOUT interval.
+                    }
+                }, TIMEOUT);
```
plamut: By default, `$interval` invokes the given function within an `$apply()`, triggering a new digest cycle. The latter is not needed in our case, thus we can prevent it from happening by passing the corresponding parameter to `$interval()`.

plamut:
:+1:

amagdas:
:+1:

PR #1195, opened 2015-09-25
Fix (multiedit): On multiedit close the open items should not be closed

Aca-jov:
SD-3270: Related to @fritzSF comment

PR #1205, opened 2015-09-29
Fix: unable to deselect a macro in routing scheme rule w/o reloading th page

plamut:
Resolves [SD-2348](https://dev.sourcefabric.org/browse/SD-2348).

This fix simply adds a blank option to all relevant dropdown lists. On top of that, it uses `ng-options` instead of `ng-repeat` to create their options, improving the speed and reducing the memory usage (because `ng-repeat` creates a new scope for every element in the collection, but `ng-options` does not).

### DIFF
--- a/client/app/scripts/superdesk-desks/views/actionpicker.html
+++ b/client/app/scripts/superdesk-desks/views/actionpicker.html
@@ -1,20 +1,33 @@
 <div class="multiple">
-	<div class="field">
-		<label for="desk">Desk</label>
-		<select id="desk" ng-model="desk" name="desk">
-		    <option ng-repeat="d in desks" value="{{d._id}}" ng-selected="d._id === desk">{{d.name}}</option>
-		</select>
-	</div>
-	<div class="field">
-		<label for="stage">Stage</label>
-		<select id="stage" ng-model="stage" name="stage">
-		    <option ng-repeat="s in deskStages[desk]" value="{{s._id}}" ng-selected="s._id === stage">{{s.name}}</option>
-		</select>
-	</div>
-	<div class="field">
-		<label for="macro">Macro</label>
-		<select id="macro" ng-model="macro" name="macro">
-		    <option ng-repeat="m in deskMacros" value="{{m.name}}" ng-selected="m.name === macro">{{m.label}}</option>
-		</select>
-	</div>
+
+    <div class="field">
+        <label for="desk" translate>Desk</label>
+        <select
+            id="desk" name="desk"
+            ng-options="d._id as d.name for d in desks"
+            ng-model="desk">
+            <option value=""></option>
+        </select>
+    </div>
+
+    <div class="field">
+        <label for="stage" translate>Stage</label>
+        <select
+            id="stage" name="stage"
+            ng-options="s._id as s.name for s in deskStages[desk]"
+            ng-model="stage">
+            <option value=""></option>
+        </select>
+    </div>
+
+    <div class="field">
+        <label for="macro" translate>Macro</label>
+        <select
+            id="macro" name="macro"
+            ng-options="m.name as m.label for m in deskMacros"
+            ng-model="macro">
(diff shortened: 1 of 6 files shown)


mugurrus:
:+1: (test?)

plamut:
I don't really know what to test here, it's effectively just a small template change, and all the existing tests for this feature still pass? Or am I missing something?

mdhaman:
:+1:

plamut:
@mugurrus Please respond? Is it fine to merge this in your opinion or not?

mugurrus:
An e2e test would be needed to verify the possibility of deselecting a macro (the fix you just made). This is what's called a regression test.

client/app/scripts/superdesk-desks/views/actionpicker.html
```diff
@@ -1,20 +1,33 @@
...
-	</div>
-	<div class="field">
-		<label for="macro">Macro</label>
-		<select id="macro" ng-model="macro" name="macro">
-		    <option ng-repeat="m in deskMacros" value="{{m.name}}" ng-selected="m.name === macro">{{m.label}}</option>
-		</select>
-	</div>
+
+    <div class="field">
+        <label for="desk">Desk</label>
```
petrjasek: all labels can use `translate`

plamut:
@mugurrus My question was more pointed at how detailed our tests should be, i.e. what things do we even test in general? (because some people say that _"everything"_ is not a correct answer :smile: )

For example, why was [PR #1201](https://github.com/superdesk/superdesk/pull/1201) considered fine to merge without having any tests? And the same for the recent minor fixes in [#1156](https://github.com/superdesk/superdesk/pull/1165), [#1176](https://github.com/superdesk/superdesk/pull/1176) and [#1195](https://github.com/superdesk/superdesk/pull/1195/files)?

There is quite a noticeable inconsistency between reviews of different PRs...

client/app/scripts/superdesk-desks/views/actionpicker.html
plamut: I can refactor this, too, along the way, good catch!

mugurrus:
If you want to have a conversation about testing then we should have a developers meeting. Writing comments does not make a very efficient conversation.

Re PR #1201 I made a mistake, I should have asked for an e2e test, while PR #1176 was trivial. I'm not responsible for PRs I didn't review.

In this case I consider an e2e test needed.

plamut:
This PR effectively only adds a single static blank `<option value=""></option>` tag to each of the `<select>` elements without any additional logic, why is it then considered any less trivial than #1176, for example?

Please don't get me wrong, I'm not opposing the test _per se_, I'm just curious to hear the rationale behind it, preferably something determinable. Thanks!

ioanpocol:
@plamut @mugurrus  
Please don't start here another very long discussion. As Mugur proposed please meet together and have a discussion about this.

plamut:
Sorry, yes, you are right. Just please tell me whether this (trivial IMO) fix justifies the e2e test or not? Thanks!

ioanpocol:
@plamut 
If there is already an e2e test re: this functionality there will be trivial to modify it and add a check for this situation. 
If there are no e2e test for this functionality, this can be a good moment to start writing at least one test.

You are a champion in telling other people to write tests so you should give to all of us a good example and write tests yourself.

plamut:
@ioanpocol OK thanks, agreed then, I'll add it (like 99% of the time, if I may add).

I would just appreciate to see more consistency in general, i.e. a better clarity when tests should be written and when not, since it is becoming apparent that the prevailing consensus on the project is _"not everything should be tested"_, despite a different conclusion from one of the tech meetings.
I hope this is not too much to ask for.

plamut:
PR updated with an e2e test, also added some utility functions and `ElementFinder` objects for testing the `Ingest Settings` page.

amagdas:
:+1:

PR #1206, opened 2015-09-29
[SD-3281] - Cleanup of the media files on expiry

akintolga:


server/apps/archive/common.py
```diff
@@ -234,6 +234,44 @@ def remove_unwanted(doc):
...
 
 
+def remove_media_files(doc):
+        """
+        Removes the media files of the given doc if they are not references by any other
+        story across all repos. Returns true if the medis files are removed.
+        """
+        print('Removing Media Files...')
+
+        if doc.get(ITEM_TYPE) in [CONTENT_TYPE.PICTURE, CONTENT_TYPE.VIDEO, CONTENT_TYPE.AUDIO]:
```
plamut: Should we reverse the condition here and `return false` early? That way we won't have to indent the main block, some of its lines are longer and already pushed quite far to the right...

The `list` itself can be replaced with a `set` (slightly better), but it's negligible, would mainly serve just as an example of a good practice for newcomers who might copy the pattern to other parts of the code.

plamut:
Looks good. :+1:

amagdas:
:+1:

PR #1208, opened 2015-09-29
upgrading eve to release 0.6 instead of commit and modifying the commit of superdesk-core

sivakuna-aap:


amagdas:
I think that we should wait to release superdesk-core library so we won't have any git deps.

sivakuna-aap:
Agreed. But whats the timeline for the first release of superdesk-core?

plamut:
I think there still a few things that have not been moved to core yet, but the latter depends on them, e.g. a few services, the base TestCase class that has further non-core dependencies, etc.

It's probably not something that will be finished that soon, but it might be in the next sprint if we allocate time for it.

sivakuna-aap:
I totally agree and needs to be addressed in a different. Until then the version of each library/framework should be consistent/same across repos.

plamut:
If I'm not missing anything, fixating the Eve version makes sense to do even now IMO.
:+1:

PR #1209, opened 2015-09-29
feat(authoring): Improved header and slide panel

pavlovicnemanja:
- SD-3168 Added tab index for input fields in authoring screen and improved UI for header in authoring
- SD-3000 Added focus for `slugline` field on opening
- SD-3333 If publish option in slide pane is not allowed, rename it to 'Send To'
- SD-3240 Added default news value (urgency) icon
- SD-3344 , SD-3341 Fixed layout issue for slide pane

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
@@ -1,14 +1,28 @@
-<div class="dropdown" dropdown>
-	<div class="dropdown-toggle" dropdown-toggle ng-if="field !== 'place'">
+<div class="dropdown" ng-class="{'dropright': icon, 'dropdown-noarrow': !icon}" dropdown sd-dropdown-position>
+    <button class="dropdown-toggle line-input" dropdown-toggle ng-if="field !== 'place' && !icon">
         {{ item[field].name || item[field] ? item[field].name || item[field] : "none" | translate }}
         <b class="caret"></b>
-    </div>
-	<div class="dropdown-toggle" dropdown-toggle ng-if="field === 'place'">
+    </button>
+    <button class="dropdown-toggle line-input" dropdown-toggle ng-if="field === 'place' && !icon">
         {{ item[field].length > 0 ? item[field][0].name : "none" | translate }}
         <b class="caret"></b>
-    </div>
-	<ul class="dropdown-menu pull-left">
-		<li ng-click="select()" translate><button ng-disabled="disabled">None</button></li>
-		<li ng-repeat="o in list" ng-click="select(o)"><button ng-disabled="disabled">{{::o.name}}</button></li>
-	</ul>
+    </button>
+    <button class="dropdown-toggle line-input" dropdown-toggle ng-if="icon">
+        <i class="{{icon}}-{{item[field]}}">
+            <span ng-if="field === 'urgency'">{{item[field] ? item[field] : 0}}</span>
+        </i>
+         <b class="caret"></b>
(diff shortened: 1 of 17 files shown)


client/app/scripts/superdesk-authoring/metadata/views/metadata-dropdown.html
```diff
@@ -1,14 +1,28 @@
...
+    <button class="dropdown-toggle line-input" dropdown-toggle ng-if="icon">
+        <i class="{{icon}}-{{item[field]}}">
+            <span ng-if="field === 'urgency'">{{item[field] ? item[field] : 0}}</span>
+        </i>
+         <b class="caret"></b>
+    </button>
+    <ul class="dropdown-menu">
+        <li ng-if="label"><div class="menu-label">{{label}}</div></li>
+        <li ng-click="select()" translate><button ng-disabled="disabled">None</button></li>
+        <li ng-repeat="o in list" ng-click="select(o)">
```
plamut: If `o.name` is unique, we can use it for `track by`, otherwise it's fine as it is.
If possible, I would also consider choosing a more expressive variable name, though, something that explains it better what kind of an item `o` is.

client/app/scripts/superdesk-authoring/views/article-edit.html
```diff
@@ -1,114 +1,121 @@
...
+        </div>
+
+        <div class="dateline-date">
+            <select id="datelineMonth" ng-disabled="!item.dateline.located || !_editable" ng-change="resetNumberOfDays(true)" ng-model="datelineMonth">
+                <option value=""></option>
+                <option value="{{monthValue}}" ng-repeat="(monthName, monthValue) in monthNames track by monthName">{{monthName}}</option>
+            </select>
+            <select id="datelineDay" ng-disabled="!item.dateline.located || !_editable" ng-change="modifyDatelineDate()" ng-model="datelineDay">
+                <option value=""></option>
+                <option value="{{day}}" ng-selected="datelineDay === day" ng-repeat="day in daysInMonth track by day">{{day}}</option>
```
plamut: For both dropdown menus it is better to replace `ng-repeat` with `ng-options`. The latter is [more efficient](https://docs.angularjs.org/api/ng/directive/select) as explained in the linked docs, since it does not create a new scope for every item in the collection.

plamut:
Looks pretty good, IMO only the creation of the dropdown menus should be refactored.

:+1:

client/app/scripts/superdesk-authoring/views/article-edit.html
```diff
@@ -1,114 +1,127 @@
...
+             data-item="item"
+             data-fieldprefix="dateline"
+             data-field="located"
+             data-postprocessing="updateDateline(item, city)"
+             data-change="autosave(item)">
+        </div>
+
+        <div class="dateline-date">
+            <select id="datelineMonth"
+                    ng-options="monthValue as monthName for (monthName, monthValue) in monthNames track by monthName"
```
pavlovicnemanja: @plamut, pls review this part, I hope this is well written.

plamut: @pavlovicnemanja  Actually, there is a potential issue here. If you look at the [ng-options docs](https://code.angularjs.org/1.4.6/docs/api/ng/directive/ngOptions), it says that `select as` and `track by` are not  designed to work together. `track by` should thus be omitted here.

BTW, `monthValue as` is also redundant, because if `select as` part is omitted, it defaults to `value` (which `monthValue` in this case).

amagdas:
:+1:

PR #1211, opened 2015-09-30
[SD-3249] Add aggregate search facilities to AAP Mulitmedia search

marwoodandrew:


server/apps/aap_mm/aap_mm_datalayer.py
```diff
@@ -54,13 +65,59 @@ def find(self, resource, req, lookup):
         query_keywords = '*:*'
         if 'query' in req['query']['filtered']:
             query_keywords = req['query']['filtered']['query']['query_string']['query']
+            query_keywords = query_keywords.replace('slugline:', 'objectname:')
+            query_keywords = query_keywords.replace('description:', 'captionabstract:')
+
+        fields = {}
+        for filter in req.get('post_filter', {}).get('and', {}):
```
plamut: I would recommend choosing a different name for the loop variable here, because it shadows the built-in `filter` class.

amagdas:
Looks good except the comment from @plamut  :+1:

amagdas:
@superdesk/developers ping

akintolga:
Andrew is off for few days, I'll try to patch it today, otherwise the update will be later this week.

akintolga:
:+1:

PR #1215, opened 2015-10-01
chore(e2e): Refactor highlights e2e tests

ioanpocol:
Rewrite tests in order to use monitoring view instead content view
Compose tests in larger suites in order to reduce the number of database
reset and prepopulation

### DIFF
--- a/client/spec/authoring_spec.js
+++ b/client/spec/authoring_spec.js
@@ -74,11 +74,12 @@ describe('authoring', function() {
         expect(authoring.getHistoryItem(0).getText()).toMatch(/Story \d+ \(Politic Desk\/one\) Created by.*/);
     });
 
-    it('view item history spike-unspike operations', function() {
+    xit('view item history spike-unspike operations', function() {
+        //it will be fixed on next PR when authoring tests will be refactored
         workspace.open();
         workspace.switchToDesk('SPORTS DESK');
-        workspace.actionOnItem('Spike', 'item6', 'Politic');
-        workspace.actionOnItem('Unspike Item', 'item6', 'Politic', 'Spiked');
+        //workspace.actionOnItem('Spike', 'item6', 'Politic');
+        //workspace.actionOnItem('Unspike Item', 'item6', 'Politic', 'Spiked');
         workspace.editItem('item6', 'Politic');
 
         authoring.showHistory();
--- a/client/spec/helpers/workspace.js
+++ b/client/spec/helpers/workspace.js
@@ -7,6 +7,7 @@ var content = require('./content'),
     nav = require('./utils').nav;
 
 function Workspace() {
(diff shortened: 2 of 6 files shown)


client/spec/authoring_spec.js
```diff
@@ -74,11 +74,12 @@ describe('authoring', function() {
...
 
-    it('view item history spike-unspike operations', function() {
+    xit('view item history spike-unspike operations', function() {
+        //it will be fixed on next PR when authoring tests will be refactored
         workspace.open();
         workspace.switchToDesk('SPORTS DESK');
-        workspace.actionOnItem('Spike', 'item5', 'Politic');
-        workspace.actionOnItem('Unspike Item', 'item5', 'Politic', 'Spiked');
+        //workspace.actionOnItem('Spike', 'item5', 'Politic');
+        //workspace.actionOnItem('Unspike Item', 'item5', 'Politic', 'Spiked');
```
mugurrus: Remove if not needed anymore

client/spec/helpers/workspace.js
```diff
@@ -67,6 +68,94 @@ function Workspace() {
...
+     */
+    this.showList = function(name) {
+        this.sideMenu.element(by.css('[title="' + name + '"]')).click();
+    };
+
+    /**
+     * Show the name highlight list
+     *
+     * @param {string} name
+     */
```
mugurrus: The documentation string is superfluous

ioanpocol: Yes, agree. Currently the rules are document all, so already proposed a discussion about this.

plamut: Just a remark here: docstrings can also be used for automatically generating documentation, and since the source code is not included there, there is a merit in writing them even if one thinks that the code itself is too straightforward for that.

(it's another story, though, whether we actually make use of that...)

client/spec/authoring_spec.js
```diff
@@ -74,11 +74,12 @@ describe('authoring', function() {
         expect(authoring.getHistoryItem(0).getText()).toMatch(/Story \d+ \(Politic Desk\/one\) Created by.*/);
     });
 
-    it('view item history spike-unspike operations', function() {
+    xit('view item history spike-unspike operations', function() {
+        //it will be fixed on next PR when authoring tests will be refactored
```
plamut: Maybe add a `TODO` word to the comment, makes it easier to find such pieces of code by a simple search - unless the promised PR is already in progress and it will indeed be issued soon. :)

plamut:
Nice work, @ioanpocol 

:+1:

client/spec/authoring_spec.js
ioanpocol: Already started to work on it, but yes a TODO statement is needed.

petrjasek:
:+1:

PR #1216, opened 2015-10-01
fix(vocabularies): SD-3371 Fixed modals width issue

pavlovicnemanja:
SD-3371

Also I enabled test, now it should pass!


amagdas:
:+1:

ioanpocol:
:+1:

PR #1219, opened 2015-10-02
fix(highlights): SD-3105 Fixed layout issue for highlights

pavlovicnemanja:
SD-3105

### DIFF
--- a/client/app/scripts/superdesk-highlights/highlights.js
+++ b/client/app/scripts/superdesk-highlights/highlights.js
@@ -165,77 +165,49 @@
     function HighlightsTitleDirective(highlightsService, $timeout) {
         return {
             scope: {
-                item: '=item',
-                orientation: '=?'
+                item: '=item'
             },
             templateUrl: 'scripts/superdesk-highlights/views/highlights_title_directive.html',
             // todo(petr): refactor to use popover-template once angular-bootstrap 0.13 is out
             link: function(scope, elem) {
-                var unmarkBox = elem.find('.unmark').hide(),
-                    icon = elem.find('i'),
-                    isOpen,
-                    closeTimeout;
 
-                scope.orientation = scope.orientation || 'right';
+                /*
+                 * Set 'open' class on dropdown menu element
+                 * @param {string} isOpen
+                 */
+                scope.toggleActions = function (isOpen) {
+                    scope.open = isOpen;
+                };
 
-                scope.$watch('item.highlights', function(_ids) {
-                    if (_ids) {
+                scope.$watch('item.highlights', function(items) {
+                    if (items) {
                         highlightsService.get().then(function(result) {
                             scope.highlights = _.filter(result._items, function(highlight) {
-                                return _ids.indexOf(highlight._id) >= 0;
-                            });
-                            // it has to first update the template before we use its html
-                            scope.$applyAsync(function() {
-                                scope.htmlTooltip = unmarkBox.html();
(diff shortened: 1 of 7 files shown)


client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -165,77 +165,49 @@
...
             link: function(scope, elem) {
-                var unmarkBox = elem.find('.unmark').hide(),
-                    icon = elem.find('i'),
-                    isOpen,
-                    closeTimeout;
 
-                scope.orientation = scope.orientation || 'right';
+                /*
+                 * Set 'open' class on dropdown menu element
+                 * @param {string} isOpen
```
plamut: This docstring is a bit misleading, since the method itself does not set any CSS classes, it just sets/clears the `open` flag on the model (it's the template that later uses this flag). This also means that `isOpen` is not a string but a boolean.
The method does also not _set_ a value, it _toggles_ it.

client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -165,77 +165,49 @@
...
                     }
                 });
 
                 /*
-                 * Removing higghlight from an item
+                 * Removing highlight from an item
                  * @param {string} highlight
                  */
-                function unmarkHighlight(highlight) {
+                scope.unmarkHighlight = function unmarkHighlight(highlight) {
```
plamut: The function name is kind of redundant here, since it is already assigned to a variable with the same name.

client/app/scripts/superdesk-highlights/views/highlights_title_directive.html
```diff
@@ -1,18 +1,23 @@
...
+
+    <ul class="dropdown-menu highlights-list-menu" ng-class="{open: open}" ng-click="$event.stopPropagation()">
+        <li>
+            <div class="menu-label" translate>Marked For</div>
+            <button class="close-button" dropdown-toggle>
+                <i class="icon-close-small icon-white"></i>
+            </button>
+        </li>
+        <li ng-repeat="highlight in highlights">
+            {{ highlight.name }}
```
plamut: `track by highlight._id` should be added here, since the highlights list tends to change. We should also probably use one-time binding for the name.

plamut:
Not sure if an e2e test is needed, I'll let the others make the verdict, but looks fine more or less, please just have a look at a couple of minor remarks.

:+1:

client/app/scripts/superdesk-highlights/highlights.js
mugurrus: Actually it sets a value: scope.open = isOpen :)
But I agree with the rest of the comment

plamut: Ah, of course, I meant just the opposite, i.e. the method _sets_ a value but does _not_ toggle it. Thanks! :)

mugurrus:
:+1:

pavlovicnemanja:
@plamut e2e tests are there, I just improved them. :)

plamut:
@pavlovicnemanja That's great!

pavlovicnemanja:
Not sure, I also so that, but that wasn't added by me, so I'm not sure should we remove it or not

plamut:
It seems this was added by accident 3 weeks ago, but only now Travis started complaining about it (the script returning an error). Just remove it and it should be fine. :)

ioanpocol:
:+1:

PR #1220, opened 2015-10-05
fix(preferences): reload preferences on session expiry

petrjasek:
in case fetching preferences fails with auth error it will try
to refetch those using old session id, but that does not exist
anymore.. so now on 404 it will fetch again using new session id.


amagdas:
:+1:

pavlovicnemanja:
:+1:

PR #1221, opened 2015-10-05
Move publish_queue, vocabularies and some formatters to core.

amagdas:
Refactor archive_publish to publish.content module.


server/settings.py
```diff
@@ -162,6 +162,7 @@ def env(variable, fallback_value=None):
     'superdesk.upload',
     'superdesk.notification',
     'superdesk.activity',
+    'superdesk.vocabularies',
```
sivakuna-aap: Any specific reason for registering `superdesk.vocabularies` in both `superdesk` and `superdesk-core`? Ideally it should be registered in `superdesk-core` only.

plamut:
In general it looks good to me, although it's a bit difficult to keep track of everything, might have overlooked something...

:+1:

server/settings.py
amagdas: Nop, it's not like that. The `settings.py` file from sd-core is not shared at all, that is used only for setting up the app for testing purposes.
Each app has to define complete settings as we don't want to impose on apps what modules to use.

sivakuna-aap: Gotcha

sivakuna-aap:
@amagdas whats the timeline to merge this PR? [PR#1222](https://github.com/superdesk/superdesk/pull/1222) is blocked by this.

amagdas:
@sivakuna-aap the timeline is as soon as I get the build passing :)

PR #1223, opened 2015-10-06
(fix) Dateline

sivakuna-aap:
1. Values for all the options in Dateline Day are same. Because of this, the selected day is always the last day in the month although the current date is before last day of the month.

2. Timezone of a city selected from locators vocabulary is overwritten by UTC timezone by onblur event of the directive.


akintolga:
:+1:

plamut:
Do the existing tests cover the functionality or should a few more be added?
Looks good though.

:+1:

mdhaman:
:+1:

PR #1229, opened 2015-10-07
(feat): Allow users to view content send to another desk.

mdhaman:
This feature allow the user to monitor content sent from a particular desk to all desks or a particular desk. 
See https://dev.sourcefabric.org/browse/SD-3352 for more information:

- [x] Modify the desk schema to specify `desk_types`
- [x] Modify the desk frontend for `desk_types`.
- [x] Modify the archive/item schema to track the `last_authoring_desk` and `last_production_desk`.
- [x] Modify the 'send_to' related functionality to track the desk movement.
- [x] Modify Global search on frontend to allow to selection for desks.

### DIFF
--- a/client/app/scripts/superdesk-search/tests/search_spec.js
+++ b/client/app/scripts/superdesk-search/tests/search_spec.js
@@ -21,7 +21,7 @@ describe('search service', function() {
         expect(criteria.query.filtered.query.query_string.query).toBe('test');
     }));
 
-    it('can create query for from_desk and to_desk', inject(function($rootScope, search) {
+    it('can create query for from_desk', inject(function($rootScope, search) {
         // only from desk is specified
         var criteria = search.query({from_desk: 'test-authoring'}).getCriteria();
         var filters = criteria.query.filtered.filter.and;
@@ -29,49 +29,26 @@ describe('search service', function() {
         criteria = search.query({from_desk: 'test-production'}).getCriteria();
         filters = criteria.query.filtered.filter.and;
         expect(filters).toContain({term: {'task.last_production_desk': 'test'}});
+    }));
 
+    it('can create query for to_desk', inject(function($rootScope, search) {
         // only to desk is specified
-        criteria = search.query({to_desk: '456-authoring'}).getCriteria();
-        filters = criteria.query.filtered.filter.and;
+        var criteria = search.query({to_desk: '456-authoring'}).getCriteria();
+        var filters = criteria.query.filtered.filter.and;
         expect(filters).toContain({term: {'task.desk': '456'}});
         expect(filters).toContain({exists: {field: 'task.last_production_desk'}});
         criteria = search.query({to_desk: '456-production'}).getCriteria();
         filters = criteria.query.filtered.filter.and;
         expect(filters).toContain({term: {'task.desk': '456'}});
         expect(filters).toContain({exists: {field: 'task.last_authoring_desk'}});
-
-        // both from desk and to desk are specified
-        criteria = search.query({from_desk: '123-authoring', to_desk: '456-production'}).getCriteria();
-        filters = criteria.query.filtered.filter.and;
-        expect(filters).toContain({term: {'task.last_authoring_desk': '123'}});
-        expect(filters).toContain({term: {'task.desk': '456'}});
     }));
 
-    it('create tags for from desk', inject(function ($location, $rootScope, $q, tags, _desks_) {
-        var desks = _desks_;
-        spyOn(desks, 'initialize').and.returnValue($q.when([]));
-
-        desks.deskLookup = {
-            from: {
-                name: 'National'
-            },
-            to: {
-                name: 'Sport'
-            }
-        };
-
-        $location.search('from_desk', 'from-authoring');
-        $location.search('to_desk', 'to-authoring');
-
-        var tagsList = null;
-        tags.initSelectedFacets().then(function(value) {
-            tagsList = value;
-        });
-
-        $rootScope.$digest();
-        expect(tagsList.selectedParameters.length).toEqual(2);
-        expect(tagsList.selectedParameters[0]).toEqual('From Desk:National');
-        expect(tagsList.selectedParameters[1]).toEqual('To Desk:Sport');
+    it('can create query for from_desk and to_desk', inject(function($rootScope, search) {
+        // both from desk and to desk are specified
+        var criteria = search.query({from_desk: '123-authoring', to_desk: '456-production'}).getCriteria();
+        var filters = criteria.query.filtered.filter.and;
+        expect(filters).toContain({term: {'task.last_authoring_desk': '123'}});
+        expect(filters).toContain({term: {'task.desk': '456'}});
     }));
 
     it('can sort items', inject(function(search, $location, $rootScope) {
--- a/client/app/scripts/superdesk-search/tests/search_spec.js
+++ b/client/app/scripts/superdesk-search/tests/search_spec.js
@@ -21,6 +21,59 @@ describe('search service', function() {
         expect(criteria.query.filtered.query.query_string.query).toBe('test');
     }));
 
+    it('can create query for from_desk and to_desk', inject(function($rootScope, search) {
+        // only from desk is specified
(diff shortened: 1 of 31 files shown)


server/apps/desks.py
```diff
@@ -17,6 +18,9 @@
 from superdesk.notification import push_notification
 from superdesk.activity import add_activity, ACTIVITY_UPDATE
 
+desk_types = ['authoring', 'production']
```
sivakuna-aap: Since we moved to `Python 3.4` we can use Enums

mdhaman:
**Requirements:** The requirement is to show the output of the desk i.e. if item created on "Sydney" desk and then sent to "National" desk then the users (Sydney Desk and maybe others) need to see the version that was submitted to the "National" desk. Basically, the content that is moved between desk (not published).

Following are possible solutions for ticket [SD-3352](https://dev.sourcefabric.org/browse/SD-3352):
- Add new resource as `archive_desk_output` with elasticsearch as search backend . On move action, create an document in the collection. Create a view in the monitoring for each desk as well as allow user to add desk output view in custom workspace. Initially, items in the desk output view are  going to be read only and would need feedback from POs for actions that can be performed on the desk output item. This will have document integrity issues like `published` collection.

- To enable elasticsearch index on the archive_versions collection and use that collection to filter out documents base `item_operation='move'` and `from_desk` (`from_desk` field will be populated in the move action). This might create a large index (for example 20000/day docs with 5 versions/doc = 100000/day document in elasticsearch).

@petrjasek @amagdas any opinions for the proposed solutions.

amagdas:
How about to keep the operations performed on  a story in an array so the item in `archive` endpoint will have the full history of the actions performed on it.
This will allow to use the current index to perform the filtering. 
If my proposal doesn't make sense I would vote for indexing `archive_versions`, creating yet another collection where to keep the data doesn't really make sense in my opinion.

@petrjasek @mdhaman thoughts?

mdhaman:
@amagdas if we keep full history of actions performed in the document then how will the search work. what will be stored in the array? Something like below?
```
    "action_history": [
        {
            "operation": "update",
            "version": 1,
            "headline": "This is finance headline",
            "slugline": "This is finance slugline",
            .....
        },
        {
            "operation": "move",
            "version": 2,
            "headline": "This is news headline",
            "slugline": "This is news slugline",
            .....
        },
        {
            "operation": "update",
            "version": 3,
            "headline": "This is blah headline",
            "slugline": "This is blah slugline",
            .....
        },
        {
            "operation": "move",
            "version": 4,
            "headline": "This is sports headline",
            "slugline": "This is sports slugline",
            .....
        }        
        
    ]
```
Look at eve could not find any nested document info. Maybe archive_versions might be a good idea but I am worried about the index size.

amagdas:
Yes, something like that but a lot less shorter, we should keep only the
info needed to actually implement the search, we have the versions already
if we want to show a specific version.

mdhaman:
if we use the array approach then you might get unwanted search results. basically if you have search for 'finance' in headline then you would get the hit even though the current headline is 'this is sports headline'

client/app/scripts/superdesk-search/tests/search_spec.js
```diff
@@ -21,6 +21,59 @@ describe('search service', function() {
...
+        filters = criteria.query.filtered.filter.and;
+        expect(filters).toContain({term: {'task.desk': '456'}});
+        expect(filters).toContain({exists: {field: 'task.last_authoring_desk'}});
+
+        // both from desk and to desk are specified
+        criteria = search.query({from_desk: '123-authoring', to_desk: '456-production'}).getCriteria();
+        filters = criteria.query.filtered.filter.and;
+        expect(filters).toContain({term: {'task.last_authoring_desk': '123'}});
+        expect(filters).toContain({term: {'task.desk': '456'}});
+    }));
```
plamut: I would recommend splitting this into three separate test cases, with each covering only a single use case - more specific error report if something goes wrong.

client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -88,6 +88,30 @@
                 <div sd-meta-terms data-item="subjectitems" data-field="subject" data-unique="qcode" data-list="subjectcodes" data-change="subjectSearch(subjectitems)"></div>
             </div>
         </div>
+        <div class="field">
+            <label for="from-desk">
+                {{:: 'From Desk' | translate}}
+            </label>
+            <select
+                id="from-desk" name="from-desk"
+                ng-options="d._id as d.name for d in desks._items track by d._id"
```
plamut: As per [ng-options docs](https://docs.angularjs.org/api/ng/directive/ngOptions), the `select` and `track by` parts of the expression are not designed to work together.
It is not an issue in this case, since they are the same, but it could become if somebody later makes changes to it. It's thus probably best to simply remove the `track by` part.

client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -88,6 +88,30 @@
...
+            </select>
+        </div>
+
+        <div class="field">
+            <label for="to-desk">
+                {{:: 'To Desk' | translate}}
+            </label>
+            <select
+                id="to-desk" name="to-desk"
+                ng-options="d._id as d.name for d in desks._items track by d._id"
```
plamut: Same comment as the previous one (`select` and `track by` together).

client/app/scripts/superdesk-search/views/search-facets.html
```diff
@@ -13,7 +13,7 @@
 
         <div sd-search-within></div>
 
-        <div sd-toggle-box data-title="{{ 'Parameters' | translate }}" data-open="false">
+        <div id="search-parameters" sd-toggle-box data-title="{{ 'Parameters' | translate }}" data-open="false">
```
plamut: We could use one-time binding here... or is `translate` a stateless filter? We can use it anyway, just to be safe...

client/spec/search_spec.js
```diff
@@ -71,4 +73,40 @@ describe('Search', function() {
...
+        globalSearch.selectDesk('to-desk', 'Politic Desk');
+        expect(element.all(by.repeater('items._items')).count()).toBe(1);
+        expect(globalSearch.getHeadlineElement(0).getText()).toBe('From-Sports-To-Politics');
+
+        globalSearch.selectDesk('from-desk', '');
+        expect(element.all(by.repeater('items._items')).count()).toBe(1);
+        expect(globalSearch.getHeadlineElement(0).getText()).toBe('From-Sports-To-Politics');
+
+        globalSearch.selectDesk('to-desk', '');
+        expect(element.all(by.repeater('items._items')).count()).toBe(11);
```
plamut: Would it make sense to extract the `element.all(by.repeater('items._items'))` part into a helper object/function? It is repeated a lot throughout the code...

plamut:
It's a bit difficult to see a forest because of all the trees, but it looks good generally (apart for some minor comments). :)
I didn't check the exact test coverage, but it seems like everything is covered.

:+1:

client/spec/search_spec.js
```diff
@@ -71,4 +73,40 @@ describe('Search', function() {
         priority3.click();
         expect(element.all(by.repeater('items._items')).count()).toBe(1);
     });
+
+    it('can search by from desk field', function() {
+        workspace.switchToDesk('SPORTS DESK').then(content.setListView);
+        expect(element.all(by.repeater('items._items')).count()).toBe(2);
```
mugurrus: element.all(by.repeater('items._items')) should be moved in a search helper method

mugurrus:
@amagdas As @mdhaman pointed there are issues on search, there is also data redundancy. If there is the need to search through operations archive_versions can be indexed.

mugurrus:
:+1: 
seems OK, quite a long PR :)

sivakuna-aap:
:+1: Looks good. Is it possible to merge the small commits into logical ones?

mdhaman:
@mugurrus , @amagdas  the requirements have change again. I will strike thru my comments.

mdhaman:
@sivakuna-aap may be next time. :wink:

PR #1232, opened 2015-10-09
(feat & imp) SD-3183 and Vocabulary UI Improvements

sivakuna-aap:
1. [SD-3183](https://dev.sourcefabric.org/browse/SD-3183) Using keywords as labels to group content arbitrarily
2. Adding following features to Vocabularies:
      1. Adding "type" to each vocabulary which defines whether a vocabulary is system driven or manageable by user.
      2. Adding "display_name" to each vocabulary which is used in UI.
      3. Vocabulary UI displays only manageable type vocabularies
      4. Vocabulary UI displays both active and inactive vocabularies
3. Fixed the below issues in Vocabularies UI:
      1. Updating a vocabulary item removes is_active flag from collection
      2. No vocabulary should be allowed to delete as it might have been used in transactional data. So replaced delete with sd-switch

Requires https://github.com/superdesk/superdesk-core/pull/35

##### Instructions to the Merger(s):
1. Underlying structure of Vocabulary is changed. After merging please drop vocabularies collection and recreate collection using `vocabularies.json` file.

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-words-list.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-words-list.html
@@ -1,8 +1,16 @@
-<div class="term-editor data visible">
-    <input class="line-input" ng-model="term" type="text" ng-disabled="disabled" autocomplete="off" ng-keyup="selectTerm($event, term)"/>
+<div class="term-editor data visible" ng-show="header">
+    <div sd-typeahead items="words" term="selectedTerm" search="search(term)" select="select(item)" data-disabled="disabled">
+        <ul>
+            <li typeahead-item="t" ng-repeat="t in words">
+                {{ :: t.name }}
+            </li>
+        </ul>
+    </div>
 </div>
+
 <div class="terms" ng-if="item[field].length">
     <ul>
-        <li ng-repeat="t in item[field]" ng-click="removeTerm(t)">{{ t }}</li>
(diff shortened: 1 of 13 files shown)


client/app/scripts/superdesk-authoring/metadata/views/metadata-words-list.html
```diff
@@ -1,8 +1,16 @@
...
+            </li>
+        </ul>
+    </div>
 </div>
+
 <div class="terms" ng-if="item[field].length">
     <ul>
-        <li ng-repeat="t in item[field]" ng-click="removeTerm(t)">{{ t }}</li>
+        <li ng-if="!disabled" ng-repeat="t in item[field]" ng-click="removeTerm(t)">{{ t }}<i class="icon-close-small"></i></li>
+        <li ng-if="disabled" ng-repeat="t in item[field]">{{ t }}</li>
```
plamut: Terms are not unique, right? Can't track them by anything?

plamut:
Looks good IMO. :+1:

client/app/scripts/superdesk-authoring/metadata/views/metadata-words-list.html
sivakuna-aap: I think the terms are unique. I will add `tack by`

mdhaman:
:+1:

akintolga:
:+1:

PR #1233, opened 2015-10-12
(fix) [SD-2475] Bylines and datelines need to be upper and lower case

sivakuna-aap:


akintolga:
:+1:

mdhaman:
:+1:

client/app/scripts/superdesk-authoring/styles/themes.less
```diff
@@ -270,11 +270,14 @@ body, html {
 
 	  	.dateline {
 		  	display: flex;
+            input {
+              text-transform: uppercase;
+            }
+
```
plamut: (minor) Can we fix the indentation of this block for better readability?

plamut:
Is there a test for the correct dateline format? IMO this is something worth having.

Looks good though. :+1:

sivakuna-aap:
There are behave tests around dateline but none of them check the format because it has current date. I will check those once again and add one for checking the format if possible.

plamut:
OK, sounds good (but if it is not possible, then we can leave it as is I guess, since it doesn't increase the technical debt).

sivakuna-aap:
@plamut added some unit tests for checking the format

plamut:
@sivakuna-aap Cool!

PR #1234, opened 2015-10-12
[SD-3399] - Text archive import command registered

akintolga:


server/apps/aap/import_text_archive/commands.py
```diff
@@ -152,7 +153,7 @@ def _process_bunch(self, x):
             item['firstcreated'] = utc.normalize(value) if value.tzinfo else value
             item['versioncreated'] = item['firstcreated']
 
-            item['unique_id'] = doc.find('dcdossier').get('unique')
+            generate_unique_id_and_name(item)
             item['ingest_id'] = id
```
plamut: Just checking - is the `ingest_id` attribute supposed to be a function?

plamut:
:+1:

server/apps/aap/import_text_archive/commands.py
akintolga: I think it is an attribute.

plamut: Sorry, should have written more clearly. :) Is the _value_ of this attribute meant to be a function?

sivakuna-aap:
:+1: Can we replace strings with constants in `server/apps/aap/import_text_archive/commands.py`?

marwoodandrew:
Wouldn't bother making it to pretty, should only ever be used once!

:+1:

PR #1235, opened 2015-10-12
(fix) [SD-2725] Locator code works but it would be nice to separate the dropdown into two lists - Australia and Rest of World

sivakuna-aap:
##### Instructions to the Merger(s):
1. Underlying structure of Vocabulary is changed. After merging please drop vocabularies collection and recreate collection using `vocabularies.json` file.


sivakuna-aap:
@superdesk/developers

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -193,6 +193,14 @@ function MetadataDropdownDirective() {
                 _.extend(scope.item, o);
                 scope.change({item: scope.item});
             };
+
+            $timeout(function() {
```
akintolga: Is the $timeout necessary here?

sivakuna-aap: Yes, what I observed is, sometimes the variables declared in directive scope doesn't get values until the digest completes.

client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -193,6 +193,14 @@ function MetadataDropdownDirective() {
...
+                if (scope.list && scope.field === 'place') {
+                    scope.places = _.groupBy(scope.list, function(place) {
+                        return place.country === 'Australia' ? 'Australia' : 'Rest of World';
```
amagdas: `gettext`?

akintolga:
:+1:

client/app/scripts/superdesk-authoring/metadata/metadata.js
ioanpocol: Can you make the 'Australia' string as a setting value in settings file?

ioanpocol:
:+1:  except comments

client/app/scripts/superdesk-authoring/metadata/metadata.js
sivakuna-aap: We had a discussion about this. I will move it to vocabularies itself. Adding it there will give more flexibility in terms if tomorrow we want to group them into more than 2.

ioanpocol: Yes, this is better

sivakuna-aap:
@ioanpocol replaced hard coding with vocabulary configuration.

ioanpocol:
:+1:

PR #1237, opened 2015-10-13
[SD-3334] - On send or send and continue unlock isn't showing

Aca-jov:
And on desk - adding people, people are added on top (discussed with
Fritz, no issue for this)

### DIFF
--- a/client/app/scripts/superdesk-authoring/views/authoring-topbar.html
+++ b/client/app/scripts/superdesk-authoring/views/authoring-topbar.html
@@ -9,7 +9,7 @@
     </div>
 
     <div class="button-stack right-stack">
-        <button class="btn btn-medium pull-left" ng-if="item._locked && can_unlock() && itemActions.save" ng-click="unlock()" ng-disabled="unlockClicked" translate>Unlock</button>
+        <button class="btn btn-medium pull-left" ng-if="item._locked && !item.more_coming && can_unlock() && itemActions.save" ng-click="unlock()" ng-disabled="unlockClicked" translate>Unlock</button>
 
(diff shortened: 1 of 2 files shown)


pavlovicnemanja:
:+1:

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
```diff
@@ -9,7 +9,7 @@
     </div>
 
     <div class="button-stack right-stack">
-        <button class="btn btn-medium pull-left" ng-if="item._locked && can_unlock() && itemActions.save" ng-click="unlock()" ng-disabled="unlockClicked" translate>Unlock</button>
+        <button class="btn btn-medium pull-left" ng-if="item._locked && !item.more_coming && can_unlock() && itemActions.save" ng-click="unlock()" ng-disabled="unlockClicked" translate>Unlock</button>
```
plamut: I suggest wrapping  this line as it is almost 200 characters long, requiring substantial horizontal scrolling.
Maybe also consider putting the `ng-if` condition logic into the controller, it's starting to become a bit too complex to have it in the template.

plamut:
:+1:

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
ioanpocol: Is this working? You added a new condition in a list of 'and' conditions that evaluated is false, so the new added condition change nothing.

Aca-jov: This value is only set on send to, so only there is true and in all other cases its false, so that's why I added that.

PR #1243, opened 2015-10-14
[sd-3400]-(Content Upload):Allow upload and image or video when metad…

sjunaid:
…ata provided (Covers SD[1348, 1908, 1910,1957])

### DIFF
--- a/client/app/scripts/superdesk-archive/controllers/upload.js
+++ b/client/app/scripts/superdesk-archive/controllers/upload.js
@@ -7,6 +7,8 @@ define(['lodash'], function(_) {
         $scope.items = [];
         $scope.saving = false;
         $scope.failed = false;
+        $scope.enableSave = false;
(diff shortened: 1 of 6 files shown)


client/app/scripts/superdesk-archive/controllers/upload.js
```diff
@@ -7,6 +7,8 @@ define(['lodash'], function(_) {
         $scope.items = [];
         $scope.saving = false;
         $scope.failed = false;
+        $scope.enableSave = false;
+        var requiredFields = ['headline', 'description', 'byline', 'slugline'];
```
sivakuna-aap: if `byline` is available in user profile then it should be auto-populated. This is how it behaves in case of text articles/packages. It should behave same in case of upload also.

ioanpocol:
:+1:

amagdas:
:+1:

PR #1245, opened 2015-10-14
fix(editor): Fixed spellchecker to be responsive during updating markup.

eozan:
SD-3309

### DIFF
--- a/client/app/scripts/superdesk/editor/editor.js
+++ b/client/app/scripts/superdesk/editor/editor.js
@@ -125,8 +125,8 @@ function HistoryStack(initialValue) {
     };
 }
 
-EditorService.$inject = ['spellcheck', '$rootScope'];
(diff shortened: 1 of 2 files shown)


mugurrus:
:+1:

ioanpocol:
:+1: 
except one unittest that is failing

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -317,15 +317,22 @@ function EditorService(spellcheck, $rootScope) {
...
 
-        self.resetSelection(node);
+        if (!preventStore) {
+            self.storeSelection(node);
+        }
+        var token = tokens.shift();
+        hiliteToken(node, token, className);
+        $timeout(function() {
+            hilite(node, tokens, className, true);
+        }, 0);
```
petrjasek: i think you can also disable `$apply` here

PR #1246, opened 2015-10-14
test(e2e): Test PR

ioanpocol:
This PR will not be merged and the scope is to see if reconnect popup
slowdown e2e tests


ioanpocol:
Reconnect popup is not an issue for e2e tests

PR #1250, opened 2015-10-14
[SD-3424] - Improvements in associated items

akintolga:


client/app/scripts/superdesk/itemList/views/relatedItem-list-widget.html
```diff
@@ -69,7 +69,7 @@
                     <div class="basic item-block" 
                         ng-click="view(item)">
                         <div class="content">
-                            <p><i class="filetype-icon-{{item.type}}"></i><span class="keyword" ng-if="item.slugline">{{ item.slugline }}</span> {{item.headline || item.type}}</p>
+                            <p><i class="filetype-icon-{{item.type}}"></i><span class="keyword" ng-if="item.slugline">{{ item.slugline }}</span> {{item.headline || item.type}} {{item.anpa_take_key}}</p>
```
sivakuna-aap: It would be good if we display the `item.type` always.

akintolga: Item type is displayed as the icon

sivakuna-aap: Right. What I meant was the icon should be displayed always and shouldn't be ignored if headline exists.

marwoodandrew:
:+1:

PR #1252, opened 2015-10-15
Fix handling different time zones in ingest routing schedule

plamut:
Resolves [SD-3350](https://dev.sourcefabric.org/browse/SD-3350).

This PR fixes various issues with ingest routing schedule, e.g.:
* `sdTimepickerAlt` can be optionally configured to *not* convert selected times to UTC in the client browser, avoiding problems when time zone settings on the client's machine differ from the server time zone.
* It is now possible to select the time zone of the times of day in a routing schedule rule. This avoids the trouble of defining a time interval in a local time zone that would span across two different days in UTC due to the time shift (the latter was pretty much impossible to define until now).
The default behavior of the directive is still the same, preserving the backward compatibility.
If no time zone is defined in the routing schedule, UTC is assumed, meaning that all the existing schedules that do not have this information will still work normally.
* It is now possible to omit the ending time of day in a routing rule, which is interpreted as "until the end of the day". Until now it was not possible to do that, the latest possible time to choose for an end time was 23:55, meaning that there was a 5-minute gap until midnight, which was impossible  to cover.
* Optimization: time zone data is now fetched from the server only once - when the `tzdata` factory is instantiated. Before the change this data was fetched every time when a service/directive that uses `tzdata` was created.

**To be aware of:**
Currently the list of all time zones on the client and the server differ. While the server recognizes almost 500 common time zones (using `pytz`, based on the Olson database), the client only know about 50 of them. The reason is that `tzdata` service fetches the `world-clock/timezones-all.json` file which seems to be incomplete.

### DIFF
--- a/client/app/scripts/superdesk-ingest/module.js
+++ b/client/app/scripts/superdesk-ingest/module.js
@@ -1031,11 +1031,14 @@ define([
         };
     }
 
-    IngestRoutingSchedule.$inject = [];
-    function IngestRoutingSchedule() {
+    IngestRoutingSchedule.$inject = ['config'];
+    function IngestRoutingSchedule(config) {
         return {
             scope: {rule: '='},
-            templateUrl: 'scripts/superdesk-ingest/views/settings/ingest-routing-schedule.html'
+            templateUrl: 'scripts/superdesk-ingest/views/settings/ingest-routing-schedule.html',
+            link: function (scope, element, attrs) {
+                scope.serverTimezone = config.server.timezone;
+            }
         };
     }
 
--- a/client/app/scripts/superdesk-ingest/module.js
+++ b/client/app/scripts/superdesk-ingest/module.js
@@ -1055,10 +1055,7 @@ define([
                 scope.matchingTimeZones = [];
 
                 tzdata.$promise.then(function () {
-                    scope.timeZones = _.union(
-                        _.keys(tzdata.zones),
-                        _.keys(tzdata.links)
-                    );
+                    scope.timeZones = tzdata.getTzNames();
                 });
 
                 /**
--- a/client/app/scripts/superdesk-ingest/module.js
+++ b/client/app/scripts/superdesk-ingest/module.js
@@ -1031,13 +1031,86 @@ define([
         };
     }
 
-    IngestRoutingSchedule.$inject = ['config'];
-    function IngestRoutingSchedule(config) {
+    /**
+     * @memberof superdesk.ingest
+     * @ngdoc directive
+     * @name sdIngestRoutingSchedule
+     * @description
+     *   Creates the Schedule section (tab) of the routing rule edit form.
+     */
+    IngestRoutingSchedule.$inject = ['config', 'tzdata'];
+    function IngestRoutingSchedule(config, tzdata) {
         return {
-            scope: {rule: '='},
+            scope: {
+                rule: '='  // the routing rule whose schedule is being edited
+            },
             templateUrl: 'scripts/superdesk-ingest/views/settings/ingest-routing-schedule.html',
             link: function (scope, element, attrs) {
+                // TODO: remove? or rename to default timezone?
                 scope.serverTimezone = config.server.timezone;
+
+                scope.timeZones = [];     // all time zones to choose from
+                scope.tzSearchTerm = '';  // the current time zone search term
+
+                // filtered time zone list containing only those that match
+                // user-provided search term
+                scope.matchingTimeZones = [];
+
+                tzdata.$promise.then(function () {
+                    scope.timeZones = _.union(
+                        _.keys(tzdata.zones),
+                        _.keys(tzdata.links)
+                    );
+                });
+
+                /**
+                 * Sets the list of time zones to select from to only those
+                 * that contain the given search term (case-insensitive).
+                 * If the search term is empty, it results in an empty list.
+                 *
+                 * @method searchTimeZones
+                 * @param {string} searchTerm
+                 */
+                scope.searchTimeZones = function (searchTerm) {
+                    var termLower;
+
+                    scope.tzSearchTerm = searchTerm;
+
+                    if (!searchTerm) {
+                        scope.matchingTimeZones = [];
+                        return;
+                    }
+
+                    termLower = searchTerm.toLowerCase();
+                    scope.matchingTimeZones = _.filter(
+                        scope.timeZones,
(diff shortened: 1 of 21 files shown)


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -1031,11 +1031,85 @@ define([
...
+            link: function (scope, element, attrs) {
+
+                scope.timeZones = [];     // all time zones to choose from
+                scope.tzSearchTerm = '';  // the current time zone search term
+
+                // filtered time zone list containing only those that match
+                // user-provided search term
+                scope.matchingTimeZones = [];
+
+                tzdata.$promise.then(function () {
```
sivakuna-aap: The same is being done in multiple places. Can we refactor it to a method in tzdata and use that method in other places?

plamut: What do you mean, the `$promise` part or the `_.union()` logic?
The former is needed because TZ data is fetched asynchronously and services that use it must know when the data has arrived.
The latter could be done, the "unioned" data could be pre-caclulated and stored as ... hm, some sort of a property of the `tzdata` Array? Is that what you had in mind?

client/app/scripts/superdesk-ingest/views/settings/ingest-routing-schedule.html
```diff
@@ -6,13 +6,55 @@
...
+                items="matchingTimeZones"
+                term="tzSearchTerm"
+                search="searchTimeZones(term)"
+                select="selectTimeZone(item)"
+                always-visible="tzSearchTerm">
+                <li
+                    typeahead-item="tz"
+                    ng-repeat="tz in matchingTimeZones track by tz"
+                    >{{ ::tz }}</li>
+                <li ng-show="tzSearchTerm && matchingTimeZones.length === 0"
```
sivakuna-aap: Is displaying `no matches` necessary? We are not displaying this text anywhere across superdesk. Either we keep for all or for none but not mixed. @superdesk/developers what you think?

plamut: I found it quite user-unfriendly when you type something and get no results, it almost seems like the widget is not working because of the lack of visual feedback. I can remove it for the sake of consistency, but would personally recommend adding it to other places as well.

Of course it's the UX specialists who have the final word on this, please let us know what you think, thanks! :)

server/apps/rules/routing_rules.py
```diff
@@ -140,6 +143,10 @@ class RoutingRuleSchemeResource(Resource):
                             },
                             'hour_of_day_to': {
                                 'type': 'string'
+                            },
+                            'time_zone': {
+                                'type': 'string',
+                                'nullable': True
```
sivakuna-aap: The PR says the default timezone in UTC. if that's the case then this should be
```
'time_zone': {
    'type': 'string',
    'nullable': False,
    'default': 'UTC'
```

I know that this breaks the existing rules. But that's fine we can clean up that collection.

server/apps/rules/routing_rules.py
```diff
@@ -301,20 +329,52 @@ def __check_if_rule_name_is_unique(self, routing_scheme):
...
             schedule = rule.get('schedule', {})
             if schedule:
-                from_time = set_time(current_datetime, schedule.get('hour_of_day_from', '0000'))
-                to_time = set_time(current_datetime, schedule.get('hour_of_day_to', '2359'), second=59)
-                if not (Weekdays.is_scheduled_day(current_datetime, schedule.get('day_of_week', []))
-                        and (from_time < current_datetime < to_time)):
-                    is_scheduled = False
+                # adjust current time to the schedule's timezone
+                tz_name = schedule.get('time_zone')
+                schedule_tz = pytz.timezone(tz_name) if tz_name else pytz.utc
```
sivakuna-aap: `pytz.timezone(tz_name)` will fail if tz_name is None. Right?

plamut: Well no, because it won't be executed in that case. :)

server/apps/rules/routing_rules.py
plamut: I wanted to avoid breaking the existing data that does not have this info, and any other code that might be using this schema and not sending the time zone name (it's basically the same way the things have worked until now, the times were stored in UTC without explicitly mentioning the time zone).

What do others think, is it worth breaking the backward compatibility in this case? I can adjust the schema of course.

plamut:
**Updated PR:** introduced `getTzNames()` utility method to the `tzdata` service to avoid repeating the same logic at  multiple places.

client/app/scripts/superdesk-ingest/module.js
plamut: For the reference: this has been resolved, but GitHub did not detect it, as the commented-on line has remained unchanged.

amagdas:
:+1:

client/app/scripts/superdesk-ingest/views/settings/ingest-routing-schedule.html
petrjasek: ping @fritzSF

server/apps/rules/routing_rules.py
petrjasek: if we want to be able to use server local time then `nullable: True` makes sense, but would probably use `default: 'UTC'` anyhow and let client set explicitly if user wants to use server local time. but not sure if it's something we want to support..

plamut:
Rebased so that merge conflicts are resolved.

server/apps/rules/routing_rules.py
plamut: To not lose too much time on guessing what we should or should not support (:smile:), I'll just specify the 'UTC' as a default  time zone in schema as suggested. It's also easier to break backward compatibility now when the system is not in production yet.

plamut:
Updated, UTC is now defined as a default routing schedule time zone in the schema, too.

server/apps/rules/routing_rules.py
petrjasek: :ok:

petrjasek:
:+1:

amagdas:
:+1:

plamut:
Has @fritzSF confirmed that a default `"no matches"` line should be retained when the time zone search yields no results?
Should I open a ticket to add the same improvement to other places, too, where `sd-typeahead` is used?

PR #1257, opened 2015-10-16
[SD-3166] - Use the 'en' dictionary if the language of story is 'en-US' or 'en-AU'

akintolga:


sivakuna-aap:
:+1:

plamut:
:+1:

mdhaman:
:+1:

ioanpocol:
If there are also en-US and/or en-AU dictionaries, those dictionaries should be used.
I think this case is not covered here (at least in tests)

akintolga:
You are right Ioan, I'll add a test for that case. Thanks.

petrjasek:
looks good, but you might also need to make some changes on backend - in `dictionaries/service` there is `get_model_for_lang` function, it is used for suggestions, so it should use the same logic

akintolga:
I'm closing this PR for the time being, will create another one later.

PR #1259, opened 2015-10-16
chore(e2e): Refactor authoring e2e tests

ioanpocol:


client/protractor-conf.js
```diff
@@ -25,7 +25,7 @@ exports.config = {
     framework: 'jasmine2',
     jasmineNodeOpts: {
         showColors: true,
-        defaultTimeoutInterval: 55000
+        defaultTimeoutInterval: 120000
```
amagdas: Whoa, this is quite a bit timeout.

ioanpocol: This is the time a test can run and because we have now larger tests suits, more time is needed.

mugurrus:
:+1:

amagdas:
:+1:

client/protractor-conf.js
ioanpocol: Anyway seams that larger tests are not a good idea, so we have to talk about it on retrospective meeting.

PR #1266, opened 2015-10-19
feat(workspace): SD-3411 By using keyboard shortcuts, change the current showed view

pavlovicnemanja:
Main task: SD-774

### DIFF
--- a/client/app/scripts/superdesk-workspace/workspace.js
+++ b/client/app/scripts/superdesk-workspace/workspace.js
@@ -303,8 +303,8 @@
         };
     }
 
-    WorkspaceSidenavDirective.$inject = ['superdeskFlags'];
-    function WorkspaceSidenavDirective(superdeskFlags) {
+    WorkspaceSidenavDirective.$inject = ['superdeskFlags', '$location', 'keyboardManager'];
+    function WorkspaceSidenavDirective(superdeskFlags, $location, keyboardManager) {
         return {
             templateUrl: 'scripts/superdesk-workspace/views/workspace-sidenav-items.html',
             link: function(scope) {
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-workspace/workspace.js
```diff
@@ -324,6 +324,36 @@
                         superdeskFlags.flags.hideMonitoring = false;
                     }
                 };
+
+                /*
+                 * By using keyboard shortcuts, change the current showed view
+                 *
+                 */
+                var opt = {global: true};
+                keyboardManager.bind('alt+h', function (e) {
```
petrjasek: we were using `ctrl` + something shortcuts, so not sure we should use `alt` for these, what do you think @fritzSF ?

pavlovicnemanja: I had a talk with @fritzSF about this this morning, and we agreed to use `alt +` for global shortcuts (like changing views) and `ctrl +` for authoring shortcuts.

petrjasek: ok then

petrjasek:
does it unbind when you get out of workspace or it works always?

pavlovicnemanja:
No, it stays... Wherever you are, when you press some combination, url changes.

petrjasek:
ok that should be fine
:+1:

plamut:
Do we test keyboard shortcuts?

Looks good though. :+1:

pavlovicnemanja:
Yeah, I need to write e2e tests for them! :-) I will update this PR today ;-)

plamut:
OK, good, I added the WIP label so that this does not get merged prematurely. :)

PR #1275, opened 2015-10-20
SD-3438

darconny:
Improvements od modal windows

PR #1280, opened 2015-10-20
[SD-3473] - Content filter test popup body is hidden

akintolga:
Modal-Body was not visible after PR #1275


plamut:
:+1: 

I am not sure though if this solves all the problems, can you please check if this fixes the routing scheme edit modal as well? (that one got broken as well, and there may be others, too...)

akintolga:
These fixes routing and content-filter issues. @Aca-jov if you think it is not the right way or other pop ups have problems please go ahead with your fix.

amagdas:
:+1:

PR #1281, opened 2015-10-21
[SD-3427] - Default values for urgency and priorty

akintolga:


sivakuna-aap:
:+1: 
Quick Question: Should we override `urgency and priority` always on article creation? Shouldn't it have same values what were given in the template if the article is being created from template?

akintolga:
Thanks @sivakuna-aap. Now it uses the default values if they're empty.

server/apps/archive/archive.py
```diff
@@ -165,6 +167,12 @@ def on_create(self, docs):
             if not doc.get('ingest_provider'):
                 doc['source'] = DEFAULT_SOURCE_VALUE_FOR_MANUAL_ARTICLES
 
+            if 'urgency' not in doc:
+                doc['urgency'] = DEFAULT_URGENCY_VALUE_FOR_MANUAL_ARTICLES
+
+            if 'priority' not in doc:
+                doc['priority'] = DEFAULT_PRIORITY_VALUE_FOR_MANUAL_ARTICLES
```
plamut: Tip: the same can be achieved in a more concise way by using the `dict`'s [setdefault()](https://docs.python.org/3.4/library/stdtypes.html#dict.setdefault) method. Example:
```py
doc.setdefault('priority', DEFAULT_PRIORITY_VALUE_FOR_MANUAL_ARTICLES)
```

plamut:
Apart from a code reafactoring tip that will make you look more sophisticated (:laughing:) it looks good.

:+1:

mdhaman:
:+1:

amagdas:
:+1:

PR #1284, opened 2015-10-22
(fix) Fixing the bug for desk update.

mdhaman:


akintolga:
:+1:

sivakuna-aap:
:+1:

server/apps/desks.py
```diff
@@ -120,7 +120,7 @@ def on_created(self, docs):
             push_notification(self.notification_key, created=1, desk_id=str(doc.get(config.ID_FIELD)))
 
     def on_update(self, updates, original):
-        if updates.get('desk_type', '') != original.get('desk_type', ''):
+        if updates.get('desk_type') and updates.get('desk_type', '') != original.get('desk_type', ''):
```
plamut: The default default value in `updates.get('desk_type', '')` is redundant and might be slightly misleading when comprehending the expression IMO.

In plain English the logic would be described as _"if desk type exists in updates and is non-empty AND is different from the original desk type then..."_, is that correct? The default value in the mentioned term kind of questions this interpretation, as it implies that `desk_type` in the updates might still not exist when comparing it to the original desk_type.

Or maybe it's just me...

plamut:
Looks good, I just dropped a comment where I had to read the code several times before understanding the logic and verifying that it indeed does what is (supposedly) intended. :)

:+1:

server/apps/desks.py
mdhaman: Thanks for the comment. Will make change so that it is easier to understand.

PR #1285, opened 2015-10-22
[SD-3426] Order the results by name alphabetically:

sivakuna-aap:
[SD-3426](https://dev.sourcefabric.org/browse/SD-3426): Order the results by name alphabetically
Requires https://github.com/superdesk/superdesk-core/pull/49

### DIFF
--- a/client/app/scripts/superdesk-content-filters/content-filters.js
+++ b/client/app/scripts/superdesk-content-filters/content-filters.js
@@ -69,7 +69,11 @@
         this.getGlobalContentFilters = function() {
             return api.query('content_filters', {'is_global': true})
                 .then(angular.bind(this, function(params) {
-                    return params._items;
+                    var filters = _.sortBy(params._items, function(filter) {
+                        return filter.name.toLowerCase();
+                    });
+
+                    return filters;
                 }));
         };
 
@@ -229,8 +233,11 @@
         };
 
         var fetchFilterConditions = function() {
-            contentFilters.getAllFilterConditions().then(function(f) {
-                $scope.filterConditions = f;
+            contentFilters.getAllFilterConditions().then(function(_filterConditions) {
+                _filterConditions = _.sortBy(_filterConditions, function(_filterCondition) {
+                    return _filterCondition.name.toLowerCase();
+                });
+                $scope.filterConditions = _filterConditions;
             });
         };
 
@@ -415,9 +422,13 @@
         };
 
         var fetchFilterConditions = function() {
-            contentFilters.getAllFilterConditions().then(function(_filters) {
-                $scope.filterConditions = _filters;
-                _.each(_filters, function(filter) {
+            contentFilters.getAllFilterConditions().then(function(_filterConditions) {
+                _filterConditions = _.sortBy(_filterConditions, function(_filterCondition) {
+                    return _filterCondition.name.toLowerCase();
+                });
+
+                $scope.filterConditions = _filterConditions;
+                _.each(_filterConditions, function(filter) {
                     $scope.filterConditionLookup[filter._id] = filter;
                 });
             });
@@ -425,7 +436,10 @@
 
         var fetchContentFilters = function() {
             contentFilters.getAllContentFilters().then(function(_filters) {
-                $scope.contentFilters = _filters;
+                $scope.contentFilters = _.sortBy(_filters, function(filter) {
+                    return filter.name.toLowerCase();
+                });
+
                 _.each($scope.contentFilters, function(filter) {
                     $scope.contentFiltersLookup[filter._id] = filter;
                 });
--- a/client/app/scripts/superdesk-content-filters/content-filters.js
+++ b/client/app/scripts/superdesk-content-filters/content-filters.js
@@ -20,8 +20,8 @@
      *   providing higher-level methods for fetching and modifying all content
      *   related to content filters on the server.
      */
-    ContentFiltersService.$inject = ['api'];
-    function ContentFiltersService(api) {
+    ContentFiltersService.$inject = ['api', '$filter'];
+    function ContentFiltersService(api, $filter) {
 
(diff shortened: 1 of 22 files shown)


mugurrus:
:+1: (some tests would be needed)

client/app/scripts/superdesk-content-filters/content-filters.js
```diff
@@ -69,7 +69,11 @@
         this.getGlobalContentFilters = function() {
             return api.query('content_filters', {'is_global': true})
                 .then(angular.bind(this, function(params) {
```
plamut: `angular.bind()` can be removed here, the bound function does not make any use of `this` in its body.

client/app/scripts/superdesk-content-filters/content-filters.js
```diff
@@ -425,7 +436,10 @@
 
         var fetchContentFilters = function() {
             contentFilters.getAllContentFilters().then(function(_filters) {
-                $scope.contentFilters = _filters;
+                $scope.contentFilters = _.sortBy(_filters, function(filter) {
+                    return filter.name.toLowerCase();
+                });
```
plamut: This comparison function with identical logic is used at multiple places, IMO it should be extracted.
```js
function lowercaseName (item) {
    return item.name.toLowerCase();
}

$scope.something = _.sortBy(collection, lowercaseName);
...
$scope.somethingElse = _.sortBy(anotherCollection, lowercaseName);
...
```

client/app/scripts/superdesk-desks/desks.js
```diff
@@ -871,6 +885,9 @@
                                 _.extend(origDesk, scope.desk.edit);
                             }
 
+                            scope.desks._items = _.sortBy(scope.desks._items, function(desk) {
+                                return desk.name.toLowerCase();
+                            });
```
plamut: Similar suggestion here, the same comparison function is repeated multiple times.

client/app/scripts/superdesk-desks/views/desk-config-modal.html
```diff
@@ -180,7 +180,7 @@ <h3 ng-show="desk.edit._id" translate>Edit "{{desk.edit.name}}" Desk</h3>
                         <p ng-if="message">{{ message }}</p>
                         <div class="member-list clearfix">
                             <ul>
-                                <li ng-repeat="user in deskMembers" ng-class="{'inactive' : !user.is_enabled || !user.is_active}">
+                                <li ng-repeat="user in deskMembers | orderBy: 'display_name'" ng-class="{'inactive' : !user.is_enabled || !user.is_active}">
                                     <i class="icon-close-small" ng-click="remove(user)"></i>
                                     <figure class="avatar"><img sd-user-avatar data-src="user.picture_url"></figure>
                                     <div class="name">{{user.display_name}}</div>
```
plamut: We can probably use one-time binding here?

client/app/scripts/superdesk-ingest/module.js
```diff
@@ -762,6 +782,9 @@ define([
                         if (_new) {
                             scope.schemes.push(_orig);
                         }
+                        scope.schemes = _.sortBy(scope.schemes, function(routingScheme) {
+                            return routingScheme.name.toLowerCase();
+                        });
```
plamut: Repeated comparison function... :)

client/app/scripts/superdesk-content-filters/content-filters.js
sivakuna-aap: Even I thought the same. Since it depends on the resource properties which are controlled by server I thought extracting it to common function may not be relevant.

sivakuna-aap:
@mugurrus you mean tests for sorting?

mugurrus:
@sivakuna-aap yes

client/app/scripts/superdesk-templates/templates.js
```diff
@@ -180,12 +183,14 @@
...
+                        var d = new Date();
 
-                    d.setUTCHours(time.substr(0, 2));
-                    d.setUTCMinutes(time.substr(2, 2));
+                        d.setUTCHours(time.substr(0, 2));
+                        d.setUTCMinutes(time.substr(2, 2));
 
-                    return d;
+                        return d;
+                    }
```
plamut: Just checking - is it OK to rely on the user's machine time and time zone here? Could a misconfigured value cause any problems?

sivakuna-aap: As of it's not relevant here. We can deal with Timezone across superdesk in a separate ticket. There's plan for this.

sivakuna-aap:
@mugurrus I think it's not relevant. We are not writing tests for such functions. If you think it's necessary then we have to hard code the data.

@superdesk/developers what do you guys think?

mugurrus:
@sivakuna-aap I agree it's pretty trivial but I don't understand what you mean by "hard code the data". You create the tests data.

plamut:
Looks good in general, although please check if the test coverage is sufficient (IMO tests should be added). I'm just not sure what you mean by _"we have to hard code the data"_?

Re: refactoring comparison functions - IMO it still makes sense to refactor them, the resource properties are generally stable and do not change very often - especially the `name` which is very standard. :)

client/app/scripts/superdesk-templates/templates.js
plamut: Ah OK, then of course don't bother with it in this PR. :)

sivakuna-aap:
@mugurrus I was referring the tests data by saying hardcoding the data. But yes you were right, we create tests data. 

@plamut Agreed for the refactoring.

petrjasek:
imo it would make sense to write unittest for that filter

sivakuna-aap:
Added a unit test for the filter.

petrjasek:
:+1:

PR #1292, opened 2015-10-23
[SD-3386] Fixup datelines using macros

marwoodandrew:
The problem is that the datelines we receive from some of the ingest providers/parsers do not conform to what we want. 

I've implemented a couple of macros to try to sort them out, the macros can be called as part of routing or maybe when we re-factor the ingest it might be an idea to nominate a macro to call on ingest.

We can't implement this stuff in the parsers as the same parser may be used for different providers.

### DIFF
--- a/server/macros/dpa_derive_dateline.py
+++ b/server/macros/dpa_derive_dateline.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.locators.locators import find_cities
+
+
+def dpa_derive_dateline(item, **kwargs):
+    """
+    DPA content is recieved in IPTC7901 format, this macro attempts to parse a dateline from the first few lines of
+    the item body and populate the dataline location, it also populates the dateline source.
+    If a dateline is matched the coresponding string is removed from the article text.
+    :param item:
+    :param kwargs:
+    :return:
+    """
+    lines = item['body_html'].splitlines()
+    if lines:
+        # expect the dateline in the first 5 lines, sometimes there is what appears to be a headline preceeding it.
+        for line_num in range(0, min(len(lines), 5)):
+            city, source, the_rest = lines[line_num].partition(' (dpa) - ')
+            # test if we found a candidate and ensure that the city starts the line and is not crazy long
+            if source and lines[line_num].find(city) == 0 and len(city) < 20:
+                cities = find_cities()
+                located = [c for c in cities if c['city'].lower() == city.lower()]
+                if 'dateline' not in item:
+                    item['dateline'] = {}
+                item['dateline']['located'] = located[0] if len(located) > 0 else {'city_code': city, 'city': city,
+
+                                                                                   'tz': 'UTC', 'dateline': 'city'}
+                item['dateline']['source'] = 'dpa'
+                item['dateline']['text'] = city
+                lines[line_num] = lines[line_num].replace(city + source, '')
+                item['body_html'] = '\r\n'.join(lines)
+                break
+    return item
+
+
+name = 'Derive dateline from article text for DPA'
+label = 'DPA dateline derivation'
+shortcut = '$'
+callback = dpa_derive_dateline
--- a/server/macros/dpa_derive_dateline_test.py
+++ b/server/macros/dpa_derive_dateline_test.py
@@ -38,4 +38,4 @@ class DPADeriveDatelineTests(TestCase):
         item = dict()
         item['body_html'] = 'No dateline here'
         dpa_derive_dateline(item)
-        self.assertNotIn('located', item)
+        self.assertNotIn('dateline', item)
--- a/server/macros/dpa_derive_dateline_test.py
+++ b/server/macros/dpa_derive_dateline_test.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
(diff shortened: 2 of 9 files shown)


server/macros/dpa_derive_dateline.py
```diff
@@ -0,0 +1,48 @@
...
+        # expect the dateline in the first 5 lines, sometimes there is what appears to be a headline preceeding it.
+        for line_num in range(0, min(len(lines), 5)):
+            city, source, the_rest = lines[line_num].partition(' (dpa) - ')
+            # test if we found a candidate and ensure that the city starts the line and is not crazy long
+            if source and lines[line_num].find(city) == 0 and len(city) < 20:
+                cities = find_cities()
+                located = [c for c in cities if c['city'].lower() == city.lower()]
+                if 'dateline' not in item:
+                    item['dateline'] = {}
+                item['dateline']['located'] = located[0] if len(located) > 0 else {'city_code': city, 'city': city,
```
sivakuna-aap: Don't we get date from DPA?

server/macros/dpa_derive_dateline_test.py
```diff
@@ -0,0 +1,41 @@
...
+        item['body_html'] = 'Dubbo (dpa) - Where on earth is that'
+        dpa_derive_dateline(item)
+        self.assertEqual(item['dateline']['located']['city'], 'Dubbo')
+        self.assertEqual(item['dateline']['located']['country'], 'Australia')
+
+    def no_match_test(self):
+        item = dict()
+        item['body_html'] = 'No dateline here'
+        dpa_derive_dateline(item)
+        self.assertNotIn('located', item)
```
sivakuna-aap: This should be `self.assertNotIn('located', item['dateline'])` as located will always be not in item

server/macros/noise11_derive_metadata.py
```diff
@@ -0,0 +1,58 @@
...
+    appropriately
+    :param item:
+    :param kwargs:
+    :return:
+    """
+    try:
+        if 'anpa_category' not in item:
+            category_map = superdesk.get_resource_service('vocabularies').find_one(req=None, _id='categories')
+            if category_map:
+                map_entry = next((code for code in category_map['items'] if code['qcode'] == 'e' and code['is_active']),
```
sivakuna-aap: Looks like this is been used in multiple places. Can we add a method to `VocabulariesService` which gives a generator for active vocabularies for a given vocabulary id?

server/macros/noise11_derive_metadata.py
```diff
@@ -0,0 +1,58 @@
...
+                                 None)
+                item['anpa_category'] = [{'qcode': 'e', 'name': map_entry['name']}]
+
+        if 'subject' not in item:
+            qcode = '01000000'
+            item['subject'] = [{'qcode': qcode, 'name': subject_codes[qcode]}]
+
+        cities = find_cities()
+        located = [c for c in cities if c['city'].lower() == 'sydney']
```
sivakuna-aap: If noise11 always comes from Australia/Sydney then we can fetch cities within a country using `find_cities(country_code='AU')`

server/macros/reuters_derive_dateline.py
```diff
@@ -0,0 +1,64 @@
...
+    :param items:
+    :return:
+    """
+    try:
+        html = item.get('body_html')
+        if html:
+            soup = BeautifulSoup(html)
+            pars = soup.findAll('p')
+            if len(pars) >= 2:
+                if 'byline' in item and item.get('byline') in pars[0].get_text():
```
sivakuna-aap: can use constant for 'byline'

server/macros/reuters_derive_dateline.py
```diff
@@ -0,0 +1,64 @@
...
+                    first = pars[1].get_text()
+                else:
+                    first = pars[0].get_text()
+                city, source, the_rest = first.partition(' (Reuters) - ')
+                if source:
+                    # sometimes the city is followed by a comma and either a date or a state
+                    city = city.split(',')[0]
+                    if any(char.isdigit() for char in city):
+                        return
+                    cities = find_cities()
```
sivakuna-aap: These 2-3 lines are copied across modules. Can we extract it to a function in ingest module?

server/macros/reuters_derive_dateline.py
```diff
@@ -0,0 +1,64 @@
...
+                    located = [c for c in cities if c['city'].lower() == city.lower()]
+                    # if not dateline we create one
+                    if 'dateline' not in item:
+                        item['dateline'] = {}
+                    # there is already a dateline that is not Bangalore don't do anything just return
+                    elif 'located' in item['dateline'] and 'BANGALOR' != item['dateline']['located'].get(
```
sivakuna-aap: 'e' is missing in 'Bangalore'. Is it done deliberately?

server/macros/reuters_derive_dateline.py
```diff
@@ -0,0 +1,64 @@
...
+                            'city').upper():
+                        return
+
+                    item['dateline']['located'] = located[0] if len(located) > 0 else {'city_code': city,
+                                                                                       'city': city,
+                                                                                       'tz': 'UTC',
+                                                                                       'dateline': 'city'}
+                    item['dateline']['source'] = item.get('original_source', 'Reuters')
+                    item['dateline']['text'] = city
```
sivakuna-aap: I think the text should be equal to what we get from reuters.

server/macros/dpa_derive_dateline.py
marwoodandrew: In the footer we get something like "071341 GMT Okt 15" [sic] I assume that's 15th October, will need to check the spec for the definition.

marwoodandrew: This is defined as the transmission date and time.

server/macros/reuters_derive_dateline.py
marwoodandrew: If Reuters send a "dateline" it seems to actually be a  date string something like "2015-10-26 02:45:12 GMT+00:00", It is not always present, it seems to depend on the channel.

server/macros/noise11_derive_metadata.py
marwoodandrew: Might do it in another PR as it will be a core change!

amagdas:
:+1:

amagdas:
@superdesk/developers Ping

akintolga:
:+1:

PR #1293, opened 2015-10-23
feat(highlights): Add the option to define templates for highlights

ioanpocol:


client/app/scripts/superdesk-highlights/highlights.js
```diff
@@ -291,12 +291,17 @@
...
-    function HighlightsSettingsController($scope, desks) {
+    HighlightsSettingsController.$inject = ['$scope', 'api', 'desks'];
+    function HighlightsSettingsController($scope, api, desks) {
         desks.initialize().then(function() {
             $scope.desks = desks.deskLookup;
         });
 
+        api.query('content_templates', {'template_type': 'highlights'}).then(function(result) {
+            $scope.templates = result._items || [];
+            console.log('read templates: ', result);
```
sivakuna-aap: We can remove `console.log()`

server/apps/highlights/generate.py
```diff
@@ -48,15 +44,23 @@ def create(self, docs, **kwargs):
...
                         item = service.find_one(req=None, _id=ref.get('residRef'))
-                        body.append(item_to_str(item))
-                        body.append('<p></p>')
-            doc['body_html'] = join_html(body)
-
+                        if item:
+                            html = item.get('body_html')
+                            if html:
+                                soup = BeautifulSoup(html)
+                                item['body_html'] = str(soup.p)
```
petrjasek: ok now I see it's here. maybe it would be better to put there `first_paragraph` and also the body itself so you can do something else with those templates

amagdas:
LGTM :+1:

mugurrus:
:+1:

PR #1295, opened 2015-10-23
fix(authoring): SD-3486 Removed minimize button for locked items

pavlovicnemanja:
SD-3486


sivakuna-aap:
:+1: 

Looks good. Although the fix is different from the description mentioned in the ticket it's the right fix. Please change the description of the ticket to avoid reopen from QA as **For all locked items the minimize button will be invisible irrespective of their state**. 

Can you please also look related tickets, I think QA opened couple of tickets around minimize button?

mdhaman:
:+1:

amagdas:
:+1:

PR #1296, opened 2015-10-23
fix(authoring): Changed the way widget is assigned to number

pavlovicnemanja:
Previous case: Widget is assigned to number by number in order field. Problem here is that in some cases same number is assigned for two widgets.

New case: Widget is assigned by index number from forEach loop. Also, widgets are first sorted by order.

Improvement for task: SD-3414


mdhaman:
:+1:  Looks Good. Any test for these changes?

pavlovicnemanja:
Test for this where added in previous PR, where I implemented this... This is just improvement for it.

sivakuna-aap:
:+1:

PR #1299, opened 2015-10-26
(fix) [SD-3390] When cursor leaves the Dateline field it changes the value

sivakuna-aap:
Resolves [SD-3390](https://dev.sourcefabric.org/browse/SD-3390) When cursor leaves the Dateline field it changes the value


akintolga:
:+1:

sivakuna-aap:
ping @superdesk/developers

marwoodandrew:
:+1:

PR #1305, opened 2015-10-27
templates module renamed to superdesk.templates-cache

vied12:
`templates` wasn't safe enough to prevent conflict with other dependencies


ioanpocol:
:+1:

petrjasek:
just wondering if that file `scripts/templates-cache.js` should be there at all, this is generated via grunt task right?

vied12:
Yes it is generated by a grunt task. You mean that you prefer to gitignore it ?

petrjasek:
yes if possible. it should be also part of bower folder so i don't think we need this one

vied12:
Ok but then we need to reconnect the task that generate this file when templates are modified,
because this file is imported in `index.html`

petrjasek:
I see it's not part of `grunt server` task. so then it seems like a bigger task and lets not do it as part of this PR

vied12:
:ok_hand:

Le mar. 27 oct. 2015 à 17:38, Petr Jašek <notifications@github.com> a
écrit :

amagdas:
:+1:

ioanpocol:
Please rebase

vied12:
https://github.com/superdesk/superdesk/pull/1379 includes these changes

PR #1308, opened 2015-10-28
[SD-3425] - User save fails when change is on different tabs

akintolga:


client/app/scripts/superdesk-users/users.js
```diff
@@ -1298,22 +1313,26 @@
             };
         }])
 
-        /**
+		/**
          * @memberof superdesk.users
          * @ngdoc directive
-         * @name sdUserPrivileges
+         * @name sdUserPreferences
```
sivakuna-aap: This should be privileges only not preferences.

client/app/scripts/superdesk-users/users.js
```diff
@@ -1080,7 +1090,11 @@
                         preSaveCategoriesCheck()
                         .then(function () {
                             var update = createPatchObject();
-                            return preferencesService.update(update);
+                            return preferencesService.update(update).then(function() {
+                                userList.getUser(scope.user._id, true).then(function(u) {
+                                    scope.user = u;
+                                });
+                            });
```
plamut: Would it make sense to flatten this promise chain here?

plamut:
Looks good IMO. :+1:

mdhaman:
:+1:

PR #1311, opened 2015-10-28
(feat) [SD-2616] Create broadcast story.

mdhaman:
Elasticsearch indexes needs to be rebuild.


server/apps/archive_broadcast/broadcast.py
```diff
@@ -0,0 +1,142 @@
...
+                }
+            }
+        }
+
+        request = ParsedRequest()
+        request.args = {'source': json.dumps(query)}
+        broadcast_items = get_resource_service(SOURCE).get(req=request, lookup=None)
+
+        for broadcast_item in broadcast_items:
+            item = next((item for item in items if item.get(config.ID_FIELD) == broadcast_item.get('family_id')))
```
plamut: Just checking - is it possible that the generator expression does not yield any results here? This would result in an error, because there is no default value argument passed to the `next()` function.

plamut:
Looks good in general IMO. :+1:

server/apps/archive_broadcast/broadcast.py
```diff
@@ -0,0 +1,293 @@
...
+            desk = get_resource_service('desks').find_one(req=None, _id=desk_id)
+
+        doc.pop('desk', None)
+        doc['task'] = {}
+        if desk:
+            doc['task']['desk'] = desk.get(config.ID_FIELD)
+            doc['task']['stage'] = desk.get('incoming_stage')
+
+        doc['task']['user'] = get_user().get('_id')
+        genre_list = get_resource_service('vocabularies').find_one(req=None, _id='genre') or {}
```
sivakuna-aap: The `VocabulariesService.find_one()` gives inactive vocabularies also. So, Inactive Vocabularies also needs to be filtered.

sivakuna-aap:
:+1: Except the minor one. Needs rebase.

marwoodandrew:
:+1:

PR #1312, opened 2015-10-29
[SD-1960] Search by creator and byline

marwoodandrew:


sivakuna-aap:
:+1: 
**Minor suggestion:** the `selectDesk()` used for selecting user in e2e tests misleads the reader. Can we have different function?

akintolga:
:+1:

marwoodandrew:
Will add another function and fix the lint errors!

client/app/scripts/superdesk-search/search.js
```diff
@@ -421,6 +421,23 @@
                         }
                     }
                 });
+                /*
+                * if one of the parameters is the original creator then replace the user id with the
+                * display name.
+                */
+                var i = -1;
+                for (i in tags.selectedParameters) {
```
plamut: Using the `for..in` construct to loop over an array might result in unexpected behavior, thus I suggest to rewrite this to e.g. use the Array's `forEach()` method or some other looping construct.

**Explanation:** Let's say some module adds a new enumerable property to the Array's prototype, e.g. a convenient utility method. The `for..in` construct would include such property, too!

```js
>>> Array.prototype.foo = 'bar';
>>> var a = [1,2,3];
>>> for (var i in a) {
...     console.log(i);
... }
0
1
2
foo  # <--- oops!
```

client/app/scripts/superdesk-search/search.js
```diff
@@ -421,6 +421,23 @@
...
+                    if (tags.selectedParameters[i].indexOf('original_creator') >= 0) {
+                        break;
+                    }
+                }
```
plamut: ... actually, if the intent is to find the first matching _selected parameter_, it's probably better to use an existing function than replicating its logic.

```js
var matchingParam = _.find(tags.selectedParameters, function (param) {
    return param.indexOf('original_creator') >= 0;
});

if (matchingParam) {
    var user_id = matchingParam.split(':')[1];
    // etc.
}
```

client/app/scripts/superdesk-search/search.js
```diff
@@ -421,6 +421,23 @@
...
+                if (i >= 0) {
```
plamut: ... actually, if the intent is to find the first matching _selected parameter_, it's probably better to use an existing function than replicating its logic.

```js
var matchingParam = _.find(tags.selectedParameters, function (param) {
    return param.indexOf('original_creator') >= 0;
});

if (matchingParam) {
    var user_id = matchingParam.split(':')[1];
    // etc.
}
```

client/app/scripts/superdesk-search/search.js
```diff
@@ -680,6 +697,21 @@
 
                                 metadata.removeSubjectTerm(elementName);
                             }
+                            /*
+                            * remove the original creator parameter
+                            */
+                            if (param.indexOf('creator:') >= 0) {
+                                var parray = params.q.split(' ');
+                                for (var p in parray)
```
plamut: Similar remark to above (using `for..in` to loop over an array).

client/app/scripts/superdesk-search/views/item-search.html
```diff
@@ -72,14 +72,20 @@
...
+                {{:: 'Byline' | translate}}
+            </label>
+            <input type="text" id="search-byline" ng-model="meta.byline">
+        </div>
         <div class="field" ng-if="repo.archive">
             <label for="search-creator">
                 {{:: 'Creator' | translate}}
             </label>
-            <input type="text" id="search-creator" ng-model="meta.original_creator">
+            <select id="search-creator" ng-model="meta.original_creator" ng-options="user._id  as user.display_name for user in userList track by user._id">
```
plamut: When using ng-options, using both the `select as` and `track by` parts in the expression might result in unexpected behavior as explained in the [ng-options documentation]( https://docs.angularjs.org/api/ng/directive/ngOptions). It's thus probably best to remove the `track by` part in order to avoid potentially introducing a bug if the expression gets changed in the future.

plamut:
IMO this still needs a bit of polishing and should not be merged just yet. :)

PR #1314, opened 2015-10-29
feat(authoring): SD-3471 Added user avatar in top header of locked items

pavlovicnemanja:
SD-3471

### DIFF
--- a/client/app/scripts/superdesk-authoring/views/authoring-topbar.html
+++ b/client/app/scripts/superdesk-authoring/views/authoring-topbar.html
@@ -1,16 +1,38 @@
 <div id="subnav" class="subnav" ng-class="{'auth-sendto': views.send}">
+
+    <div class="dropdown dropright lock-avatar pull-left" ng-if="item._locked" dropdown>
+        <figure class="avatar small avatar-with-info" dropdown-toggle>
+            <div class="logged-info"></div>
+            <div class="user-image">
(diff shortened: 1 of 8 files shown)


client/app/scripts/superdesk-authoring/views/authoring-topbar.html
```diff
@@ -1,16 +1,38 @@
...
+            <div class="user-image">
+                <img sd-user-avatar data-src="item.lock_user.picture_url" data-initials="item.lock_user | username">
+            </div>
+        </figure>
+
+        <div class="dropdown-menu">
+            <div class="menu-label" translate>Locked by</div>
+            <figure class="avatar large">
+                <div class="user-image">
+                    <img sd-user-avatar data-src="item.lock_user.picture_url" data-initials="item.lock_user | username">
```
plamut: The `` tag should be closed, right?

pavlovicnemanja: By W3, img tag doesn't need to be closed at the end: http://www.w3schools.com/tags/tag_img.asp

plamut:
IMO a test should be added for changes in the `sdUserAvatar` directive, but looks good to me otherwise.

client/app/scripts/superdesk-authoring/views/authoring-topbar.html
plamut: Aha, I've re-checked and you're right, this is not the case anymore in HTML5.

ioanpocol:
:+1:

PR #1316, opened 2015-10-30
(fix) [SD-3513] Kill operation aborts as publish validators throw exception for pubstatus

sivakuna-aap:


mdhaman:
:+1: Maybe behave test can be modified to verify the pubstatus.

amagdas:
:+1:

PR #1319, opened 2015-10-30
feat(templates): Added profile field to templates and displaying prof…

eozan:
…ile in list and authoring.

SDPA-8


mugurrus:
:+1:

client/app/scripts/superdesk-archive/views/media-box-list.html
```diff
@@ -27,6 +27,7 @@
             <time sd-datetime data-date="item.versioncreated"></time>
         </div>
         <div class="line">
+            <div class="label label--{{item.profile}}" ng-if="item.profile">{{ item.profile }}</div>
```
petrjasek: just wondering - why double dash in class name?

eozan: It's what Vladimir told me to use, might be a convention maybe.

petrjasek: ok

PR #1326, opened 2015-11-03
feat(search ingest): Integrate PA Images search ingest

ioanpocol:


akintolga:
:+1:

mdhaman:
:+1: LGTM

server/apps/io/search_ingest.py
```diff
@@ -14,24 +14,37 @@
...
+from superdesk.metadata.utils import generate_guid
+from superdesk.resource import Resource
+
 
 logger = logging.getLogger(__name__)
 
 
-class AapMMService(superdesk.Service):
+class SearchIngestResource(superdesk.Resource):
+    resource_methods = ['GET', 'POST']
```
sivakuna-aap: Why we need POST for search?

ioanpocol: We need post in order to fetch an item from search result.

server/apps/pa_img/pa_img_datalayer.py
```diff
@@ -0,0 +1,265 @@
...
+    if result:
+        params['text'] = result.group(1)
+    else:
+        result = re.search('[^:]+\(([\w\s]+)\)', query)
+        if result:
+            params['text'] = result.group(1)
+    return params
+
+
+class PaImgDatalayer(DataLayer):
```
petrjasek: I'm not sure it should extend datalayer, you can't replace other data layers with this one, we should probably find some different base class for this one, maybe something like *ContentProvider* ? Or *MediaProvider* ?

ioanpocol: The part that is common to 'search ingest' providers is on SearchIngestService that is used now for both aap_mm and pa_img. I think for now to extend the DataLayer to implement the custom access to media storage service should be ok. It can be refactored easy later if it will be required.

petrjasek: yep for now it's ok

server/apps/io/search_ingest.py
```diff
@@ -66,12 +79,12 @@ def create(self, docs, **kwargs):
         return new_guids
 
     def get(self, req, lookup):
-        provider = get_resource_service('ingest_providers').find_one(source='aapmm', req=None)
+        provider = get_resource_service('ingest_providers').find_one(source=self.source, req=None)
         if provider:
             if 'config' in provider and 'username' in provider['config']:
                 self.backend.set_credentials(provider['config']['username'], provider['config']['password'])
```
petrjasek: so here you get an error if `password` is empty, same issue is on line 49

ioanpocol:
l cosed this pr because I create a new(PR#1338) one against pa-poc branch instead master branch.

PR #1327, opened 2015-11-03
(fix) Reuters ingest routing

marwoodandrew:
Requires superdesk/superdesk-core#65
It seems that the dateRange request to Reuters may return duplicated guids

### DIFF
--- a/server/apps/io/reuters.py
+++ b/server/apps/io/reuters.py
@@ -13,7 +13,7 @@
 import traceback
 import datetime
 from urllib.parse import urlparse, urlunparse
-
+from superdesk.logging import logger
 import requests
 
 from superdesk.io.ingest_service import IngestService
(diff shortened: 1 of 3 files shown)


sivakuna-aap:
Needs rebasing

server/apps/io/reuters.py
```diff
@@ -102,9 +102,12 @@ def get_ids(self, channel, last_updated, updated):
         ids = []
         payload = {'channel': channel, 'fieldsRef': 'id'}
         payload['dateRange'] = "%s-%s" % (self.format_date(last_updated), self.format_date(updated))
+        logger.info('Reuters requesting Date Range |{}| for channel {}'.format(payload['dateRange'], channel))
         tree = self.get_tree('items', payload)
         for result in tree.findall('result'):
-            ids.append(result.find('guid').text)
+            id = result.find('guid').text
```
plamut: I suggest picking a different variable name to avoid any confusion with the built-in `id()` function.

server/apps/io/reuters.py
```diff
@@ -102,9 +102,12 @@ def get_ids(self, channel, last_updated, updated):
...
+            if id not in ids:
+                ids.append(id)
```
plamut: **Tip:** We can get rid of this containment check if we use `set()` for `ids` instead of a `list` - we get duplicate detection on new element insertion for free. It's minor though, the code is already readable as is, and its `O(n^2)` complexity probably doesn't hurt here, the expected length of the `ids` list is relatively small, right?

plamut:
Apart from a few minor suggestions it looks good to me.

:+1:

server/apps/io/reuters.py
marwoodandrew: The list can be up to 300 items (not unique) or maybe 220 unique items on initial start-up when it attempts to request 24Hrs of content.  Usually when polling it would only be a couple. I guess 300 is a significant number so I'll change it to use a set, it will simplify the code!

mugurrus:
:+1:

petrjasek:
:+1:

PR #1328, opened 2015-11-03
fix(templates): SD-3587 Impossible to save a new or edited template

ioanpocol:


sivakuna-aap:
:+1:

petrjasek:
imo would be good to have some test for this

ioanpocol:
We need a e2e some tests on template add&edit. But for now I will skip the tests because there is no e2e test on templates and the next sprint the add/edit template will be reimplemented.

petrjasek:
i ment unittest for deleting of that property if it prevents template from saving..

ioanpocol:
I changed one of existing unittets

PR #1329, opened 2015-11-03
fix(saved search): Set a larger limit(200) for saved search list

ioanpocol:


sivakuna-aap:
I think 200 was ok but 2000 is more than the required. Is there a real need/requirement for this?

ioanpocol:
There is this issue on saved_search that some saved search are not showed. Seams that currently on systems used for preview there are a lot of saved_searches and I'm not sure if 200 is enough.

sivakuna-aap:
:+1:

akintolga:
:+1:

PR #1330, opened 2015-11-03
[SD-3121] - Highlight in monitoring and authoring is consistent

Aca-jov:
SD-3121, SD-3315, SD-3420

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1308,6 +1308,14 @@
                     }
                 });
 
+                $scope.$on('item:mark', function(e, data) {
+                    if ($scope.item._id === data.item_id){
+                        $scope.item.highlights = $scope.item.highlights.indexOf(data.highlight_id) >= 0 ?
+                            _.without($scope.item.highlights, data.highlight_id) :
+                            [data.highlight_id].concat($scope.item.highlights);
+                    }
+                });
+
                 macros.setupShortcuts($scope);
             }
         };
@@ -2015,9 +2023,9 @@
                 });
 
                 metadata.initialize().then(function() {
-                    scope.item.hasCrops = false;
                     scope.metadata = metadata.values;
-                    if (scope.metadata.crop_sizes) {
+                    if (scope.item.type === 'picture'){
+                        scope.item.hasCrops = false;
                         scope.item.hasCrops = scope.metadata.crop_sizes.some(function (crop) {
                             return scope.item.renditions[crop.name];
                         });
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -1308,11 +1308,15 @@
                     }
                 });
 
-                $scope.$on('item:mark', function(e, data) {
+                $scope.$on('item:highlight', function(e, data) {
                     if ($scope.item._id === data.item_id){
-                        $scope.item.highlights = $scope.item.highlights.indexOf(data.highlight_id) >= 0 ?
-                            _.without($scope.item.highlights, data.highlight_id) :
-                            [data.highlight_id].concat($scope.item.highlights);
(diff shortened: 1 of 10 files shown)


pavlovicnemanja:
:+1:

client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -274,6 +274,7 @@
                 scope.$on('ingest:update', queryItems);
                 scope.$on('item:spike', queryItems);
                 scope.$on('item:unspike', queryItems);
+                scope.$on('item:mark', queryItems);
```
petrjasek: it would be better to handle it in list item directive, without reloading the list every time item is marked for highlight

server/apps/highlights/service.py
```diff
@@ -72,17 +78,27 @@ def create(self, docs, **kwargs):
...
             updates = {
                 'highlights': highlights,
                 '_updated': item['_updated'],
                 '_etag': item['_etag']
             }
             service.update(item['_id'], updates, item)
-        push_notification('item:mark', marked=1)
+
+            push_notification(
+                'item:mark',
```
sivakuna-aap: I think a better name for the event is required.

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -2015,11 +2023,13 @@
                 });
 
                 metadata.initialize().then(function() {
-                    scope.item.hasCrops = false;
                     scope.metadata = metadata.values;
-                    scope.item.hasCrops = scope.metadata.crop_sizes.some(function (crop) {
-                        return scope.item.renditions[crop.name];
-                    });
+                    if (scope.item.type === 'picture'){
+                        scope.item.hasCrops = false;
```
sivakuna-aap: If some piece of code is dependent on `hasCrops` then it might fail. I think it's better to have `scope.itam.hasCrops = false;` outside if.

server/apps/highlights/service.py
plamut: I agree. I kept the existing name, though `"mark"` is indeed kind of too generic a label.

client/app/scripts/superdesk-authoring/authoring.js
Aca-jov: I added this flag, just for images, to check if images are cropped, to know to change label on button, so this flag is not needed outside if, because items of different type cannot have crops, only images can have it.

sivakuna-aap: I totally understand it. But this PR isn't addressing the code which is already using `hasCrops`. That's the reason I added the comment.

server/apps/highlights/service.py
petrjasek: much better

plamut: @Aca-jov Can you make the corresponding change(s) yourself? It's just replacing the relevant strings in (both in Python and Javascript), no other Python logic is needed.

Aca-jov: Yes, I will make changes.

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1308,6 +1308,18 @@
                     }
                 });
 
+                $scope.$on('item:highlight', function(e, data) {
+                    if ($scope.item._id === data.item_id){
```
mugurrus: DRY - this function could be extracted

server/apps/highlights/service.py
```diff
@@ -72,17 +78,27 @@ def create(self, docs, **kwargs):
             highlights = item.get('highlights', [])
             if not highlights:
```
mugurrus: This if should be removed, it's redundant

plamut: Actually it isn't, `highlights` can also exist in the item dictionary and be equal to `None`, hence the need for the additional check.

mugurrus:
:+1:

PR #1331, opened 2015-11-03
feat(associations): implement featured image field

petrjasek:
used for text items only, you can drop there pictures from any list

### DIFF
--- a/client/app/scripts/superdesk-archive/directives.js
+++ b/client/app/scripts/superdesk-archive/directives.js
@@ -396,6 +396,21 @@
                 }
             };
         }])
+        .directive('sdDraggableItem', function() {
+            return {
+                link: function(scope, elem) {
+                    if (scope.item) {
+                        elem.attr('draggable', true);
+                        // set item data on event
+                        elem.on('dragstart', function(event) {
+                            var dt = event.originalEvent.dataTransfer;
+                            dt.setData('application/superdesk.item.' + scope.item.type, angular.toJson(scope.item));
+                            dt.effectAllowed = 'link';
+                        });
+                    }
+                }
+            };
+        })
         .directive('sdItemCrops', ['metadata', function(metadata) {
             return {
                 templateUrl: 'scripts/superdesk-archive/views/item-crops.html',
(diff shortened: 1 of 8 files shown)


client/app/scripts/superdesk-archive/directives.js
```diff
@@ -396,6 +396,21 @@
                 }
             };
         }])
+        .directive('sdDraggableItem', function() {
+            return {
+                link: function(scope, elem) {
+                    if (scope.item) {
```
plamut: It's somewhat unclear where `item` comes from, could be harder to track it in the scope hierarchy when debugging - would it make sense it to the directive explicitly so that it is immediately clear what `item` is by just looking at the HTML element the directive is used on?

server/apps/publish/content/tests.py
```diff
@@ -718,3 +719,20 @@ def test_added_removed_in_a_package(self):
...
+    def test_resolve_associations(self):
+      item = {
+        'associations': {
+          'sidebar': {'uri': '1'}
+        }
+      }
+      ArchivePublishService()._resolve_associations(item)
+      self.assertEqual('text', item['associations']['sidebar']['type'])
+
+      with self.assertRaises(ValidationError):
```
plamut: Do we want to check for the correct error message as well, or is `_resolve_associations()` simply raising `ValidationError` already enough for our needs?

plamut:
It would probably make sense to add an e2e test, too, to test the directives themselves - if it's feasible?

Looks good though. :+1:

client/app/scripts/superdesk-archive/directives.js
petrjasek: it's used on element with `ng-repeat="item in ..."` so not sure it can have isolated scope

server/apps/publish/content/tests.py
petrjasek: right, it could check the error message

client/app/scripts/superdesk-archive/directives.js
plamut: Ah, if that's the case then it's probably not needed, since it's clear what `item` is in the template.

PR #1335, opened 2015-11-04
[SD-2201] No filter matches all articles

marwoodandrew:
Change the action of not having a filter to pass content.

### DIFF
--- a/server/apps/content_filters/content_filter_tests.py
+++ b/server/apps/content_filters/content_filter_tests.py
@@ -310,7 +310,7 @@ class RetrievingDataTests(ContentFilterTests):
             self.assertEqual(1, docs.count())
             self.assertTrue('3' in doc_ids)
 
-    def test_does_match_returns_false_for_nonexisting_filter(self):
+    def test_does_match_returns_true_for_nonexisting_filter(self):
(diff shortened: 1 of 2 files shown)


sivakuna-aap:
:+1:

server/apps/content_filters/content_filter_tests.py
```diff
@@ -312,7 +312,7 @@ def test_build_elastic_query_using_like_filter_multi_content_filter4(self):
 
     def test_does_match_returns_false_for_nonexisting_filter(self):
         for article in self.articles:
-            self.assertFalse(self.f.does_match(None, article))
+            self.assertTrue(self.f.does_match(None, article))
```
plamut: The name of this test case method should be updated as well to avoid confusion. :)

plamut:
Looks good, just the test case name needs to be changed, too.

:+1:

server/apps/content_filters/content_filter_tests.py
marwoodandrew: Yes, I my mind I'd done that!!! Will fix.

mdhaman:
:+1:

PR #1338, opened 2015-11-04
feat(search ingest): Integrate PA Images search ingest

ioanpocol:


server/apps/pa_img/pa_img_datalayer.py
```diff
@@ -0,0 +1,265 @@
...
+        raise NotImplementedError
+
+    def replace(self, resource, id_, document, original):
+        raise NotImplementedError
+
+    def remove(self, resource, lookup=None):
+        raise NotImplementedError
+
+    def is_empty(self, resource):
+        raise NotImplementedError
```
plamut: Looking at all those `raise NotImplementedError` method bodies, is the `PaImgDatalayer` meant to be an abstract class? If so, it would probably make sense to set its metaclass to `ABCmeta` and mark its abstract method with the `@abstractmethod` decorator to prevent instantiation?
(instead of raising the `NotImplementedError` only after an attempt has been made to use a non-overriden abstract method).

ioanpocol: Is not an abstract class, just that those methods should be not used.

plamut: Aha, I see now, thanks, it wasn't immediately clear from the class definition.
BTW, can we also add this extra explanation (just a sentence or two) to the class'es docstring? It will make the purpose more clear.

PR #1340, opened 2015-11-04
Update aggregate.js

tictactatic:
Added a wordier description, worthier of this very worthy widget. Edits welcome.


plamut:
I would recommend to merge related commits into a single one, makes it easier to review history later in case needed. :)

PR #1344, opened 2015-11-05
(fix) [SD-3609] - Introduce search by headline in publish queue

sjunaid:


sivakuna-aap:
@sjunaid the requirement wasn't for case insensitive search. But for fuzzy search. Check how headline search works in Global Search. Since the publish queue entries aren't in elasticsearch this might be bit tricky.

petrjasek:
maybe we can try using mongo text index and `$text` search operator here, but that would require some backend work (index setup).

sjunaid:
@sivakuna-aap and @petrjasek , IMO, `$regex` operator sufficiently working for this case, i.e, serving for approximate string matching on headline field. or am i still missing something?

sivakuna-aap:
I missed the `$regex`. Thanks for clarifying. :+1:

akintolga:
LGTM :+1:

petrjasek:
:+1:

PR #1347, opened 2015-11-05
feature(renditions): Add more default renditions

ioanpocol:


plamut:
:+1:

server/settings.py
```diff
@@ -229,6 +229,10 @@ def env(variable, fallback_value=None):
         'thumbnail': {'width': 220, 'height': 120},
         'viewImage': {'width': 640, 'height': 640},
         'baseImage': {'width': 1400, 'height': 1400},
+
+        'preview': {'width': 1024, 'height': 768},
```
petrjasek: I think those won't make it to ui, it should be part of image cropping vocabulary

ioanpocol: Should we generate them on publish?

petrjasek: no, but we should be able to edit it in the ui, and there it's using this vocabulary afaik. but I will check it

petrjasek: so in the ui you can only edit those from vocabularies, so maybe we should show there also those system defined ones, I can handle it there, it would make sense to be editable imo

petrjasek:
:+1:

plamut:
What is this massive PR now?? :)

ioanpocol:
Is an update of pa-poc branch

petrjasek:
maybe you can merge master into pa-poc without PR and then rebase this one @ioanpocol ?

PR #1348, opened 2015-11-06
need recommendations on the frontend part

actionless:
Hi, we're going to make whatsapp ingest to be a separate plugin.
For the server it's pretty trivial, but for client part i have no idea how to make that third-party plugin to override the list of ingest settings templates.

(this PR shouldn't be merged)


client/app/scripts/superdesk-ingest/module.js
```diff
@@ -59,6 +59,10 @@ define([
         search: {
             label: 'Search provider',
             templateUrl: 'scripts/superdesk-ingest/views/settings/searchConfig.html'
+        },
+        whatsapp: {
```
petrjasek: so you want to have this part in separate module right?

plamut: If I understood correctly, he wants to change/override the 'providerTypes' value?

actionless: yes, i need to append settings template for whatsapp ingest provider

petrjasek: then you can create a new module and change it there:
```
angular.module('superdesk.ingest.whatsapp', ['superdesk.ingest']).run(function(providerTypes) {
    providerTypes.whatsApp = {...};
});
```
and you have to add this module to list of modules in index.html

petrjasek: maybe it could be set to constant and not value and then you can do this using in `.config` block and not only `.run` one, but it should work the same

actionless: thanks!

ioanpocol:
Please rebase.
What is the status of this PR?

ioanpocol:
Please rebase.

ioanpocol:
What is the status of this PR?

ioanpocol:
Can we close this PR?

PR #1352, opened 2015-11-08
(feat) [SD-3602] Add outgoing stage to monitoring config

sivakuna-aap:


petrjasek:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk-archive/module.js
```diff
@@ -217,7 +217,7 @@ define([
          * and Killed.
          *
          * @param {Object} item
-         * @return true if the state of the item is in one of the published states, false otherwise.
+         * @return boolean if the state of the item is in one of the published states, false otherwise.
```
plamut: The original docstring was fine as it was, although if the intent is to describe the method's return type, it should be enclosed in curly braces.

```js
/**
 * ...
 * @return {boolean} true if ..., false otherwise
 */

File: client/app/scripts/superdesk-desks/aggregate.js [hunk 2]
```diff
@@ -492,6 +520,14 @@
                     values = _.sortBy(values, function(item) {
                         return item.order;
                     });
+
+                    _.each(values, function(item) {
+                        if (item.type === 'deskOutput') {
+                            var desk_id = item._id.substring(0, item._id.indexOf(':'));
```
plamut: It might make sense to refactor this logic of extracting content object's ID into a helper function, since identical lines appear at two other places in this file already.

client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -52,8 +52,8 @@
             });
     }
 
-    CardsService.$inject = ['api', 'search', 'session'];
-    function CardsService(api, search, session) {
+    CardsService.$inject = ['api', 'search', 'session', 'desks'];
```
plamut: It would be useful to add a docstring, summarizing in a few words what the `CardService` does.

client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -108,6 +108,27 @@
                 ]});
                 break;
 
+            case 'deskOutput':
+                var desk_id = card._id.substring(0, card._id.indexOf(':'));
```
plamut: This could use a helper utility function, too, as mention in an earlier comment (extracting desk ID from the content item's ID).

plamut:
Looks good in general. There are a few polishing tips / recommendations, but no showstoppers.

I also presume test coverage is sufficient (at least it seems like it), though I can't 100% verify it by just reading the code.

:+1:

client/app/scripts/superdesk-archive/module.js
sivakuna-aap: The editor `pycharm` I use was complaining about it and used one of the options given by it. Also jshint and eslint didn't complain about it. So left it like that.

client/app/scripts/superdesk-monitoring/monitoring.js
sivakuna-aap: Love to do that. But, I have no idea what exactly it does. We should ask someone who created or did major contribution to monitoring.

sivakuna-aap:
I've added couple of end-end tests. I haven't added unit tests though as I felt e2e would be more appropriate for this feature.

Since e2e are failing continuously I can't take the risk of addressing minor review comments in this one. I will address them in the upcoming PR's.

client/app/scripts/superdesk-archive/module.js
plamut: Hm, that's a bit strange... could it be that PyCharm's settings need to be adjusted?
(just for the reference, here's the [ngDocs documentation](https://github.com/idanush/ngdocs/wiki/API-Docs-Syntax#methods) for the method annotation syntax)

In any case, it's fine for now, especially since there are some minor inconsistencies in the docstrings at other places, too, and a more holistic approach would be needed anyway.

PR #1353, opened 2015-11-08
feat(avatars): SD-3613 Make user avatars consistent on every list

pavlovicnemanja:
SD-3613

### DIFF
--- a/client/app/scripts/superdesk/directives/sdUserAvatar.js
+++ b/client/app/scripts/superdesk/directives/sdUserAvatar.js
@@ -14,16 +14,21 @@ define(['angular'], function (angular) {
                         element.hide();
                     });
 
-                    if (scope.src) {
-                        element.attr('src', scope.src).show();
-                        figure.addClass('no-bg');
-                    } else if (scope.initials) {
-                        var initials = scope.initials.replace(/\W*(\w)\w*/g, '$1').toUpperCase();
-                        element.hide().parent().html(initials);
-                        figure.addClass('initials');
-                    } else {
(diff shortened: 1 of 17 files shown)


sivakuna-aap:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk/directives/sdUserAvatar.js
```diff
@@ -14,16 +14,21 @@ define(['angular'], function (angular) {
...
-                        element.attr('src', scope.src).show();
-                        figure.addClass('no-bg');
-                    } else if (scope.initials) {
-                        var initials = scope.initials.replace(/\W*(\w)\w*/g, '$1').toUpperCase();
-                        element.hide().parent().html(initials);
-                        figure.addClass('initials');
-                    } else {
-                        element.hide();
-                        figure.removeClass('no-bg');
+                    scope.$watch('src', initAvatar);
```
petrjasek: imo it would be better if you pass `user` to directive and there it can decide if it should render image or initials.
also with those 2 watchers it will run `initAvatar` at least 2x - so if you want to keep those params you can use `$watchGroup(['src', 'initials'], initAvatar)`

pavlovicnemanja: Good point! I will change this to pass the user to directive and there to check if it needs to show avatar or initials!

pavlovicnemanja:
#1431 Rewritten `sd-user-avatar` directive.

PR #1358, opened 2015-11-09
fix(client): copy fonts into dist folder

petrjasek:
when building client use fonts from repo


plamut:
:+1:

pavlovicnemanja:
:+1:

petrjasek:
will merge this to fix the icons, this e2e test failure is not related

PR #1359, opened 2015-11-09
feat(server: docker): read from env vars in fig wrapper

actionless:


scripts/docker-local-create-user.sh
```diff
@@ -1,7 +1,10 @@
...
-. $SCRIPT_DIR/env/bin/activate &&
-pip install -r $SCRIPT_DIR/../docker/requirements.txt &&
-cd $SCRIPT_DIR/../docker &&
+. $SCRIPT_DIR/env/bin/activate
+set -ue
+pip install -r $SCRIPT_DIR/../docker/requirements.txt
+cd $SCRIPT_DIR/../docker
+docker-compose -p sddemo -f ./docker-compose-prebuilt.yml run backend ./scripts/fig_wrapper.sh python3 manage.py app:initialize_data
 docker-compose -p sddemo -f ./docker-compose-prebuilt.yml run backend ./scripts/fig_wrapper.sh python3 manage.py users:create -u admin -p admin -e 'admin@example.com' --admin=true
+docker-compose -p sddemo -f ./docker-compose-prebuilt.yml run pubapi ./scripts/fig_wrapper.sh python3 content_api_manage.py app:prepopulate
```
petrjasek: there should be no content api imo, it's a different app

actionless: docker-compose project is located in this repo

petrjasek: which is wrong then

server/scripts/fig_wrapper.sh
```diff
@@ -2,23 +2,27 @@
 
 TIMEOUT=2
 
+MONGO_HOSTNAME=${MONGO_HOSTNAME:="mongodb"}
```
petrjasek: it would be better to use `MONGO_URI`, `ELASTICSEARCH_URL` and `REDIS_URL` env vars, those are used by the app and have port number in it, so it should work on different ports as well

actionless:
it was decided to move it to manageable code instead of using script wrapper

PR #1362, opened 2015-11-09
fix(authoring): Fixed rendition problem.

eozan:


plamut:
IMO a regression test should be added as well, while the fix itself looks fine.

petrjasek:
@eozan can you rebase or close this one? not sure it's a duplicate

eozan:
Rebased. This fixes the "Cannot read property '4-3' of undefined" js error, but I'm not sure if it's the best way.

ioanpocol:
:+1:

petrjasek:
I don't understand how it's fixing the problem, seems identical to me. @eozan can you explain?

eozan:
@petrjasek right, it's exactly identical. I will close this pr and check it again.

PR #1364, opened 2015-11-10
(fix) [SD-3630] "place" drop down list is empty in the user profile section

sivakuna-aap:
Resolves [SD-3630](https://dev.sourcefabric.org/browse/SD-3630)

### DIFF
--- a/client/app/scripts/superdesk-users/views/user-preferences.html
+++ b/client/app/scripts/superdesk-users/views/user-preferences.html
@@ -26,7 +26,7 @@
                 <div class="legend" translate>Article Defaults</div>
 
                 <div class="legend" translate>Dateline</div>
-                <div sd-info-item>
+                <div sd-info-item ng-if="preferencesLoaded === true">
                     <label translate>{{ :: preferences['dateline:located'].label }}</label>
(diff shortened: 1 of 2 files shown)


plamut:
Should we add a regression e2e test to cover the issue that has now been fixed?

Looks good though. :+1:

client/app/scripts/superdesk-users/views/user-preferences.html
```diff
@@ -26,7 +26,7 @@
                 <div class="legend" translate>Article Defaults</div>
 
                 <div class="legend" translate>Dateline</div>
-                <div sd-info-item>
+                <div sd-info-item ng-if="preferencesLoaded === true">
```
petrjasek: this `=== true` seems extra, it's only `true/false` right?

petrjasek:
:+1:

client/app/scripts/superdesk-users/views/user-preferences.html
sivakuna-aap: Agreed. But I wanted to be explicit here as even a non-null value evaluates to true.

petrjasek: I see, explicit is good

PR #1368, opened 2015-11-10
[SD-3337] Add the slugline to spike packaged item error message

marwoodandrew:


server/apps/archive/archive_spike.py
```diff
@@ -78,13 +78,14 @@ def on_update(self, updates, original):
...
         Raises an exception if the item is linked in a non-take package, the idea being that you don't whant to
-        inadvertently remove thing from packages, this force that to be done as a conscious action.
+        inadvertently remove the item from the packages, this force that to be done as a conscious action.
         :param original:
         :raise: An exception or nothing
         """
         if is_item_in_package(original):
-            raise SuperdeskApiError.badRequestError(message="This item is in a package" +
-                                                            " it needs to be removed before the item can be spiked")
+            raise SuperdeskApiError.badRequestError(
```
sivakuna-aap: You can format the whole string after concatenating instead of formatting first and then concatenating it next.

sivakuna-aap:
:+1: except the minor one

server/apps/archive/archive_spike.py
marwoodandrew: Not sure it would make it any easier to read.

plamut: The best readability would IMO be achieved if the error message is extracted into its own variable, which would then be passed as an argument to the `badRequestError()` method, though it's not really that big of a deal.

plamut:
Looks good :+1:

petrjasek:
:+1:

server/features/packages.feature
```diff
@@ -798,5 +798,5 @@ Feature: Packages
         When we spike "tag:example.com,0000:newsml_BRE9A606"
         Then we get error 400
         """
-        {"_issues": {"validator exception": "400: This item is in a package it needs to be removed before the item can be spiked"}, "_status": "ERR"}
+        {"_issues": {"validator exception": "400: The item \"WORMS\" is in a package it needs to be removed before the item can be spiked"}, "_status": "ERR"}
```
mdhaman: liked the example.

mdhaman:
:+1:

PR #1369, opened 2015-11-10
On selecting template modal window closes

Aca-jov:
SD-3377, SD-3586


pavlovicnemanja:
:+1:

plamut:
:+1: 

As always, docstrings and tests are recommended, but since the changes here are trivial and we often don't test these, it's not a showstopper here.

PR #1371, opened 2015-11-10
feat(keywords): Integration with AlchemyAPI for keywords generation

ioanpocol:


mdhaman:
:+1:  Looks Good. 

Would it be better to extract the `AlchemyAPI` calls outside of the `KeywordsService` i.e. `KeywordsService` not dependent on the `AlchemyAPI` calls. Basically, allow other third party services.

server/apps/keywords/service.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if not app.config['KEYWORDS_KEY_API']:
+            raise SuperdeskApiError.notFoundError('AlchemyAPI key is not set')
+
+        params = {'apikey': app.config['KEYWORDS_KEY_API'],
+                  'outputMode': 'json'}
+
+        url = app.config['KEYWORDS_BASE_URL'] + '/text/TextGetRankedKeywords' + '?' + urllib.parse.urlencode(params)
+        values = {'text': doc.get('body_html', '')}
+        r = self._http.post(url, data=values)
```
plamut: What happens if the request fails here? IMO such errors should be caught and handled appropriately (e.g. re-raising an error using one of the Superdesk's errors defined for such purpose).

plamut:
IMO the tests should cover a few more cases, e.g. what happens on possible errors (HTTP, config, etc.). It Would also be useful to describe the purpose of the Keywords service in a sentence or two (in the docstring of the class).

In general, though, it looks good.

petrjasek:
I think the issue is that it's only working for items in archive, this way you can't suggest when typing which is the feature they were talking about. so imo there should be api that gets text and returns suggestions. and as @sivakuna-aap said alchemy should be provider, and this keyword service should use it, we can make it configurable later, but this separation to have service and provider would make sense now - like when testing with mock provider..

ioanpocol:
Created Keywords service
Improved errors handling
Changed the code to receive the text and send back the keywords

mugurrus:
:+1:

PR #1372, opened 2015-11-11
(fix) Modified publishing code to kill broadcast content if master story is killed

mdhaman:


server/apps/archive_broadcast/broadcast.py
```diff
@@ -293,5 +295,71 @@ def spike_item(self, original):
         If Original item is re-write then it will remove the reference from the broadcast item.
         :param: dict original: original document
         """
-        if original.get('rewrite_of'):
+        broadcast_items = [item for item in self.get_broadcast_items_from_master_story(original)
+                           if item.get(ITEM_STATE) not in PUBLISH_STATES]
+
+        if broadcast_items:
```
plamut: This check is not necessary, since the following `for` loop will not run if `broadcast_items` ends up being an empty list.

server/apps/archive_broadcast/broadcast.py
```diff
@@ -293,5 +295,71 @@ def spike_item(self, original):
...
+        """
+        broadcast_items = [item for item in self.get_broadcast_items_from_master_story(original)
+                           if item.get(ITEM_STATE) in PUBLISH_STATES]
+
+        for item in broadcast_items:
+            item_id = item.get(config.ID_FIELD)
+            packages = self.packageService.get_packages(item_id)
+
+            if packages.count() > 0:
+                processed_packages = []
```
plamut: If the expected number of packages is on the high side, it might make sense to use a `set()` for this lookup collection.

server/apps/archive_broadcast/broadcast.py
```diff
@@ -293,5 +295,71 @@ def spike_item(self, original):
...
+        :return:
+        """
+        broadcast_items = [item for item in self.get_broadcast_items_from_master_story(original)
+                           if item.get(ITEM_STATE) in PUBLISH_STATES]
+
+        for item in broadcast_items:
+            item_id = item.get(config.ID_FIELD)
+            packages = self.packageService.get_packages(item_id)
+
+            if packages.count() > 0:
```
plamut: Is this needed? If empty, the following `for` loop will not run anyway...

plamut:
Looks good, with just a few non-essential recommendations.

:+1:

petrjasek:
:+1:

PR #1377, opened 2015-11-11
Pa poc headline limit

eozan:


server/apps/templates/content_templates.py
```diff
@@ -78,6 +78,7 @@ class ContentTemplatesResource(Resource):
         }},
         'last_run': {'type': 'datetime', 'readonly': True},
         'next_run': {'type': 'datetime', 'readonly': True},
+        'headline_limit': {'type': 'integer', 'default': 0},
```
sivakuna-aap: can we have default headline limit as 64 instead of 0?

petrjasek: imo it would be better to have there `limits` as a dictionary so that we don't have to update schema whenever we add a new one. eg. for `slugline` that should be there

ioanpocol:
Please rebase.

petrjasek:
obsolete

PR #1379, opened 2015-11-11
Bower build

thnkloud9:
- rename templates cache module (built with ngtemplates grunt task) to superdesk.templates-cache
- compile files for use as bower dependency (superdesk-core.js, superdesk-apps.js, vendor.js)

I'm consistently seeing two e2e tests failing:

authoring_spec.js: 
    26:41 - Expected 'TwoWordsitem5 text' to be 'Twoitem5 text'.
    28:41 - Expected 'Twoitem5 text' to be 'TwoWordsitem5 text'.

spec/helpers/search.js:
    11:1 - Failed: unknown error: Element is not clickable at point (193, 178). Other element would receive the click: <span class="state-border"></span>

Not sure if the tests themselves need updating, or what else should be considered to resolve these failures.


ioanpocol:
E2e tests are ok
There is an issue on the other tests/checks:

tasks/options/clean.js
  23:61  error  Unexpected trailing comma  comma-dangle

vied12:
:bowtie::ok_hand::sunny:

client/app/scripts/superdesk/superdesk.js
```diff
@@ -97,8 +97,13 @@ define([
         }]);
 
     // make sure there is templates module defined
-    angular.module('templates', []);
-    modules.push('templates');
+    try {
```
petrjasek: so superdesk.js must be loaded always after templates-cache right?

vied12: Yes. Or we could copy the same try-catch in templates-cache to remove this constraint if it's important

vied12:
After that, it would make a lot of sense to use the release system in order to update the bower files regularly :)

client/app/scripts/superdesk/superdesk.js
petrjasek: I'm just wondering if it's obvious. maybe this logic should be in main.js and not superdesk.js ?
why should it care about templates?

vied12: It is there because the previous template module was added here before. (Look at the removed lines)
And this is the template file of the superdesk module only, isn't it @thnkloud9  ?

thnkloud9: actually, its templates cache for all superdesk modules.  probably does make more sense to move the try to main.js instead of superdesk.js, I'll make the change.

thnkloud9:
moved templates-cache module init to main, but not I'm having e2e test failures again.  This time timeout.

thnkloud9:
Aside from the failing e2e test that is timing out, I think this is ready to go.  Please let me know if there is anything else that should be considered.

petrjasek:
:+1:

ioanpocol:
Please rebase.

thnkloud9:
I've based, and everything is green now.

vied12:
it looks there is a conflict :)

thnkloud9:
so close, but so far away.

vied12:
Don't give up !

Pop corn

thnkloud9:
eh, can't seem to win with this one.  looks like unit tests are now failing, although they seem fine for me when run locally.

petrjasek:
all green now :+1:

PR #1390, opened 2015-11-16
(fix) [SD-3663] Takes are being created in the default incoming stage instead of the working stage

sivakuna-aap:
Resolves: [SD-3663](https://dev.sourcefabric.org/browse/SD-3663)


mugurrus:
:+1:

plamut:
:+1:

PR #1393, opened 2015-11-16
[sd-3666]-UI: Newly created desk doesn't reflect the stage count

sjunaid:
Fix covers SD-3676: Update is not initiated when adding/removing a stage to/from a desk

### DIFF
--- a/client/spec/helpers/desks.js
+++ b/client/spec/helpers/desks.js
@@ -53,7 +53,7 @@ function Desks() {
     /**
      * Returns the stage count for named desk on desks settings list
      * @param {string} name of desk
-     * @return {string} stage count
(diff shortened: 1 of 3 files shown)


client/spec/helpers/desks.js
```diff
@@ -51,6 +51,19 @@ function Desks() {
     };
 
     /**
+     * Returns the stage count for named desk on desks settings list
+     * @param {string} name of desk
+     * @return {string} stage count
```
plamut: This doesn't seem right, since the method actually returns a _promise_ which is resolved with the stage count?

akintolga:
:+1:

ioanpocol:
Please rebase.

ioanpocol:
:+1:

PR #1394, opened 2015-11-16
feat(content_types): introduce content types

petrjasek:
it can be used for validation if exists

@eozan I've only added fields to schema that should be validated, not sure if we should hide everything else or how you want to use it?

### DIFF
--- a/server/apps/content_types/content_types.py
+++ b/server/apps/content_types/content_types.py
@@ -1,0 +1,38 @@
+
+import superdesk
+
+
+CONTENT_TYPE_PRIVILEGE = 'content_type'
+
+
+class ContentTypesResource(superdesk.Resource):
+    schema = {
+        '_id': {
+            'type': 'string',
+            'unique': True,
+        },
+        'label': {
+            'type': 'string',
+        },
+        'schema': {
+            'type': 'dict',
+        },
+        'priority': {
+            'type': 'integer',
+            'default': 0
+        }
+    }
+
+    item_url = 'regex("[\w,.:-]+")'
+
+    privileges = {'POST': CONTENT_TYPE_PRIVILEGE,
+                  'PATCH': CONTENT_TYPE_PRIVILEGE,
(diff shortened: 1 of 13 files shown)


server/apps/content_types/content_types.py
```diff
@@ -0,0 +1,30 @@
+
```
mugurrus: I would suggest renaming the file to resource.py to be in line with the convention we used most of the time.

petrjasek: it's not only resource in there, it's the whole feature, that's why the name.

mugurrus: Well, basically you have here a resource and a service you don't need - you can use BasicService directly in __init__.py
See: https://github.com/superdesk/superdesk-content-api/tree/master/content_api/clients

server/apps/validate/validate.py
```diff
@@ -38,10 +37,22 @@ def create(self, docs, **kwargs):
 
         return [doc['errors'] for doc in docs]
 
+    def _get_validators(self, action, item_type, profile=None):
+        """Get validators.
+
+        In case there is profile defined for item with respective content type it will
+        use its schema for validations, otherwise it will fall back to action/item_type filter.
+        """
```
plamut: I would recommend adding parameter descriptions as well, it will make it easier to drill-down through the code without an unnecessary additional guessing burden (e.g. when debugging).

server/apps/content_types/content_types.py
```diff
@@ -0,0 +1,30 @@
...
+import superdesk
+
+
+CONTENT_TYPE_PRIVILEGE = 'content_type'
+
+
+class ContentTypesResource(superdesk.Resource):
```
plamut: Can we add a brief summary of what ContentType resource is all about?
(one or two sentences in the class docstring should already be enough)

plamut:
IMO it's good, just added a few additional minor remarks.

:+1:

ioanpocol:
Please rebase.

PR #1396, opened 2015-11-16
Feature/New editor

vied12:
## What this PR contains

- The previous `sdTextEditor` directive is now named `sdTextEditorBlockText`.
- `sdTextEditor` is now a collection of blocks, which can be either `sdTextEditorBlockText` (text input) or `sdTextEditorBlockEmbed` (embed preview + embed editor).
- The directive `sdAddEmbed` which allows to create an embed block (`sdTextEditorBlockEmbed`) from
  - a link, and we use noEmbed or Embedly via [angular-embed](https://github.com/superdesk/angular-embed) to retrieve the embed code.
  - an embed code which can contain anything (html, scripts, style) that will be inserted into the block body as it was given
  - an image drag and dropped from the archives collection

## What can be improved later

- [ ] Loading indicator when embed preview are loading
- [ ] Error messages when the link given for embed returns no results
- [ ] Drop zone for images is a lot smaller on Chrome and should be as big as for Firefox
- Improve the text editor keyboard behaviours
  - [ ] Let the ability to add a break line without creating a new block (move the current behaviour for `Enter` to `Shift` + `Enter`. Then `Enter` will react as before.)
  - [ ]  response time
  - [ ]  arrows navigation through the blocks
- [ ] Add fields on embed blocks to write descriptions or credit


mugurrus:
It would be good to consolidate the 21 commits in this PR into a few commits. Having too many commits makes the log harder to read.

vied12:
This is why I describe the work done in the PR description.
I work in this PR for more than 1 week. I think this is normal to have some commits here ^^
Also I think that change logs shouldn't rely on commits which are made to save progression but more on PR and release messages.
If you really need to limit the number of commit, we can merge it at the end with `- - squash`

ahilles107:
commits looks ok for me - i would squash just few of them (with fixed of new code). 
I'm not sure yet how it works in Superdesk but template strings translations are missing?

client/app/scripts/superdesk/editor/views/add-embed.html
```diff
@@ -0,0 +1,25 @@
+<div class="add-embed__plus" ng-if='!vm.extended' ng-click="vm.toggle()" sd-text-editor-drop-zone>
+    <i class="icon-plus-small icon-round icon-white"></i>
+    <div class="add-embed__drop-zone"></div>
+</div>
+<div class="add-embed__body" ng-if='vm.extended'>
```
plamut: Double quotes? (for consistency)

client/app/scripts/superdesk/editor/views/add-embed.html
```diff
@@ -0,0 +1,25 @@
...
+        </div>
+        <div class="form-group">
+            <button type="reset" class="btn btn-small btn-default" ng-click="vm.toggle()">
+                <i class="icon-close-small"></i>
+            </button>
+            <button type="submit" class="btn btn-small btn-info" ng-click="vm.createBlockFromEmbed()">
+                <i class="icon-white icon-ok"></i>
+            </button>
+        </div>
+        Preview:
```
plamut: Texts should be translated, a canonical pattern looks like this:
```html
{{:: 'Preview:' | translate }}
```
(basically using the `translate` filter to do the heavy lifting)

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -550,16 +550,375 @@ function EditorService(spellcheck, $rootScope, $timeout) {
...
-            anchorInputPlaceholder: gettext('Paste or type a full link'),
-            disablePlaceholders: true,
+    .directive('sdAddEmbed', function() {
+        return {
+            scope: true,
+            require: ['sdAddEmbed', '^sdTextEditor'],
+            templateUrl: 'scripts/superdesk/editor/views/add-embed.html',
+            controllerAs: 'vm',
+            controller: ['embedService', '$element', '$timeout', '$q', 'lodash',
+                        function(embedService, $element, $timeout, $q, _) {
```
plamut: It might make sense to define controller as a standalone function that is then just passed here as the value of the `controller` property.
It's a subject to personal preferences, though, I would personally find the directive definition more readable (also, inlining the controller needs a few extra levels of indentation, hindering readability on smaller screens).

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -550,16 +550,375 @@ function EditorService(spellcheck, $rootScope, $timeout) {
...
+                                };
+                            });
+                        // otherwise we use the content of the field directly
+                        } else {
+                            var embedType = 'Custom';
+                            // try to guess the provider of the custom embed
+                            if (vm.input.indexOf('twitter.com/widgets.js')) {
+                                embedType = 'Twitter';
+                            } else if (vm.input.indexOf('https://www.youtube.com')){
+                                embedType = 'YouTube';
```
plamut: It might make sense to define these constant strings in some sort of an "enum", easier to reuse them at other places if needed.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -550,16 +550,375 @@ function EditorService(spellcheck, $rootScope, $timeout) {
...
+                        vm.retrieveEmbed().then(function(embed) {
+                            vm.createFigureBlock(embed.provider, embed.body);
+                            // close the addEmbed form
+                            vm.toggle(true);
+                        });
+                    },
+                    createBlockFromPicture: function(picture) {
+                        vm.createFigureBlock('Image', [
+                            '<img alt="' + picture.description + '" src="' + picture.renditions.viewImage.href + '">',
+                            '<figcaption>' + picture.headline + '</figcaption>'
```
plamut: Just checking - is `picture.headline` properly escaped by this point?
(the same for its `description` and any other attributes)

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -550,16 +550,375 @@ function EditorService(spellcheck, $rootScope, $timeout) {
...
+                });
+            }
+        };
+    })
+    .directive('sdTextEditor', ['$timeout', 'lodash', function ($timeout, _) {
+        return {
+            scope: {type: '=', config: '=', language: '='},
+            require: ['sdTextEditor', 'ngModel'],
+            templateUrl: 'scripts/superdesk/editor/views/editor.html',
+            controllerAs: 'vm',
```
plamut: Since using `controllerAs`, consider combining this with `bindToController: true` so that the directive properties get bound to the controller instead of the scope.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -550,16 +550,375 @@ function EditorService(spellcheck, $rootScope, $timeout) {
...
+        return {
+            scope: {type: '=', config: '=', language: '=', sdTextEditorBlockEmbed: '='},
+            require: ['sdTextEditorBlockEmbed', 'ngModel'],
+            template: [
+                '<textarea ng-show="vm.editable" ng-model="vm.embedCode" class="embed-editor"></textarea>',
+                '<div ng-show="!vm.editable" class="embed-preview"></div>',
+                '<a class="btn btn-default btn-small" ng-show="!vm.editable" ng-click="vm.toggleEdition()">Edit the code</a>',
+                '<a class="btn btn-default btn-small" ng-show="vm.editable" ng-click="vm.cancel()">Cancel</a>',
+                '<a class="btn btn-default btn-small" ng-show="vm.editable" ng-click="vm.save()">Save</a>'].join(''),
+            controllerAs: 'vm',
```
plamut: `bindToController: true` here, too?

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -550,16 +550,375 @@ function EditorService(spellcheck, $rootScope, $timeout) {
...
+    .directive('sdTextEditorBlockEmbed', ['$timeout', function ($timeout) {
+        return {
+            scope: {type: '=', config: '=', language: '=', sdTextEditorBlockEmbed: '='},
+            require: ['sdTextEditorBlockEmbed', 'ngModel'],
+            template: [
+                '<textarea ng-show="vm.editable" ng-model="vm.embedCode" class="embed-editor"></textarea>',
+                '<div ng-show="!vm.editable" class="embed-preview"></div>',
+                '<a class="btn btn-default btn-small" ng-show="!vm.editable" ng-click="vm.toggleEdition()">Edit the code</a>',
+                '<a class="btn btn-default btn-small" ng-show="vm.editable" ng-click="vm.cancel()">Cancel</a>',
+                '<a class="btn btn-default btn-small" ng-show="vm.editable" ng-click="vm.save()">Save</a>'].join(''),
```
plamut: Should we have the template in its own file? IMO more readable, since it's not that short...

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -725,11 +1107,74 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck'])
 
                     scope.cursor = {};
                     render(null, null, true);
+
+                    function extractBlockContentsFromCaret() {
+                        function getBlockContainer(node) {
+                            while (node) {
+                                if (node.nodeType === 1 && /^(P|H[1-6]|DIV)$/i.test(node.nodeName)) {
```
plamut: Maybe consider adding a comment that we test for paragraphs, divs and headings, makes it somewhat easier to read. :)

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -725,11 +1107,74 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck'])
...
+                            }
+                        }
+                    }
+
+                    // Actions to support multi blocks edition
+                    if (scope.config.multiBlockEdition) {
+                        editorElem.on('keyup', function(e) {
+                            $timeout(function () {
+                                // press enter, create a new block
+                                 if (e.keyCode === 13) {
```
plamut: Despite the comment, I think it still makes sense to define a key map for even better readability, e.g.
```js
if (e.keyCode === KEYS.ENTER) {...
```

plamut:
I really like this feature! And it's great to see the comments in code, as well as in the PR description itself, it  makes it much easier to review because it explains the big picture at the very start.

Since the feature is obviously complex, tests are almost mandatory before merging this. To make future maintenance easier, I recommend adding docstrings to all those controller methods as well, because they are mostly non-trivial.

IMO some things should not be inlined (e.g. templates, controllers), too much nesting makes it harder to read the code later.

All in all this is a very good attempt and I'm looking forward to when this is finished and merged. Keep up the good work!

ioanpocol:
Please rebase.

petrjasek:
:+1: merged it now, comments will be addressed in later PR

PR #1397, opened 2015-11-16
feat(shortcuts): SD-3509 Added ability to open highlights list and mark some item with keyboard

pavlovicnemanja:
SD-3509

Please fill free to give some tips for code improvement! I couldn't find better way to implement this.

### DIFF
--- a/client/app/scripts/superdesk-monitoring/monitoring.js
+++ b/client/app/scripts/superdesk-monitoring/monitoring.js
@@ -350,12 +350,38 @@
                     superdesk.findActivities(intent, item).forEach(function (activity) {
                         if (activity.keyboardShortcut && workflowService.isActionAllowed(item, activity.action)) {
                             monitoring.bindedItems.push(activity.keyboardShortcut);
-
                             keyboardManager.bind(activity.keyboardShortcut, function () {
-                                activityService.start(activity, {data: {item: scope.selected}});
+                                if (activity._id === 'mark.item') {
+                                    bindMarkItemShortcut();
+                                } else {
+                                    activityService.start(activity, {data: {item: scope.selected}});
+                                }
                             }, {inputDisabled: true});
                         }
                     });
+                }
+
+                /*
+                 * Bind highlight dropdown action
+                 * Keyboard shortcut is defined with action
+                 *
+                 * @param {Object} item
+                 */
+                function bindMarkItemShortcut() {
+                    elem.find('.active .more-activity-toggle').click();
+                    var highlightDropdown = angular.element('.more-activity-menu.open .dropdown-noarrow');
+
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -350,15 +350,41 @@
...
+                    var highlightDropdown = angular.element('.more-activity-menu.open .dropdown-noarrow');
+
+                    highlightDropdown.addClass('open');
+                    if (highlightDropdown.find('button').length > 0) {
+                        highlightDropdown.find('button')[0].focus();
+                        keyboardManager.push('up', function () {
+                            highlightDropdown.find('button:focus').parent('li').prev().children('button').focus();
+                        });
+                        keyboardManager.push('down', function () {
+                            highlightDropdown.find('button:focus').parent('li').next().children('button').focus();
```
plamut: `highlightDropdown.find('button:focus').parent('li')` repeats itself. Instead of repeatedly obtaining a reference to the same DOM element (takes time...) it is better to store that reference into a variable and use it in the code (it's also better for readability).

Example:
```js
// name it anything you want
var $focusBtnParent = highlightDropdown.find('button:focus').parent('li');
...
$focusBtnParent.prev().chidlren(...) ...
$focusBtnParent.next().chidlren(...) ...
```

plamut:
@pavlovicnemanja What parts of the code are you unsure about, what do you think it might be less than ideal in this PR? Or is it just a general question?

IMO a test would be good to have (probably an e2e), proving that the new addition indeed works.

pavlovicnemanja:
@plamut I was thinking in general. Hmmm, I was thinking about all this clicking and focusing. I tought that the best way to implement this is to imitate the actual click (not including directives, calling functions etc.). Because of that, I did this with jquery. I had a problem to return binders to previously defined keys (for example `up` and `down` in list are moving throw the list, but when I open highlights list I need to go throw highlights list, and when I close highlights list I need to bind walking throw the list again.

Yeah, I will write test for this for sure. I'm just fixing some stupid issues.

client/app/scripts/superdesk-monitoring/monitoring.js
pavlovicnemanja: It is not pointing to the same element every time. `Find` function is looking for focused button, so if I put this out of `keyboardManager` function, it will pointing to the same button every time.

plamut: Ah, I missed the `:focus` pseudo-class, you're right. Nothing to gain here then, unless maybe refactoring it out to a helper function. My bad.

ioanpocol:
Please rebase.

ioanpocol:
:+1:

PR #1398, opened 2015-11-16
fix(templates): New content item created from template will be added …

ioanpocol:
…on the current selected desk

### DIFF
--- a/client/app/scripts/superdesk-workspace/content/content.js
+++ b/client/app/scripts/superdesk-workspace/content/content.js
@@ -58,6 +58,7 @@
          */
         this.createItemFromTemplate = function(template) {
             var item = _.pick(template, templates.TEMPLATE_METADATA);
+            archiveService.addTaskToArticle(item);
             return save(item).then(function(newItem) {
                 templates.addRecentTemplate(desks.activeDeskId, template._id);
                 return newItem;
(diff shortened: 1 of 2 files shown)


client/app/scripts/superdesk-archive/module.js
```diff
@@ -140,8 +140,8 @@ define([
          * @param {Object} desk when passed the item will be assigned to this desk instead of user's activeDesk.
          */
         this.addTaskToArticle = function (item, desk) {
-            if ((!item.task || !item.task.desk) && (desk || desks.activeDeskId && desks.userDesks)) {
-                var currentDesk = desk || _.find(desks.userDesks._items, {_id: desks.activeDeskId});
+            if ((!item.task || !item.task.desk) && (desk || desks.getCurrentDeskId() && desks.userDesks)) {
```
sivakuna-aap: `getCurrentDeskId()` will always return `deskId` if user is member of desk(s). This prevents from creating articles in personal workspace. That's the reason `activeDeskId` was used.

If this is been taken care of then there should be test to check if creation of article is possible or not in personal workspace.

ioanpocol: In desks dropdown is not possible to select 'Personal' anymore so the content item should be added to the current selected desk on desks dropdown.
The change should be performed for the code that adds content items to 'Personal'

sivakuna-aap: Agreed that there's no option to select personal in desks dropdown. My point was when user is either in custom workspaces or selects personal option from left then the item shouldn't get created in a desk and this is not been taken care in the changes.

plamut:
:+1:

petrjasek:
:+1:

PR #1399, opened 2015-11-16
Add the default username and password to the readme.

will-in-wi:
This adds some info which I had trouble finding as I worked on setting up an initial install.

Thanks!


mugurrus:
:+1:

PR #1401, opened 2015-11-17
[SD-3166] - Use the 'en' dictionary if the language of story is 'en-US' or 'en-AU'

akintolga:


akintolga:
This PR replaces my earlier (closed few weeks ago) PR #1257. 

With this change system will use the specific dictionary (i.e. en-AU) if it exists otherwise it will use base dictionary (en) if it exists.  If the story is set to have en as the language then it will use 'en', not  'en-AU' even if it exists. So we expect the base dictionary (en) is set up to reflect the main language the system will use.

@petrjasek I have done some refactoring in the suggest endpoint as well.

server/apps/dictionaries/service.py
```diff
@@ -111,8 +137,8 @@ def get_model_for_lang(self, lang):
         :param lang: language code
         """
         model = {}
-        lookup = {'$and': [{'language_id': lang}, {'is_active': {'$in': ['true', None]}}]}
-        dicts = self.get(req=None, lookup=lookup)
+        dicts = self.get_dictionaries(lang)
```
mugurrus: I suggest caching dictionaries and models. This takes a lot of processing and the spelling suggestion requests should happen quite often.

mugurrus:
:+1: (except the minor comment)

client/app/scripts/superdesk-dictionaries/dictionaries.js
```diff
@@ -96,14 +96,19 @@
          *
          * @param {string} lang
          */
```
plamut: Docstring needs an update here (new parameter).

server/apps/dictionaries/service.py
```diff
@@ -103,6 +103,32 @@ def on_fetched(self, result):
...
+        """
+        languages = [{'language_id': lang}]
+        base_language = self.get_base_language(lang)
+        if base_language:
+            languages.append({'language_id': base_language})
+
+        lookup = {'$and': [{'$or': languages},
+                           {'$or': [{'is_active': {'$exists': 0}}, {'is_active': 'true'}]}]}
+        dicts = list(self.get(req=None, lookup=lookup))
+        langs = [d['language_id'] for d in dicts]
```
plamut: `langs` could be a `set()` here, but it probably doesn't matter because the expected list length is small.

plamut:
Looks pretty good all in all!

:+1:

PR #1403, opened 2015-11-17
Fixed content templates issues

Aca-jov:
SD-3432

### DIFF
--- a/client/app/scripts/superdesk-templates/templates.js
+++ b/client/app/scripts/superdesk-templates/templates.js
@@ -244,10 +244,17 @@
                     $scope.origTemplate = null;
                     $scope.template = null;
                     $scope.vars = null;
+                    fetchTemplates();
                 };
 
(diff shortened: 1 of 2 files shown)


plamut:
I would add a regression test as well.

Aca-jov:
I don't need tests are needed, because I fixed some smaller bugs. I think this is too much going in details: test for checking if some button is disabled or if some options are selected/deselected.
Can someone else also give it a look?

plamut:
I just thought maybe `$scope.validSchedule()` could be tested, although if fixes just revolve around enabling/disabling UI widgets, then we _might_ skip the tests. We generally don't test these indeed, though there were cases where similar small changes required a test, and others where bigger changes were merged without having one.

I won't oppose if the consensus is that a test is not needed here, as the changes themselves seem fine.
:+1:

ioanpocol:
Please rebase.

PR #1408, opened 2015-11-18
fix(e2e tests): Split e2e tests in multiple suites and rerun failed t…

ioanpocol:
…ests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ sudo: false
 env:
     - TARGET=unit
     - TARGET=server_ldap
-    - TARGET=e2e_chrome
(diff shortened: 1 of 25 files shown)


plamut:
@ioanpocol What do you mean by the failed tests being re-run? And what happens if a re-run fails, too?

ioanpocol:
If a test fail, all tests from that spec file will be re-run (see: protractor-flake).
If after 3 tries there are still errors, the e2e tests job will stop with fail result.

plamut:
OK, sounds good. :+1:

mugurrus:
:+1:

PR #1410, opened 2015-11-18
fix(publish): update versioncreated on kill/correct

petrjasek:
it creates a new version so versioncreated should get updated

### DIFF
--- a/server/apps/publish/content/correct.py
+++ b/server/apps/publish/content/correct.py
@@ -26,6 +26,7 @@ class CorrectPublishService(BasePublishService):
         CropService().validate_multiple_crops(updates, original)
         super().on_update(updates, original)
         updates[ITEM_OPERATION] = ITEM_CORRECT
+        updates['versioncreated'] = utcnow()
         set_sign_off(updates, original)
 
     def on_updated(self, updates, original):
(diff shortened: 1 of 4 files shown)


server/apps/publish/content/correct.py
```diff
@@ -26,6 +26,7 @@ def on_update(self, updates, original):
         ArchiveCropService().validate_multiple_crops(updates, original)
```
sivakuna-aap: Looks like your superdesk-core is pointing to old SHA. ArchiveCropService has been moved to superdesk-core as CropService.

server/apps/publish/content/correct.py
```diff
@@ -26,6 +26,7 @@ def on_update(self, updates, original):
...
         super().on_update(updates, original)
         updates[ITEM_OPERATION] = ITEM_CORRECT
+        updates['versioncreated'] = utcnow()
```
sivakuna-aap: Can we move this to super class BasePublishService?

ioanpocol:
Please rebase.

server/apps/publish/content/correct.py
petrjasek: I think this should not happen when you publish, but only when you kill and correct

server/apps/publish/content/correct.py
petrjasek: yep there was a different PR updating it, it's merged now

server/apps/publish/content/correct.py
sivakuna-aap: Any specific reason for not updating versioncreated on publish? I think there's no difference between publish, correct and kill operations.

ioanpocol:
Done, without errors.

./features/steps/steps.py:1833:121: E501 line too long (122 > 120 characters)

/home/travis/build.sh: line 45: cd: server: No such file or directory

/home/travis/build.sh: line 45: cd: server: No such file or directory

ioanpocol:
:+1:

PR #1419, opened 2015-11-19
Feat desk metadata

petrjasek:
it adds metadata to desk and populate it on item when sending it to desk or creating it there


mugurrus:
:+1:

ioanpocol:
:+1:

PR #1431, opened 2015-11-20
fix(avatars): Fixed issue with avatars and inactive state for users

pavlovicnemanja:


plamut:
I would add a few tests for the `sdUserAvatar` directive, but the fix itself looks fine to me.

pavlovicnemanja:
There are some tests for this directive, I just rewrite it. Now, I'm passing whole user to directive and there decide if it should display avatar or image.

ioanpocol:
:+1:

client/app/scripts/superdesk/directives/sdUserAvatar.js
```diff
@@ -5,27 +5,37 @@ define(['angular'], function (angular) {
...
                 link: function (scope, element, attrs) {
-
-                    var figure = element.parents('figure');
+                    scope.$watch('user', function (user) {
+                        if (user) {
+                            initAvatar(user);
+                        }
+                    });
 
                     element.on('error', function (e) {
```
petrjasek: now you can handle it gracefully and display initials also when it can't fetch user image that should be there..

petrjasek:
:+1:

PR #1432, opened 2015-11-20
new editor: use embedly instead of noEmbed and support generic embed SDPA-45

vied12:
A new env var called `EMBEDLY_KEY` is now required to use the new editor


plamut:
I would recommend adding a test or two. Also, are we _sure_ that the parameters to the methods building HTML snippets have already been escaped? If not, these methods should do that.

The rest seems good to me.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -1225,6 +1267,18 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck', 'angular-embe
                 }
             }
         };
-    }]);
+    }]).config(['embedlyServiceProvider', 'config', function(embedlyServiceProvider, config) {
+        // embed.ly private key
+        console.log(config);
```
plamut: Is this needed or was it just used during the development?

vied12: development, it's still not working. This `config` doesn’t contain what it should at this point

plamut: Oh, OK, then just don't forget to add the `WIP` label so that this does not get merged prematurely. :)

petrjasek:
:+1:

petrjasek:
hey - is this going to be ready anytime soon? I did image crop based on this but can do it also on pa-poc probably

vied12:
it can be merged now

petrjasek:
:+1:

mugurrus:
:+1:

PR #1435, opened 2015-11-23
SD-3278 Find packages in which an item is included

mugurrus:


server/apps/packages/package_service.py
```diff
@@ -146,6 +147,16 @@ def check_package_associations(self, docs):
             for assoc in associations:
                 self.extract_default_association_data(group, assoc)
 
+    def check_not_in_personal_space(self, docs):
+        for doc in docs:
```
sivakuna-aap: It would be good to add some documentation why we are adding this check for packages?

sivakuna-aap:
:+1: except the minor one

server/apps/packages/package_service.py
petrjasek: I will check this condition with @tictactatic might be too strict

plamut:
:+1:

server/apps/packages/package_service.py
```diff
@@ -146,6 +147,21 @@ def check_package_associations(self, docs):
             for assoc in associations:
                 self.extract_default_association_data(group, assoc)
 
+    def check_not_in_personal_space(self, docs):
```
petrjasek: so according to @tictactatic there shouldn't be such limitation, but as a short term fix it's ok

petrjasek:
:+1:

akintolga:
:+1:

server/apps/macros/macro_register.py
```diff
@@ -44,7 +44,6 @@ def register_macros():
                      'callback' in dir(sys.modules[m]) and
                      not m.endswith('_test')
                      and not m.startswith('__')]
-    print(macro_modules)
```
mdhaman: There was a hack introduced long time back to solve the problem where macros were not getting loaded for celery jobs. I am not sure if `print` was the hack or not. But it would be worth checking if the macro functionality works fine with celery jobs or not?

mdhaman:
:+1: looks good (except for the comment related to macros)

mugurrus:
Will merge it as it and make another PR to add back the print in macros.

PR #1436, opened 2015-11-23
Feat content types limits

eozan:


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1986,6 +1987,13 @@
                         }
 
                         _.extend(item, updates);
+
+                        if (item.profile) {
+                            content.getType(item.profile)
+                            .then(function(type) {
+                                scope.contentType = type;
```
petrjasek: imo it would be good to assign also `scope.schema` and use it in template instead of `contentType.schema...` like that, we can have some default schema in case there is no content type for profile

client/app/scripts/superdesk-workspace/content/content.js
```diff
@@ -63,6 +63,44 @@
...
+         */
+        this.getTypes = function() {
+            var self = this;
+            return api.query('content_types').then(function(result) {
+                self.types = result._items;
+                return self.types;
+            });
+        };
+
+        this.getType = function(id) {
```
plamut: Docstring should be added here as well, if all the other methods have it, too.

plamut:
IMO a few (behave) tests should be added for the `ContentTypesResource` resource. Perhaps also a test or two for the new code in the `headerInfoDirective`.

Other than that it looks pretty good IMO, and nice work with the comments!

petrjasek:
this is part of https://github.com/superdesk/superdesk/pull/1394

PR #1437, opened 2015-11-23
Sd 3507

sjunaid:


client/spec/send_spec.js
```diff
@@ -70,4 +76,26 @@ describe('send', function() {
...
+        workspace.open();
+        workspace.editItem(1);
+
+        //Skip spell check
+        authoring.toggleAutoSpellCheck();
+        expect(element(by.model('spellcheckMenu.isAuto')).getAttribute('checked')).toBeFalsy();
+
+        authoring.writeText('Text, that not saved yet');
+        authoring.sendTo('Sports Desk');
+        element(by.className('modal-content')).all(by.css('[ng-click="ok()"]')).click();
```
sivakuna-aap: This is repeated multiple times. Can we assign it to a variable in it's helper class?

PR #1440, opened 2015-11-24
[SD-2450] Automatic routing to publish

marwoodandrew:
It seems that if the publish happens due to a routing rule we do not get a version created in archive_versions, presumably due to the fact that the publish is not done under the eve context, this PR fixes that. (not in love with the solution, would be happy for alternative suggestions)

It also includes a macro that will truncate the headline, slugline and adds a dateline (bogus), this is to allow Reuters content to pass validation and successfully publish, in order for the functionality to be tested.

### DIFF
--- a/server/apps/publish/content/common.py
+++ b/server/apps/publish/content/common.py
@@ -157,6 +157,7 @@ class BasePublishService(BaseService):
         try:
             user = get_user()
             last_updated = updates.get(config.LAST_UPDATED, utcnow())
+            auto_publish = updates.pop('auto_publish', False)
 
             if original[ITEM_TYPE] == CONTENT_TYPE.COMPOSITE:
                 self._publish_package_items(original, updates)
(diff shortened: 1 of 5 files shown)


server/macros/update_to_pass_validation.py
```diff
@@ -0,0 +1,46 @@
...
+    At the moment virtually all content received from Reuters fails validation.
+    :param item:
+    :param kwargs:
+    :return:
+    """
+    try:
+        item['slugline'] = item['slugline'][:24] if len(item['slugline']) > 24 else item['slugline']
+        item['headline'] = item['headline'][:64] if len(item['headline']) > 64 else item['headline']
+        if 'dateline' not in item:
+            cities = find_cities(country_code='AU', state_code='NSW')
```
mugurrus: It would be good to have some configured defaults here instead of hardcoded values.

mugurrus:
:+1: (except comment)

ioanpocol:
:+1: (except comment)

server/apps/publish/content/common.py
```diff
@@ -157,6 +157,7 @@ def update(self, id, updates, original):
         try:
             user = get_user()
             last_updated = updates.get(config.LAST_UPDATED, utcnow())
+            auto_publish = updates.pop('auto_publish', False)
```
petrjasek: it could also do `auto_publish = config.VERSION in updates` so it would make sure there is at least one version stored for any item published, but not sure it's any better..

sivakuna-aap:
:+1:

akintolga:
:+1:

PR #1441, opened 2015-11-24
chore(app init): Add support for app init profile and add pa profile

ioanpocol:


petrjasek:
this should go to `pa-poc` branch right?

ioanpocol:
The profile support and one profile as an example can be on master.
But if not I can move it to pa-poc branch.

petrjasek:
well at this point we need this on pa branch, not sure about master, so would be easier to have it just there

ioanpocol:
ok

PR #1442, opened 2015-11-24
chore(app init): Add support for multiple profiles for app init and a…

ioanpocol:
…dd a new pa profile data


petrjasek:
not sure all this data should be public, like some ingest settings..

petrjasek:
or emails

ioanpocol:
Deleted the ingest settings and pa email address

server/apps/prepopulate/app_initialize.py
```diff
@@ -50,19 +54,20 @@ class AppInitializeWithDataCommand(superdesk.Command):
     """
 
     option_list = [
-        superdesk.Option('--entity-name', '-n', dest='entity_name', default='')
+        superdesk.Option('--entity-name', '-n', dest='entity_name', default=''),
+        superdesk.Option('--profile', '-p', dest='profile')
```
petrjasek: imo better way would be to make directory a param so that we don't have to push this into repo but it can be in some private repo/space and still it could be imported. or to be able to read each entity from a stdin so that you can import it into docker without adding it into an image (but not sure it would be possible like `cat users.json > docker exec -ti manage.py app:initialize_data --entity-name users`)

ioanpocol: Is not practical to be forced to execute a command for every file but to have it as an option can be useful.
Re: directory parameter, you should copy anyway the files to image, so why not to copy in the 'profile' directory?

petrjasek: well in case you don't use docker to deploy, it would be useful to point this to different directory outside of git repo so you don't have to deal with conflicts when updating. with docker it doesn't matter, but there is no need for multiple profiles, you can just copy to root folder

PR #1443, opened 2015-11-24
fix(paimg): Change the url used for PA Images

ioanpocol:


petrjasek:
:+1:

mugurrus:
:+1:

petrjasek:
@ioanpocol any more work to be done here?

ioanpocol:
I changed back to use 'search' endpoint. Tests are failing, so we have to update from master.

petrjasek:
why switching back to `search` ?

ioanpocol:
Ian asked on hipchat to switch back but to keep the 'ck' parameter. Now also on 'search' the response time is good.

petrjasek:
I see :+1:

PR #1444, opened 2015-11-24
feat(authoring): Changed keywords with tags, implemented fetching key…

eozan:
…words from alchemy api, changed UI.

SDPA-26

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/views/metadata-tags.html
+++ b/client/app/scripts/superdesk-authoring/metadata/views/metadata-tags.html
@@ -6,5 +6,9 @@
 <div class="terms">
     <ul>
         <li ng-class="{selected: isSelected(t)}" ng-repeat="t in tags track by t" ng-click="toggle(t)">{{ t }}</li>
+        <li tooltip="{{:: 'Refresh suggestions' | translate}}" tooltip-placement="left" ng-click="refresh()">
+            <span ng-if="refreshing" translate>Refreshing...</span>
+            <i ng-if="!refreshing" class="alt icon-refresh"></i>
+        </li>
     </ul>
 </div>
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-authoring/metadata/views/metadata-tags.html
```diff
@@ -0,0 +1,11 @@
+<div class="dropdown dropright dropdown-bigger dropdown-terms">
+    <input class="line-input" type="text" ng-show="adding" ng-model="newTag" ng-disabled="disabled" ng-keyup="key($event)">
+    <button class="dropdown-toggle" ng-show="!adding" ng-disabled="disabled" ng-click="adding = true"><i class="icon-white icon-plus-large"></i></button>
+</div>
+
+<div class="terms">
+    <ul>
+        <li ng-if="!disabled" ng-class="{selected: isSelected(t)}" ng-repeat="t in tags track by t" ng-click="toggle(t)">{{ t }}</li>
```
petrjasek: imo it would be better to handle disabled state in `toggle` func to avoid this duplication

petrjasek:
does it handle content changes? or you have to close/open the sidebar?

eozan:
@petrjasek it doesnt handle content changes automatically. I suggest having a manual refresh button for that, as we have a limited number of api calls.

client/app/scripts/superdesk-authoring/metadata/views/metadata-tags.html
eozan: done

petrjasek:
i think idea was to have it automatic but somehow cache it (this cache should be done on server).
but you can check this with pa tomorrow

petrjasek:
so it seems there should be a button. also pls add comments to your code. and tests if possible

eozan:
Refresh button and comments are added.

petrjasek:
:+1:

ioanpocol:
please rebase to get e2e test fixes

PR #1445, opened 2015-11-24
Editor: when multiBlockEdition is true, wrap all blocks into paragraphs SDPA-49

vied12:
When creating a story with a single paragraph (e.g. a SNAP) HTML P tags was not including in the resulting response.


mugurrus:
:+1:

petrjasek:
:+1:

PR #1446, opened 2015-11-24
feat(picture_crop): add picture crop resource

petrjasek:
it can create a new crop using given picture item and crop specs


ioanpocol:
:+1:

PR #1447, opened 2015-11-24
feat(authoring): SD-3698 Improved usability of Subject and Category popup

pavlovicnemanja:
SD-3698

### DIFF
--- a/client/app/scripts/superdesk/ui/ui.js
+++ b/client/app/scripts/superdesk/ui/ui.js
@@ -397,6 +397,76 @@ define([
         };
     }
 
+    DropdownFocus.$inject = ['keyboardManager'];
+    function DropdownFocus(keyboardManager) {
+        return {
+            require: 'dropdown',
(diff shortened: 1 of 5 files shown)


client/app/scripts/superdesk/ui/ui.js
```diff
@@ -397,6 +397,69 @@ define([
...
+
+                            if (prevElem.length > 0) {
+                                prevElem.focus();
+                            } else {
+                                inputField.focus();
+                                categoryButton.focus();
+                            }
+                        }, keyboardOptions);
+
+                        keyboardManager.push('down', function () {
```
petrjasek: I think it will also need to pop it at some point, otherwise this will leak memory

petrjasek: like on dropdown close

ioanpocol:
:+1:

Aca-jov:
:+1:

PR #1448, opened 2015-11-25
(feat) Apply Template on the Backend 

mdhaman:
This feature allows applying of template on the server.
endpoint: content_templates_apply
payload: {template_name, item}
Response: modified item (no database updates)

- [x] Allow registration of new Filters 
- [x] Create format_datetime filter used by Kill Template
- [x] New Endpoint to apply template
- [x] Unit/Behave test for new endpoint
- [x] Modify client use the new endpoint for killing any item
- [x] Modify content template edit/create screen so that Desk/schedule is not allowed for kill
- [x] Modify broadcast kill to use this new endpoint.
- [x] e2e tests for kill/templates


server/apps/templates/content_templates.py
```diff
@@ -119,6 +126,60 @@ def get_scheduled_templates(self, now):
         return self.find(query)
 
 
+class ContentTemplatesApplyResource(Resource):
+    endpoint_name = 'content_templates_apply'
+    resource_title = endpoint_name
+    schema = item_schema()
```
petrjasek: imo it would be easier to use if url is just `content_templates_apply` and there is template and item in schema..

mdhaman: Ok. Will modify the schema and url to use simple url.

mdhaman:
ping @superdesk/developers.

marwoodandrew:
:+1:

sivakuna-aap:
:+1:

PR #1449, opened 2015-11-25
Sd 3335 - System should remember last sent-to stage and desk

sjunaid:


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1917,6 +1939,13 @@
                     }
                 }
 
+                function updateLastDestination(deskId, stageId) {
+                    var PREFERENCE_KEY = 'destination:active';
```
sivakuna-aap: I see these declarations as redundant.

client/app/scripts/superdesk-desks/desks.js
```diff
@@ -642,6 +642,15 @@
                     },
 
                     /**
+                     * Fetch current user destination desks and stage
+                     */
+                    fetchLastDestination: function() {
```
sivakuna-aap: Ideally this should be in `preferenceService`. And I think preferenceService already takes care of this. If we use preferenceService we might not need this function.

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -1890,6 +1902,16 @@
                         })
                         .then(function(value) {
                             notify.success(gettext('Item sent.'));
+
+                            // Remember last destination desk and stage
+                            if (scope.destination_last) {
```
sivakuna-aap: The two `if's` can be merged into one

PR #1451, opened 2015-11-25
fix(content_types): rename article to story

petrjasek:
there is no way to remove an item via initialize_data command,
so adding a param - enabled - and use it for filtering in ui.


ioanpocol:
:+1:

PR #1460, opened 2015-11-26
Fix img search

petrjasek:


server/apps/pa_img/pa_img_datalayer.py
```diff
@@ -119,15 +144,13 @@ def find(self, resource, req, lookup):
...
+            url += '/latest'
+        else:
+            url += '/search'
 
-        page = int(req.get('from', '0')) // size + 1
-        if page > 1:
-            fields['page'] = page
+        offset, limit = int(req.get('from', '0')), max(10, int(req.get('size', '25')))
+        fields['limit'] = limit
+        fields['page'] = math.ceil(offset / limit) + 1  # starts with 1
```
ioanpocol: Here some items can be skipped.  For example for offset 29 and limit 5, the result is from page 6 limit 5 that will lbe from 30 to 35 instead of 29 to 34.

server/apps/pa_img/pa_img_datalayer.py
```diff
@@ -76,13 +96,18 @@ def find(self, resource, req, lookup):
...
         fields = {}
         if 'query' in req['query']['filtered']:
-            query_keywords = req['query']['filtered']['query']['query_string']['query']
-            query_keywords = query_keywords.replace('slugline:', 'keywords:')
-            query_keywords = query_keywords.replace('description:', 'caption:')
-            fields = extract_params(query_keywords, ['headline', 'caption', 'keywords'])
+            query = req['query']['filtered']['query']['query_string']['query'] \
+                .replace('slugline:', 'keywords:') \
+                .replace('description:', 'caption:')
+            for key in ('headline', 'keywords', 'caption'):
```
ioanpocol: A query like this will not work, (all) will be ignored:
'(all) headline:(head one) caption:(capt) keywords:(key)'

petrjasek: yep but that's by design, should we support such queries?

ioanpocol: The interface allows you to do that.

server/apps/pa_img/pa_img_datalayer.py
petrjasek: right updating

server/apps/pa_img/pa_img_datalayer.py
petrjasek: I know, but is this required atm?

ioanpocol: If you search in all text fields by 'test', it will appear at the top of the list x('test')
If then you search by headline 'collins' the x('test') is still at the top of the list but actually it is not used.

petrjasek: I see will look into it

petrjasek: fixed that part

ioanpocol:
:+1:  Scroll is working far better.

server/apps/pa_img/pa_img_datalayer.py
```diff
@@ -119,15 +135,13 @@ def find(self, resource, req, lookup):
                         fields['photos'] = 'true'
 
         if not fields:
```
ioanpocol: Please change the code here (https://dev.sourcefabric.org/browse/SDPA-52):
if not fields:
+            url += '/latest'
+            fields['ck'] = 'public'
+            fields['days_since'] = 5
+        else:
+            url += '/search'
+            fields['ck'] = 'superdesk'

server/apps/pa_img/pa_img_datalayer.py
```diff
@@ -34,16 +34,32 @@
...
-            params['text'] = result.group(1)
+    for name, value in findall:
+        query = query.replace('%s:(%s)' % (name, value), '')
+    query = query.strip()
+    if query:
+        params['q'] = query
     return params
 
 
+def process_limits(offset, size):
```
ioanpocol: Not used anymore, can be deleted

PR #1461, opened 2015-11-26
new editor: editable caption for images SDPA-44

vied12:
Support for:
- [x] featured image caption
- [x] embedded image caption


petrjasek:
:+1:

vied12:
e2e 1 and 3 passed, but not the number 2. What does it mean ?

petrjasek:
it means there is an error :smile:

vied12:
haha ok but what is the difference between 1,2 and 3 ?

petrjasek:
it's just splitted into 3 groups to run faster

vied12:
I thought it was to give a chance for unstable tests to succeed at least one time ^^
Ok will see what I can do if they fail again

PR #1464, opened 2015-11-27
chore(server): update core

petrjasek:
use decription_text for picture captions

### DIFF
--- a/client/app/scripts/superdesk-authoring/views/item-association.html
+++ b/client/app/scripts/superdesk-authoring/views/item-association.html
@@ -1,4 +1,4 @@
 <figure class="item-association" ng-class="{preview: related}">
     <img ng-if="related" ng-src="{{ related.renditions.viewImage.href }}" ng-click="edit(related)">
-    <figcaption contenteditable="true" ng-model="related.description" ng-change="onchange()"></figcaption>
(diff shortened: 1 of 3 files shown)


petrjasek:
@vied12 @ioanpocol pls review

ioanpocol:
:+1:

client/app/scripts/superdesk-authoring/views/item-association.html
```diff
@@ -1,4 +1,4 @@
 <figure class="item-association" ng-class="{preview: related}">
     <img ng-if="related" ng-src="{{ related.renditions.viewImage.href }}" ng-click="edit(related)">
-    <figcaption contenteditable="true" ng-model="related.description" ng-change="onchange()"></figcaption>
+    <figcaption contenteditable="true" ng-model="related.description_text" ng-change="onchange()"></figcaption>
```
vied12: why don't use the `description` field ?
if it's in order to keep the original value, I would name the new one `caption` instead.
Because it will be difficult later to understand why `description` and `description_text` can be different.

petrjasek: I've changed this also on server, it is description_text in pa and in ninjs standard, in is published like that. there should be no `description` field

PR #1467, opened 2015-11-27
fix(privileges): SD-2546 Fixed layout for user preferences

pavlovicnemanja:
SD-2546


ioanpocol:
Looks like there is a e2e error:

Failures:
1) users editing user privileges: should reset the form to the last saved state when the Cancel button is clicked
  Message:
    Failed: Index out of bound. Trying to access element at index: 0, but there are only 0 elements that match locator By.cssSelector(".roles-list input[type=\"checkbox\"]")

mugurrus:
:+1: (except e2e test fix)

petrjasek:
:+1:

PR #1468, opened 2015-11-27
LBSD-740 fix websocket reconnection

vladnicoara:
Websocket connection must be retried if there is an existing ws with the
readyState of CLOSED


ioanpocol:
:+1:

vied12:
:ok:

PR #1470, opened 2015-11-27
chore(server): update core

petrjasek:


petrjasek:
it's fixing publishing of keywords

ioanpocol:
:+1:

PR #1473, opened 2015-11-30
(imp) [SD-3387] Ingest refactoring

sivakuna-aap:
Depends: https://github.com/superdesk/superdesk-core/pull/91
Resolves: [SD-3387](https://dev.sourcefabric.org/browse/SD-3387)


server/apps/io/feeding_services/reuters.py
```diff
@@ -61,67 +46,38 @@ def _update(self, provider):
         if not last_updated or last_updated < updated - datetime.timedelta(minutes=ttl_minutes):
             last_updated = updated - datetime.timedelta(minutes=ttl_minutes)
 
-        for channel in self.get_channels():
-            for guid in self.get_ids(channel, last_updated, updated):
+        self.provider = provider
+        self.URL = provider.get('config', {}).get('url')
```
petrjasek: I'm not sure this part should be configurable, there is only 1 url

sivakuna-aap: Reason for moving `url` to `config` is there might be other http services which might have different URL and in future if reuters changes the url we don't want to touch the code. In any case the configuration is done at once.

mdhaman:
:+1:

mugurrus:
:+1:

server/apps/io/feeding_services/reuters.py
petrjasek: problem is those urls could be hard to find, it's part of developers documentation, not sure what exactly you get when you subscribe. so keeping it could make things simpler, maybe as default values user could change via config?

sivakuna-aap:
@petrjasek Modified the code as per your comments. Added the default URL in both client and server places. If everything's ok then can we merge this PR and wipe the data in master.dev?

petrjasek:
:+1:

PR #1475, opened 2015-11-30
editor: fix issue for tweet with picture inside

vied12:
Use the user-provided url instead of the embedly one because when this is an image tweet, the embedly url links to the image, instead of the tweet.


petrjasek:
:+1:

mugurrus:
:+1:

PR #1477, opened 2015-11-30
fix(templates): ignore template timestamps

petrjasek:
when creating an item from template generate first/version created fields

SD-3433


mdhaman:
:+1:

mugurrus:
It works but I think it's better to save the timestamp of the last operation in the context and compare against it.
:+1:

PR #1479, opened 2015-11-30
modals: header height shouldn't be fixed

vied12:
when the text content was longer than the fixed width, the text was cut


mugurrus:
:+1:

petrjasek:
:+1:

PR #1489, opened 2015-12-01
Extra error message when publishing with body missing

marwoodandrew:
The problem was attempting to append the body in the save and publish to a null body in the digital package item.

@mugurrus This is the task that @sivakuna-aap linked to yours, it might be worth merging just for the test. If you want to take care of t yourself just let me know and I'll close it.


mugurrus:
:+1: (it's sure worth merging)

ioanpocol:
:+1:

PR #1491, opened 2015-12-02
[sd-3170]-Make the floating, green toolbar in the editor configurable

sjunaid:
configuration implemented as grunt option. i.e: 'disableEditorToolbar'.
usage: --disableEditorToolbar = true, in grunt command, will disable editor toolbar

### DIFF
--- a/client/app/scripts/superdesk/editor/editor.js
+++ b/client/app/scripts/superdesk/editor/editor.js
@@ -556,11 +556,14 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck'])
 
     .directive('sdTextEditor', ['editor', 'spellcheck', '$timeout', function (editor, spellcheck, $timeout) {
 
+        var disableToolbar = window.config.server.disableEditorToolbar || false;
+
         var config = {
             buttons: ['bold', 'italic', 'underline', 'quote', 'anchor'],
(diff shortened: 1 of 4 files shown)


client/tasks/options/template.js
```diff
@@ -13,7 +14,7 @@ module.exports = function(grunt) {
 
         var config = {
             raven: {dsn: process.env.SUPERDESK_RAVEN_DSN || ''},
-            server: {url: server, ws: ws},
+            server: {url: server, ws: ws, disableEditorToolbar: disableEditorToolbar},
```
sivakuna-aap: It would be good to move this option to a separate group instead of mixing with server.

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -556,11 +556,14 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck'])
 
     .directive('sdTextEditor', ['editor', 'spellcheck', '$timeout', function (editor, spellcheck, $timeout) {
 
+        var disableToolbar = window.config.server.disableEditorToolbar || false;
```
petrjasek: here it should use `config` value via angular di - so `function(editor, spellcheck, $timeout, config) { ...`

ioanpocol:
:+1:

mugurrus:
:+1:

PR #1497, opened 2015-12-03
Fixed add to package for killed items

Aca-jov:
SD-3381


mugurrus:
A regression test would be good here

ioanpocol:
:+1:

PR #1490, opened 2015-12-01
Desk mentions

akintolga:
Desk Mention Changes:
- Desk names can be mentioned in the story comments
- Desk comment count will appear next to the desk/stage widget in monitoring (next to item count)
- Hover action will display the comments per that desk
- Users can open the story from the comment
- First user opening the story will acknowledge the message

Activity and Notification Changes:
- Two separate activities created for mentions: desk:mention and user:mention
- Normal users will get these notifications only
- Client doesn't poll the notifications anymore, desk and user mention messages are broadcasted to client
- Activity end point returns only the last 24h of data

When this is merged please clean the activities collection.

PR #1500, opened 2015-12-04
(imp) [SD-3754] Refactor external search repositories from Ingest Provider

sivakuna-aap:
Resolves: [SD-3754](https://dev.sourcefabric.org/browse/SD-3754)
Also, the tests which were marked as "tobefixed" in https://github.com/superdesk/superdesk/pull/1490 have been fixed in this PR.


server/apps/search_providers/aap_mm/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+def init_app(app):
+    # Must be imported here as the constant PROVIDER_NAME is referenced by the below modules
+    from .aap_mm_datalayer import AAPMMDatalayer
+    from .resource import AAPMMResource
+    from .service import AAPMMService
+
+    app.data.aapmm = AAPMMDatalayer(app)
+
+    aapmm_service = AAPMMService(datasource=None, backend=app.data.aapmm)
```
petrjasek: just wondering if there should be service/resource for every search provider we have, some `search_provider_search` resource that would accept query + search provider id and delegate that to those custom datalayers (or there could be different abstraction - with only query, find_one and fetch methods maybe). to create service and resource for every provider seems like quite some overhead, and there is not much aap specific in it.

sivakuna-aap: Intention was to restore the lost AAP Multimedia option because of Ingest Refactoring. So while doing this didn't delve into the implementation details of how the backend/resource/service is implemented. 

If this is fine then after merging this PR, we can improve on top of this by understanding how both PA Image and AAP Images is implemented.

petrjasek: sure that's fine

petrjasek:
:+1:

akintolga:
:+1:

mdhaman:
:+1:

PR #1501, opened 2015-12-04
fix(highlights): Show highlights info for published items

ioanpocol:


petrjasek:
it would be good to replace those 2 lines with a directive and keep it in on place, seems like quite some duplication

ioanpocol:
Done

pavlovicnemanja:
:+1:

PR #1502, opened 2015-12-04
template fixes

petrjasek:
- fix save as desk template
- let only user with privileges create desk templates


pavlovicnemanja:
:+1:

mugurrus:
:+1:

PR #1503, opened 2015-12-06
[SD-3021],[SD-3743],[SD-3776] An empty text editor box returns empty markup causing grief for validation

marwoodandrew:

### DIFF
--- a/server/apps/prepopulate/app_prepopulate_data.json
+++ b/server/apps/prepopulate/app_prepopulate_data.json
@@ -244,7 +244,13 @@
 		            "type" : "integer",
 		            "required": true,
 		            "allowed": [1, 2, 3, 4, 5]
-		        }
+		        },
+                "body_html": {
+                    "type": "string",
+                    "required": true,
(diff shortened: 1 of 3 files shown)


server/apps/prepopulate/app_prepopulate_data.json
```diff
@@ -244,7 +244,13 @@
...
 		            "allowed": [1, 2, 3, 4, 5]
-		        }
+		        },
+                "body_html": {
+                    "type": "string",
+                    "required": true,
+                    "nullable": false,
+                    "empty": false
+                }
 		    }
```
sivakuna-aap: I think this validation is required for correct also.

marwoodandrew: Yes, but this update is just so the test triggers a validation error

sivakuna-aap: Aha!!! :+1:

ioanpocol:
:+1:

PR #1504, opened 2015-12-07
(fix) [SD-3729] Filter condition value does not show the actual readable value

sivakuna-aap:
Resolves [SD-3729](https://dev.sourcefabric.org/browse/SD-3729)


mdhaman:
:+1:

ioanpocol:
:+1:

PR #1508, opened 2015-12-07
(fix) [SD-3729] Issues in Filter Conditions

sivakuna-aap:
Resolves: [SD-3279](https://dev.sourcefabric.org/browse/SD-3729)
1. Fixing the swift appearance of angular script while loading conditions. 
2. Also, clearing the value from the condition when the field is changed


mugurrus:
:+1: (a regression test would be nice)

ioanpocol:
:+1:

PR #1510, opened 2015-12-08
[SD-3755] refactor(server: apps: archive: common): use uuid1 instead of redis sequence for key generation

actionless:


server/apps/archive/common.py
```diff
@@ -169,16 +169,9 @@ def generate_unique_id_and_name(item, repo_type=ARCHIVE):
...
-            else '{}_SEQ'.format(repo_type.upper())
-
-        unique_id = update_key(key_name, flag=True)
-
-        if unique_id:
-            item['unique_id'] = unique_id
-            item['unique_name'] = "#" + str(unique_id)
-        else:
-            raise IdentifierGenerationError()
+        unique_id = str(uuid1())
```
sivakuna-aap: Unique Id can't be a string as per the requirement. Please check schema definition https://github.com/superdesk/superdesk-core/blob/master/superdesk/metadata/item.py#L58

Also, as per python docs https://docs.python.org/3.4/library/uuid.html uuid1() generates alpha numeric string..

actionless: should we get red of `unique_name` field and make `uniqued_id` string instead of integer?

sivakuna-aap: As per requirement we need both. `unique_id` is for internal storage and is not editable. But `unique_name` can be changed by user and also user can open a story using `unique_name (ctrl+0) is shortcut key`. 

More information can be found in https://wiki.sourcefabric.org/display/NR/Content+Metadata

actionless: what do you think about just changing `unique_id` schema to `int`?

sivakuna-aap: `unique_id` is already declared as integer as per eve/cerberus spec. Please refer https://github.com/superdesk/superdesk-core/blob/master/superdesk/metadata/item.py#L58

Please clarify if I misunderstood your comment.

actionless: how do you suggest to proceed with it then? using random int? or storing sequence in mongo?

petrjasek: we should store the sequence, probably in some `sequences` collection - there is no entity where we should keep it

actionless: but it will be not unique when MAX_SEQ_NUMBER value will be reached and it will rotate it

petrjasek: right, but items get out of the collection after expiring or publishing, it should be unique during ~24h period

actionless: how do you think it would be better, directly use pymongo or create an eve resource?

mugurrus: A service would offer a standard API. We need this sequences service because there will be more sequences that can't go into an existing collection. I need it on publishing refactoring too.

ioanpocol:
So can we close this PR?

actionless:
yes, first i will add `sequences` resource to superdesk-core and after it'll be merged, open a new PR here to use it

PR #1512, opened 2015-12-08
Monitoring widget refreshes on item update

Aca-jov:
SD-3600


petrjasek:
why should it refresh whole list on lock/unlock?

Aca-jov:
Yes, it should refresh only locked item, i will change.

Aca-jov:
It's changed.

ioanpocol:
:+1:

petrjasek:
:+1:

PR #1516, opened 2015-12-09
[SD-3787] - Desk mention improvements

akintolga:


client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -294,6 +294,13 @@
                 scope.cacheNextItems = [];
                 scope.cachePreviousItems = [];
                 scope.limited = (monitoring.singleGroup || scope.group.type === 'highlights') ? false : true;
+                scope.stageLookup = {};
+
+                if (scope.group.type === 'stage') {
```
sivakuna-aap: The ticket says Incoming Stage, but here it's working stage. Also, whats the strategy if user turns off the stage where the notification gets displayed?

akintolga: Thanks for pointing out, it is supposed to be the incoming stage. I have double checked with PO's and they're happy that it will appear only on incoming and if that stage is not there in the desk monitoring view then they won't see the notifications.

ioanpocol:
There are two tests failing: 


1) notifications create a new desk mention
  Message:
    Failed: No element found using locator: By.id("deskNotifications")


PhantomJS 1.9.8 (Linux 0.0.0) monitoring desk notification directive can initiate the desk notifications FAILED
	Expected 0 to be 1.
	    at /home/travis/build/superdesk/superdesk/client/app/scripts/superdesk-monitoring/monitoring.spec.js:9

mugurrus:
:+1:

mdhaman:
:+1:

PR #1518, opened 2015-12-09
feat(splitter): Created functionality for vertical splitter between monitoring and authoring

pavlovicnemanja:


ioanpocol:
:+1:

Aca-jov:
:+1:

client/app/scripts/superdesk/ui/ui.js
```diff
@@ -1142,8 +1142,83 @@ define([
...
 
+                /*
+                 * If authoring is not initialized,
+                 * wait, and initialize it again
+                 *
+                 * This issue is observed when you are
+                 * switching from settings back to monitoring
+                 */
+                if (!authoring.length) {
+                    $timeout(function () {
```
petrjasek: you can set third param to false not to trigger another digest if you only change css

pavlovicnemanja: Third param for $timeout?

petrjasek: yep third to $timeout. it should be still fine, as long as you don't need any binding to update, only rendering. if you would make some changes to scope then you would need a digest

pavlovicnemanja: :+1:

PR #1526, opened 2015-12-14
SD-3067 Publish stories asynchronously

mugurrus:


akintolga:
I think we better have this in a separate branch on its own while we review the functionality once the wip is removed from this PR. So we and QA can check the publishing functionality before merging.

server/apps/publish/content/common.py
```diff
@@ -86,61 +86,13 @@ class BasePublishService(BaseService):
...
-        self._validate_package_contents(original, takes_package, package_validation_errors)
-        if len(package_validation_errors) > 0:
-            raise ValidationError(package_validation_errors)
-
+        updates['digital_item_id'] = self._get_digital_id_for_package_item(original)
+        self._validate(original, updates)
         self._set_updates(original, updates, updates.get(config.LAST_UPDATED, utcnow()))
-        updates[ITEM_OPERATION] = ITEM_PUBLISH
-        convert_task_attributes_to_objectId(updates)
+        convert_task_attributes_to_objectId(updates) # ???
```
mdhaman: Problem is how the task attributes are stored. It is stored as string or `ObjectId` and for mongo queries to work we need all as `ObjectId`.  Other option might be to have schema for the `task`?

ioanpocol:
Please rebase

mugurrus:
Made the final PR on superdesk-core

PR #1530, opened 2015-12-14
chore(app initialization): Add support for ingest sources that reads …

ioanpocol:
…username/password from env variables

### DIFF
--- a/server/apps/prepopulate/app_initialize.py
+++ b/server/apps/prepopulate/app_initialize.py
@@ -2,9 +2,11 @@ import os
 import json
 import superdesk
 import pymongo
+import settings
 
 from superdesk import get_resource_service
 from flask import current_app as app
+from re import findall
 
 
 """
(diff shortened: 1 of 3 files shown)


server/apps/prepopulate/app_initialize.py
```diff
@@ -167,4 +171,21 @@ def import_file(self, entity_name, file_name, index_params, do_patch=False):
                 self.logger.info('Index: {} for collection {} created successfully.'.format(index_name, entity_name))
 
 
+def fillEnvironmentVariables(config, item):
+    variables = {}
+    text = json.dumps(item)
+
+    for variable in findall('#ENV_([^#"]+)#', text):
+        value = config.get(variable, None)
```
petrjasek: maybe it could read it from `env` so that we don't have to add it into settings for every var we want to use?

server/apps/prepopulate/app_initialize.py
```diff
@@ -167,4 +172,21 @@ def import_file(self, entity_name, file_name, index_params, do_patch=False):
                 self.logger.info('Index: {} for collection {} created successfully.'.format(index_name, entity_name))
 
 
+def fillEnvironmentVariables(item):
```
sivakuna-aap: Not a pythonic way

sivakuna-aap:
:+1: except the minor one

PR #1532, opened 2015-12-14
template fixes

petrjasek:
sorry by mistake pushed those 2 fixes together, but both are small so will keep it like that


ioanpocol:
Running "jshint:all" (jshint) task
   app/scripts/superdesk/datetime/datetime.js
     50 |            }
                      ^ Missing semicolon.

ioanpocol:
:+1:

PR #1533, opened 2015-12-14
Disabled spiking locked item in multiselect

Aca-jov:
SD-2959, SD-3563

### DIFF
--- a/client/app/scripts/superdesk-monitoring/aggregate.js
+++ b/client/app/scripts/superdesk-monitoring/aggregate.js
@@ -335,19 +335,6 @@
             });
         };
 
-        /**
-         * Reset on all groups the search set by user
-         */
-        this.resetSearch = function() {
-            this.searchQuery = null;
-            _.each(this.groups, function(item) {
-                item.query = null;
-            });
-            _.each(this.spikeGroups, function(item) {
-                item.query = null;
-            });
-        };
-
         this.state = storage.getItem('agg:state') || {};
         this.state.expanded = this.state.expanded || {};
 
--- a/client/app/scripts/superdesk-monitoring/monitoring.js
+++ b/client/app/scripts/superdesk-monitoring/monitoring.js
@@ -71,7 +71,7 @@
             var params = {};
 
             if (card.type === 'search' && card.search && card.search.filter.query) {
-                params = card.search.filter.query;
+                angular.copy(card.search.filter.query, params);
                 if (card.query) {
                     if (card.search.filter.query.q) {
                         params.q = '(' + card.query + ') ' + card.search.filter.query.q;
--- a/client/app/scripts/superdesk-search/search.js
+++ b/client/app/scripts/superdesk-search/search.js
@@ -1981,7 +1981,7 @@
         this.canSpikeItems = function() {
             var canSpike = true;
             multi.getItems().forEach(function(item) {
-                canSpike = canSpike && authoring.itemActions(item).spike;
+                canSpike = canSpike && authoring.itemActions(item).spike && !item.lock_user;
             });
             return canSpike;
         };


client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -73,11 +73,9 @@
             if (card.type === 'search' && card.search && card.search.filter.query) {
                 params = card.search.filter.query;
                 if (card.query) {
-                    if (card.search.filter.query.q) {
-                        params.q = '(' + card.query + ') ' + card.search.filter.query.q;
```
Aca-jov: @akintolga as I saw you added this code couple months ago, can you please review this changes that I made, are they ok. I didn't understand why on every search, query is added every time o previous query (even when i delete search), maybe it is needed somewhere else, and it shouldn't be deleted.

client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -73,11 +73,9 @@
...
-                    } else {
-                        params.q = '(' + card.query + ') ';
-                    }
```
Aca-jov: @akintolga as I saw you added this code couple months ago, can you please review this changes that I made, are they ok. I didn't understand why on every search, query is added every time o previous query (even when i delete search), maybe it is needed somewhere else, and it shouldn't be deleted.

client/app/scripts/superdesk-monitoring/aggregate.js
```diff
@@ -335,19 +335,6 @@
             });
         };
 
-        /**
```
ioanpocol: This code reset the current search for every group, you have to keep it.

client/app/scripts/superdesk-monitoring/monitoring.js
ioanpocol: You have to keep card.search.filter.query.q, because it is from saved search groups type.

ioanpocol:
Disabled spiking locked item in multiselect: I think that this condition should be added where the multi spike command is implemented:
https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-archive/module.js#L105

client/app/scripts/superdesk-monitoring/aggregate.js
Aca-jov: No, I checked, this function is never called, take a look. there is function:
```     this.search = function(query) {
            this.searchQuery = query;
            _.each(this.groups, function(item) {
                item.query = query;
            });
            _.each(this.spikeGroups, function(item) {
                item.query = query;
            });
        };```
On reset it is called search(null), so no need for this.

Comment (ioanpocol) [inline client/app/scripts/superdesk-monitoring/aggregate.js]: ok

File: client/app/scripts/superdesk-archive/module.js [hunk 4]
```diff
@@ -80,19 +80,21 @@ define([
...
-                    }
-                    return item;
-                }, function(response) {
-                    item.error = response;
-                    if (angular.isDefined(response.data._issues) &&
-                        angular.isDefined(response.data._issues['validator exception'])) {
-                        notify.error(gettext(response.data._issues['validator exception']));
-                    }
-                });
+            if (!item.lock_user) {
```
sivakuna-aap: There should be a message to the user why the item couldn't be spiked.

ioanpocol: I think that you have to check if the list of selected items can be spiked and enable multi spike/unspike button if the operation is possible.

Aca-jov:
If one of items is locked, spike button is hidden.

ioanpocol:
:+1:

pavlovicnemanja:
:+1:

PR #1541, opened 2015-12-15
fix(authoring): Fixed unexpected view/edit and send-to behaviour for …

eozan:
…published items.

SD-3557

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -2515,8 +2515,9 @@
          * @param {string} action
          */
         this.edit = function(item, action) {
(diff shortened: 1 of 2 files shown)


sivakuna-aap:
@eozan The fix doesn't address the cases of corrected and kill states of the article. Also, the statement in `authoring-topbar.html` hides the send-to panel if the item is in published state and thus blocking the user from correcting or killing the article.

As part of [SD-2097](https://dev.sourcefabric.org/browse/SD-2097) I have fixed this issue as well and the fix also covers few more issues. Can we close this PR if you are ok?

eozan:
@sivakuna-aap, sure, closing this one.

PR #1548, opened 2015-12-17
(fix) [SD-3705]-(fix-multiselect issue in deskoutput stage)-Multi select doesn't remember selections after scroll

sjunaid:


akintolga:
:+1:

mdhaman:
:+1:

PR #1549, opened 2015-12-17
[SD-3847] Create broadcast icon should only appear for privileged users.

mdhaman:


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -563,9 +564,9 @@
                 !current_item.embargo && current_item.package_type !== 'takes' &&
                 current_item.state !== 'killed' && !current_item.publish_schedule;
 
-            action.create_broadcast = (!_.contains(['spiked', 'scheduled', 'killed'], current_item.state)) &&
-                (_.contains(['published', 'corrected'], current_item.state)) &&
-                current_item.type === 'text' && !isBroadcast;
+            action.create_broadcast = (_.contains(['published', 'corrected'], current_item.state)) &&
+                _.contains(['text', 'preformatted'], current_item.type) &&
```
sivakuna-aap: ` _.contains(['text', 'preformatted'], current_item.type)` is already covered by `isBroadcast`

mdhaman: its not covered. `isBroadcast` is set to `True` for broadcast item and false for the rest. we need this so that `create_broadcast` action available for text items (which are not broadcast item) and not for image, audio and video.

akintolga: :+1:

sivakuna-aap:
:+1:

PR #1554, opened 2015-12-17
refactor search view using react

petrjasek:
rendering list of items in search using react, both grid and list view. it can render 1000+ items without affecting digest time. it's fast, but instead of templates you define elements in js.
what do you think? would this work for designers? for devs? didn't use [jsx](https://facebook.github.io/jsx/) not to complicate build process atm, but it would be also an option.
ping @superdesk/developers 
ps: there are some changes made to templates as well while I was first trying to make it fast with angular, but it seems to be impossible

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ install:
 before_script:
     - if [ "${TARGET}" != "${TARGET/e2e/}" ]; then
         cd server ;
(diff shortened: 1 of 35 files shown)


akintolga:
I tried this PR with my local copy that it speeds things up in rendering. It will help in continuous scroll. I'll check the memory when we stop removing the items from the top of the list during scroll and once it is merged. 
For the jsx: in the long it may make things easier to read. But since this approach will be only on list views we can getaway without jsx.

:+1: Overall it looks good.

pavlovicnemanja:
Nice Work! :+1:

PR #1558, opened 2015-12-18
(fix):Adding index and fix query for legal archive.

mdhaman:


sivakuna-aap:
:+1:

mdhaman:
will add unit test. missing lamut :angry:

akintolga:
:+1:

PR #1561, opened 2015-12-21
[SD-3353] - Ingest routing schedule is not updated

akintolga:


server/apps/rules/routing_rules.py
```diff
@@ -261,6 +261,8 @@ def _adjust_for_empty_schedules(self, routing_scheme):
             schedule = rule.get('schedule')
             if schedule and (set(schedule.keys()) == {'time_zone'}):
                 rule['schedule'] = None
+            elif 'time_zone' not in schedule.keys():
```
sivakuna-aap: This will fail if `schedule` is `None`

akintolga: Schedule can be none, then this rule won't be active as `_get_scheduled_routing_rules` will not return that rule. In that case time_zone is meaningless.

sivakuna-aap:
:+1:

mugurrus:
:+1:

PR #1566, opened 2015-12-23
(feat) [SD-3035] Kill a content item in the (dusty) archive

sivakuna-aap:


server/apps/archived/archived.py
```diff
@@ -88,5 +212,74 @@ def on_fetched_item(self, doc):
     def _get_archived_id(self, item_id, version):
         return '{}:{}'.format(item_id, version)
 
+    def _is_article_in_a_package(self, article):
```
akintolga: There is a similar function in common.py, see if you can reuse it.

server/apps/archived/archived.py
```diff
@@ -78,6 +92,116 @@ def on_create(self, docs):
                 if is_takes_package:
                     doc[SEQUENCE] = len(package_service.get_item_refs(doc))
 
+    def on_delete(self, doc):
```
akintolga: If the kill action is triggered via the master story then we should locate the broadcast stories and kill them as well.

akintolga:
:+1: Nice job.

server/apps/archived/archived.py
```diff
@@ -78,6 +92,129 @@ def on_create(self, docs):
...
+        if get_resource_service(ARCHIVE).find_one(req=None, _id=doc[GUID_FIELD]):
+            raise bad_req_error(message="Can't Kill as article is still available in production")
+
+        if is_item_in_package(doc):
+            raise bad_req_error(message="Can't kill as article is part of a Package")
+
+        takes_package_service = TakesPackageService()
+        takes_package_id = takes_package_service.get_take_package_id(doc)
+        if takes_package_id:
+            if get_resource_service(ARCHIVE).find_one(req=None, _id=takes_package_id):
```
mdhaman: All takes and takes_package are moved together into archived collection. imo, this check is not required.

server/apps/archived/archived.py
```diff
@@ -78,6 +92,129 @@ def on_create(self, docs):
...
+                raise bad_req_error(message="Can't Kill as the Digital Story is still available in production")
+
+            req = ParsedRequest()
+            req.sort = '[("%s", -1)]' % config.VERSION
+            takes_package = list(self.get_from_mongo(req=req, lookup={'item_id': takes_package_id}))
```
mdhaman: Change the query to use elastic. If the item is in elastic then only it will get displayed in the list. Moreover, there will 9M+ documents in the collection and elastic query would be faster and maybe we won't need an index on "item_id".

server/apps/archived/archived.py
```diff
@@ -78,6 +92,129 @@ def on_create(self, docs):
...
+            if not takes_package:
+                raise bad_req_error(message='Digital Story of the article not found in Archived repo')
+
+            takes_package = takes_package[0]
+            if is_item_in_package(takes_package):
+                raise bad_req_error(message="Can't kill as Digital Story is part of a Package")
+
+            for takes_ref in takes_package_service.get_package_refs(takes_package):
+                if takes_ref[RESIDREF] != doc[GUID_FIELD]:
```
mdhaman: An orphaned take should not be there in production. If there are takes in production then there is a BUG in expiry process and needs to be looked into.

sivakuna-aap: Yes, that's the reason kill process should be aborted

server/apps/archived/archived.py
sivakuna-aap: Agreed. But if the process fails in between while moving the takes package then takes package will be in production. Although the takes package eventually gets moved to archived, if user tries to kill a take before the process kicks in then this check will prevent.

server/apps/archived/archived.py
```diff
@@ -88,5 +225,61 @@ def on_fetched_item(self, doc):
...
+        :return: list of articles to be killed
+        :rtype: list
+        """
+
+        archived_doc = self.find_one(req=None, **lookup)
+
+        req = ParsedRequest()
+        req.sort = '[("%s", -1)]' % config.VERSION
+
+        archived_doc = list(self.get_from_mongo(req=req, lookup={'item_id': archived_doc['item_id']}))[0]
```
mdhaman: Query elastic not mongo.

server/apps/archived/archived.py
```diff
@@ -78,6 +92,129 @@ def on_create(self, docs):
...
+                kill_service.queue_transmission(article, subscribers)
+
+            # Step 2(iii)
+            import_into_legal_archive.apply_async(kwargs={'doc': article})
+
+            # Step 2(iv)
+            super().delete({'item_id': article[config.ID_FIELD]})
+
+            # Step 2(i) - Creating entries in published collection
+            get_resource_service('published').post([article])
```
mdhaman: How will this get record removed after expiry? Am I missing something?

server/apps/archived/archived.py
mdhaman: Agree as an fail safe check.

server/apps/archived/archived.py
sivakuna-aap: You mean the killed item in published collection?

mdhaman: yes.

sivakuna-aap: When we post to published collection the `create()` calculates the expiry and eventually gets removed right? But am not sure if the item isn't in archive collection how does it behave?

mdhaman: One of the reason behind the change to the expiry process was to avoid some items in production and some items archived. The `archive` collection is the source and `published` collection has versions of the document in `archive`. Hence, in order to efficiently move the item to `archived` collection process looks at the `archive` collection.

As per the current expiry process we have two options:
- write a new task to delete this type of items.
- create an entry in the `archive` collection and let the current expiry process take care of it.

sivakuna-aap: I ended up creating a new job which takes care of removing the expired articles from published collection.

Also replaced the Mongo queries with elasticsearch queries.

marwoodandrew:
:+1:

server/apps/publish/commands.py
```diff
@@ -56,4 +62,126 @@ def update_overdue_scheduled(self):
...
+                    return False
+
+        if item_id:
+            to_remove.append(item_id)
+
+        return True
+
+
 superdesk.command('publish:remove_overdue_scheduled', UpdateOverdueScheduledPublishedContent())
+superdesk.command('published:remove_killed_if_expired', UpdateOverdueScheduledPublishedContent())
```
mdhaman: Did you meant `RemoveExpiredKilledContent` instead of `UpdateOverdueScheduledPublishedContent`?

mdhaman:
:+1: except for the minor fix.

server/apps/publish/commands.py
sivakuna-aap: Right. Copy/Paste error...

PR #1496, opened 2015-12-03
feat(authoring): SD-3294 Created sticky header on authoring screen

pavlovicnemanja:
SD-3294

PR #1568, opened 2015-12-24
feat(layout): SD-3576, SD-3778 Improved responsive behavior | Removed underscore from state label

pavlovicnemanja:
- [x] SD-3576 Layout issues when authoring, search and search filter panel are open
- [x] SD-3778 Take out the underscore from the "in progress" label
- [x] Fixed comment from PR #1518, added false for $timeout function in splitter directive
- [x] Fixed comment from PR #1496, added debouncing for sticky header function


sivakuna-aap:
:+1:

client/app/scripts/superdesk/ui/ui.js
```diff
@@ -1222,7 +1222,47 @@
...
+                minWidth: '=',
+                maxWidth: '='
+            },
+            link: function (scope, elem) {
+                var window = angular.element($window);
+
+                window.on('resize', _.debounce(calcSize, 300));
+
+                function calcSize() {
+                    $timeout(function () {
```
petrjasek: i guess you don't need to call `$timeout` which would trigger full digest, probably you should call only 
```
scope.$parent.$applyAsync(function() { scope.$parent.elementState = 'compact'; });
```

client/app/scripts/superdesk/filters.js
```diff
@@ -221,5 +221,12 @@
                 var conditionValue = labels.length > 0 ? labels.join(', ') : filterCondition.value;
                 return '(' + filterCondition.field + ' ' + filterCondition.operator + ' ' + conditionValue + ')';
             };
+        })
+        .filter('formatString', function() {
```
petrjasek: can we think about a better name? it doesn't do much formatting.
it's more like making it human readable.. but `h` might be too short :smile:

pavlovicnemanja: Hmmm, maybe `removeLodash` is more appropriate? I'm such a bed author of function and class names :D

ioanpocol:
:+1:

PR #1572, opened 2015-12-30
(fix) [SD-2097] When a locked item is published "Edit and Correct" and "Edit and Kill" should be visible

sivakuna-aap:


akintolga:
:+1:

mdhaman:
:+1:

PR #1576, opened 2015-12-31
fix(reuters image ingestion) token was not passed on rendition requests

marwoodandrew:
This is dependent on superdesk/superdesk-core#118


sivakuna-aap:
:+1:

akintolga:
:+1:

PR #1585, opened 2016-01-04
(fix) [SD-2097] Displaying right buttons when a locked published item is unlocked by someone else

sivakuna-aap:


mdhaman:
:+1:

akintolga:
:+1:

PR #1593, opened 2016-01-06
[SD-3859] - Saved search list is sorted

akintolga:


client/app/scripts/superdesk-monitoring/aggregate.js
```diff
@@ -489,7 +489,7 @@
                     if (scope.privateSavedSearches.length > 0) {
                         scope.privateSavedSearches.length = 0;
                     }
-                    _.each(scope.searches, function(item) {
+                    _.each(_.sortBy(scope.searches, 'name'), function(item) {
```
sivakuna-aap: We are using `sortByName` filter for this kind of functionality. Can we use the same to be consistent with other lists also?

mugurrus:
Some e2e tests need to be updated because of the sort by name

mdhaman:
:+1:

marwoodandrew:
:+1:

sivakuna-aap:
:+1:

PR #1594, opened 2016-01-06
(fix) sd-meta-terms directive

sivakuna-aap:
When user clicks on remove button on any of the selected term using sd-meta-terms directive, then it's being added back to the list. Because of this the same term appears more than once in the list. This issue is fixed now.


mugurrus:
:+1: (just fix the style error)

akintolga:
:+1:

PR #1597, opened 2016-01-07
(fix) [SD-3883] Ticking an ingested item in global search then choosing it to be a package does not work

sivakuna-aap:


akintolga:
:+1:

mugurrus:
:+1:

PR #1599, opened 2016-01-07
(fix) [SD-3882] Scheduled item that get published do not have update or take option present in the action menu

sivakuna-aap:


mdhaman:
:+1:

mugurrus:
:+1:

PR #1601, opened 2016-01-07
fix(templates): fix content creation using template

petrjasek:
- remove template if user saves no changes after creating it.
- use also default desk when creating it so it won't end up
in user personal space.

SD-3431


client/app/scripts/superdesk-workspace/content/content.spec.js
```diff
@@ -71,7 +71,8 @@ describe('superdesk.workspace.content', function() {
                 slugline: 'test_slugline',
                 body_html: 'test_body_html',
                 task: {desk: '2', stage: '4', user: 'user:1'},
-                template: 'template1'
+                template: 'template1',
+                version: 0
```
akintolga: There's a change in requirements. Lasky must have updated the ticket with the change. In brief, the brand new items will be saved as v0 but the items created with some data in them suggested to be v1:
* Packages created by selecting an item
* Updates of published stories
* Stories out of templates

Please take a look at this PR as well:
https://github.com/superdesk/superdesk/pull/1602

mdhaman: And `Takes` as well.

petrjasek: ok this is not affecting items created by system so it should work as you said

ioanpocol:
:+1:

sivakuna-aap:
:+1:

ioanpocol:
:+1:

PR #1605, opened 2016-01-08
(fix) [SD-3890] Highlight button should not disappear from top panel when a published item is selected

sivakuna-aap:
Resolves [SD-3890](https://dev.sourcefabric.org/browse/SD-3890)


mugurrus:
:+1:

mdhaman:
:+1:

PR #1616, opened 2016-01-11
feat(authoring): SD-2512 Implemented case sensitive/insensitive Find Replace operation

pavlovicnemanja:
SD-2512

### DIFF
--- a/client/app/scripts/superdesk/editor/editor.js
+++ b/client/app/scripts/superdesk/editor/editor.js
@@ -285,7 +285,9 @@ function EditorService(spellcheck, $rootScope, $timeout, $q) {
         var index, text;
         while (tree.nextNode()) {
             text = tree.currentNode.textContent;
-            while ((index = text.indexOf(needle)) > -1) {
+            while ((index = self.settings.findreplace.caseSensitive ?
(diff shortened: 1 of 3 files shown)


sivakuna-aap:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk/editor/editor.js
```diff
@@ -285,7 +285,9 @@ function EditorService(spellcheck, $rootScope, $timeout, $q) {
         var index, text;
         while (tree.nextNode()) {
             text = tree.currentNode.textContent;
-            while ((index = text.indexOf(needle)) > -1) {
+            while ((index = self.settings.findreplace.caseSensitive ?
```
petrjasek: because of performance it might be better to do it before the loop, so
```
var needle = caseSensitive ? needle : needle.toLowerCase();
...
while (tree.nextNode()) {
    text = caseSensitive ? tree.currentNode.textContent : tree.currentNode.textContent.toLowerCase();
    while ((index = text.indexOf(needle)) > -1) {
      ...
```
but didn't profile it (yet) so maybe it's not worth it

PR #1620, opened 2016-01-12
(imp) [SD-3035] Adding the ability to edit the kill template text before the kill is sent

sivakuna-aap:


server/apps/archived/archived.py
```diff
@@ -164,36 +164,44 @@ def on_delete(self, doc):
         doc[config.ID_FIELD] = id_field
 
     def delete(self, lookup):
+        if app.testing and len(lookup) == 0:
+            super().delete(lookup)
+            return
+
+    def on_deleted(self, doc):
```
mugurrus: Not needed

server/apps/archived/archived.py
```diff
@@ -164,36 +164,44 @@ def on_delete(self, doc):
...
+        pass
+
+    def update(self, id, updates, original):
```
mugurrus: Why delete became update?

sivakuna-aap: Because the kill now is divided into 2 steps. 1 - check if kill is allowed on the item and if valid apply the kill template on the item and show it in a modal, allowing the user to modify the reasons for killing the item 2 - when user clicks on the button in the modal the update performs the actual kill process.

Also, the delete method ignores the body in request. Although update isn't the right way because of delete's limitation we had to opt for either patch/post.

server/apps/archived/archived.py
sivakuna-aap: OK, I will remove it

server/apps/archived/archived.py
mugurrus: I would create another endpoint (e.g. /archived/<id>/kill) with CREATE method to mark the archived item for kill. Use patch to apply the template and finally delete to complete the kill process. Using DELETE for checking is not very REST compliant. There's also an edge case, where the archived item could change between the moment when you checked and the moment when you apply the template - and the validations for DELETE could fail.

sivakuna-aap: 1. Agreed that validating item to be killed via DELETE isn't REST complaint but for just validation I want to avoid creation of one more endpoint/resource/service.

2. Until user clicks on Kill button as mentioned in the second step we shouldn't make changes to the item in archived.

3. Unlike production the items in archived doesn't change in little time as they make an entry into archived after they expire in production.

akintolga:
:+1:

mdhaman:
:+1:

PR #1627, opened 2016-01-12
fix(highlights): Keep the highlight package on edit mode when a contained item is closed (SD-3580)

ioanpocol:
…


mugurrus:
:+1:

PR #1628, opened 2016-01-12
Editor: Use iframely instead of embedly

vied12:
- [x] Use Iframely instead of embedly as default provider
- [x] Remove Embedly dependency


mugurrus:
:+1:

ioanpocol:
:+1:

PR #1629, opened 2016-01-13
[SD 2124]-Published items should not be allowed to be open in multi edit

sjunaid:


mugurrus:
:+1:

ioanpocol:
:+1:

sivakuna-aap:
@sjunaid am bit late...we still need to handle here also https://github.com/superdesk/superdesk/blob/master/client/app/scripts/superdesk-authoring/views/authoring-topbar.html#L79

PR #1632, opened 2016-01-13
[sd-2124]-(fix - undo wrong cleanup) - Published items should not be …

sjunaid:
…allowed to be open in multi edit


sjunaid:
This check (that was mistakenly removed), should be there for other than text-editor fields to be disable for published items in multi edit

ioanpocol:
:+1:

akintolga:
:+1:

PR #1643, opened 2016-01-14
(feat) [SD-3435]: Add Header info to the preview pane (without editin…

fritzSF:
…g capabilities)


pavlovicnemanja:
:+1:

Aca-jov:
:+1:

sivakuna-aap:
@fritzSF the same information is available in Metadata tab also, so we should remove fields available in Header from Metadata tab

PR #1645, opened 2016-01-14
feat(cvs): show cvs for selected service

petrjasek:
adding additional vocabularies which are displayed according to selected service (category)

### DIFF
--- a/client/app/scripts/superdesk-authoring/metadata/metadata.js
+++ b/client/app/scripts/superdesk-authoring/metadata/metadata.js
@@ -397,16 +397,15 @@ function MetadataListEditingDirective(metadata) {
             unique: '@',
             postprocessing: '&',
             change: '&',
-            header: '@'
+            header: '@',
+            cv: '='
         },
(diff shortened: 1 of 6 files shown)


client/app/scripts/superdesk-authoring/metadata/metadata.js
```diff
@@ -593,16 +616,18 @@ function MetadataService(api, $q) {
...
     var service = {
         values: {},
+        cvs: [],
         subjectScope: null,
         loaded: null,
         fetchMetadataValues: function() {
             var self = this;
 
-            return api('vocabularies').query().then(function(result) {
+            return api.query('vocabularies', {max_results: 50}).then(function(result) {
```
vied12: If there are more than 50 results, don't we need to ask the other pages here ?

vied12:
:+1: I would just avoid to use abbreviations like cvs or cv for directive properties. It's hard to guess the meaning :)

thnkloud9:
:thumbsup:

PR #1646, opened 2016-01-14
Feature/new editor 2

vied12:
* [x] Stuck toolbar
* [x] Ensure toolbar stays properly on screen when scroll down
* [x] Paragraphs in the same text field
* [x] Writing experience must be cool
* [x] Save when formatting changes (not only when text changes)
* [x] Issue with Chrome (focus text input when "add an embed" is clicked)
* [x] Ensure that produced html is valid
* [x] We need to be able to add text above and below embed. And we need to be able to add embed above and below text block
* [x] Key shortcut to open then embed input form (`ctrl` + `e`) and to close it (`escape`)
* [ ] ~~Reorder block~~ (later in another PR)

Changes on __medium-editor__ have been submitted in https://github.com/yabwe/medium-editor/pull/930

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -38,6 +38,8 @@
 
     <!-- build:js scripts/superdesk-core.js -->
     <script src="scripts/superdesk/editor/editor.js"></script>
+    <script src="scripts/superdesk/editor/editor.ctrl.js"></script>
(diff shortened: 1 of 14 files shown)


eozan:
:+1:

thnkloud9:
:thumbsup:

PR #1655, opened 2016-01-18
[sd-3917] - (fix) - I should not be able to Ctrl-z to the original, actionable text when killing an item

sjunaid:
…ctionable text when killing an item

### DIFF
--- a/client/app/scripts/superdesk/editor/editor.js
+++ b/client/app/scripts/superdesk/editor/editor.js
@@ -775,6 +775,12 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck'])
                 }
 
                 function updateModel() {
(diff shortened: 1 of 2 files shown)


client/app/scripts/superdesk/editor/editor.js
```diff
@@ -530,6 +543,12 @@ function EditorService(spellcheck, $rootScope, $timeout, $q) {
      * @param {Scope} scope
      */
     this.commitScope = function(scope) {
+        // In case of Kill action, don't undo back to initial value (i.e: original article text) - [ref: SD-3917]
+        // Set initial value to kill template text at the beginning of history.
+        if (scope.$parent.action === 'kill' && scope.history.getIndex() === -1) {
```
akintolga: with this line it looks like we're creating a dependency in editor service for the story. Maybe commitScope function shouldn't be called in the first place.

ioanpocol:
:+1:

PR #1657, opened 2016-01-18
[SD-2824] Update the user _updated when privileges are revoked

mdhaman:
This  is dependent  on https://github.com/superdesk/superdesk-core/pull/134.

We need to reload the page on revoking privileges and fetch the active privileges. Eve returns 304 if the `_updated` is not modified. Hence, raised an event to update the affected users and then reload the page.


akintolga:
:+1:

marwoodandrew:
:+1:

ioanpocol:
:+1:

PR #1658, opened 2016-01-18
feat(editor): Added custom playbuzz embed.

eozan:
SDPA-64

### DIFF
--- a/client/app/scripts/superdesk/editor/editor.js
+++ b/client/app/scripts/superdesk/editor/editor.js
@@ -1293,17 +1293,25 @@ angular.module('superdesk.editor', ['superdesk.editor.spellcheck', 'angular-embe
             }
         };
     }])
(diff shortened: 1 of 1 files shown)


vied12:
:ok_hand:

pavlovicnemanja:
:+1

mugurrus:
I think it's not a good idea to tightly couple a functionality for a specific client in the superdesk repo. This should be as generic as possible. Even if it's in our repo it should be possible to enable/disable it.

vied12:
of course we will put that in another module later. The SD client is not totally ready for that yet :grin:
I vote to merge it, and I will dissociate the specific code in another PR. Because what is here is not the only one! :bowtie:

hlmnrmr:
:+1: 
... but yeah, let's not forget about this, @mugurrus is right and should be handled in a different way in the near future.

PR #1661, opened 2016-01-19
(fix) Fixing issues related to metadata dropdown.

mdhaman:


mugurrus:
I think a regression test would be good to have for this fix.

ioanpocol:
:+1:  except comment

akintolga:
:+1:

PR #1665, opened 2016-01-19
fix(personal): Items created in personal workspace should not have expiry field present(SD-3877)

ioanpocol:
Also added the option on behave tests to specify that a field should not be present in response.

### DIFF
--- a/server/features/steps/steps.py
+++ b/server/features/steps/steps.py
@@ -88,6 +88,16 @@ def test_key_is_present(key, context, response):
         '"%s" should be empty or false, but it was "%s" in (%s)' % (key, response[key], response)
 
 
+def test_key_is_not_present(key, response):
+    """Test if given key is not present in response.
+
+    :param key
+    :param response
+    """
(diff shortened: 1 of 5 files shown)


server/features/steps/steps.py
```diff
@@ -88,6 +88,16 @@ def test_key_is_present(key, context, response):
         '"%s" should be empty or false, but it was "%s" in (%s)' % (key, response[key], response)
 
 
+def test_key_is_not_present(key, response):
```
akintolga: There's a similar function on this line...
https://github.com/superdesk/superdesk/blob/master/server/features/steps/steps.py#L887

akintolga:
:+1:

mdhaman:
:+1:  except comment for tolga

server/features/steps/steps.py
ioanpocol: The solutions have a kind of overlapping but I think the new solution is more generally.

PR #1668, opened 2016-01-19
On creating item from templates, templates list is refreshed

Aca-jov:
SD-3648

### DIFF
--- a/client/app/scripts/superdesk-workspace/content/content.js
+++ b/client/app/scripts/superdesk-workspace/content/content.js
@@ -126,6 +126,7 @@
                  */
                 scope.createFromTemplate = function(template) {
                     content.createItemFromTemplate(template).then(edit);
+                    scope.reloadTemplatesList = true;
                 };
 
                 /**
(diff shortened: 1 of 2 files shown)


client/app/scripts/superdesk-workspace/content/content.js
```diff
@@ -146,6 +148,16 @@
                     });
                 });
 
+                scope.$watch(function() {
+                    return scope.reloadTemplatesList;
```
akintolga: Just a minor comment: since we know the value of reloadTemplatesList can be set to true only at line 129 then maybe we don't need to create a watch, we can call the function to load directly.

akintolga:
:+1:

mdhaman:
:+1:

client/app/scripts/superdesk-workspace/content/content.js
Aca-jov: I tried that way first, but it can't be done that way, because there are lots of promise calls, so list would be refreshed before new template is added to list. That's why I did it this way..

PR #1669, opened 2016-01-20
fix(desk-mentions) - Problems in updating the unread messages

akintolga:


marwoodandrew:
:+1:

server/apps/comments/__init__.py
```diff
@@ -11,9 +11,13 @@
...
 import superdesk
+from .user_mentions import on_activity_updated
 
 
 def init_app(app):
     endpoint_name = 'comments'
     service = CommentsService(endpoint_name, backend=superdesk.get_backend())
     CommentsResource(endpoint_name, app=app, service=service)
+
+    app.on_updated_activity -= on_activity_updated
```
mugurrus: I would discourage the creation of events on app and use the service API instead. This can get messy, APIs are cleaner.

mdhaman: These hooks are already created by the Eve framework. As an general convention, should be avoid this? 
Moreover, I created an event hook on updating roles when privileges are revoke. If we don't create event hook the core will he calling the `Desks` service api. Is this acceptable?

mdhaman:
:+1:

server/apps/comments/__init__.py
akintolga: I think registering for the existing events is a good way of decoupling. In this instance `on_updated_activity` event is available for us by eve framework.

PR #1670, opened 2016-01-20
[SD-3914] set fetch operation for fetch from AAP multimedia, do not p…

marwoodandrew:
…ass empty dateline from client

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -2147,8 +2147,8 @@
                 /* Creates a copy of dateline object from item.__proto__.dateline */
                 scope.$watch('item', function(item) {
                     if (item) {
-                        var updates = {dateline: {}};
                         if (item.dateline) {
+                            var updates = {dateline: {}};
                             updates.dateline = _.pick(item.dateline, ['source', 'date', 'located', 'text']);
                             if (item.dateline.located) {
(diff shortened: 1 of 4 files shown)


client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -2156,9 +2156,8 @@
                                 scope.datelineDay = monthAndDay.day;
                                 scope.resetNumberOfDays(false);
                             }
+                            _.extend(item, updates);
```
mugurrus: `var updates = {dateline: {}};` can be moved inside the inner if and then the 2 ifs can be combine in a single one.

mugurrus:
:+1: (except comment)

server/apps/search_providers/aap_mm/service.py
```diff
@@ -63,6 +64,7 @@ def create(self, docs, **kwargs):
             dest_doc[ITEM_STATE] = doc.get(ITEM_STATE, CONTENT_STATE.FETCHED)
             dest_doc[INGEST_ID] = archived_doc[config.ID_FIELD]
             dest_doc[FAMILY_ID] = archived_doc[config.ID_FIELD]
+            dest_doc[ITEM_OPERATION] = ITEM_FETCH
```
mdhaman: Do we have a behave test verifying that multimedia content fetch does not have dateline?

akintolga:
:+1:

PR #1674, opened 2016-01-20
fix(authoring): Fixed removing associations.

eozan:
SDPA-113


petrjasek:
would it be possible to remove this without opening the modal?
like just adding a cross into top right corner in the preview?

eozan:
possible, let's discuss this with @fritzSF

vied12:
+1 for the cross ✖️

eozan:
I'm closing this pull request as it wil be done from scratch. (Work needed is totally different for modal and cross solutions). Let's include important details like this in the task description next time.

PR #1679, opened 2016-01-21
On firefox scroll fixed

Aca-jov:
SD-3927, SD-3922, SD-3933

### DIFF
--- a/client/app/scripts/superdesk-monitoring/monitoring.js
+++ b/client/app/scripts/superdesk-monitoring/monitoring.js
@@ -452,7 +452,8 @@
                     queryTimeout = $timeout(queryItems, 50, false);
                 }
 
-                var list = elem[0].getElementsByClassName('inline-content-items')[0],
+                var list = elem[0].getElementsByClassName('stage-content')[0],
+                    listContent = elem[0].getElementsByClassName('inline-content-items')[0],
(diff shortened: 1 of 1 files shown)


client/app/scripts/superdesk-monitoring/monitoring.js
```diff
@@ -546,7 +546,6 @@
                                 list.style.height = (scope.total * ITEM_HEIGHT) + 'px';
                             }
 
-                            list.style.paddingTop = (from * ITEM_HEIGHT) + 'px';
```
Aca-jov: @petrjasek Can you please check this, you worked on it. This padding top was problem, but I'm not sure why is this padding for, maybe I broke something somewhere else..

petrjasek: hey, I think it's needed.. if you have longer list and you scroll further down this should position the items properly. without having the scroll bar jumping.

Aca-jov: Ah, ok, I didn't notice that. I fixed it with that, so now it's fine. Thanks

pavlovicnemanja:
:+1:

ioanpocol:
:+1:

PR #1683, opened 2016-01-22
feat(ng-animate): Added support for angular-animate library

pavlovicnemanja:
After this, bower update is needed.

Requested by @fritzSF , for new and fancy animations! :)


mugurrus:
:+1:

mdhaman:
:+1:

PR #1687, opened 2016-01-25
(fix) [SD-3510] Improve formatting of related items in the editor

fritzSF:


client/app/scripts/superdesk/itemList/views/relatedItem-list-widget.html
```diff
@@ -31,12 +31,12 @@
 
     </div>
 
-    
+
    <!--  <div class="filter-box" ng-show="options.mode === 'detailed'">
         <span class="text" translate>View</span>
-        <button class="btn btn-light pull-left" 
+        <button class="btn btn-light pull-left"
```
pavlovicnemanja: Hey, why we left the commented part inside file? Do we need it for something?

fritzSF: i don't think so, the Ingest and Content widgets will be deprecated anyway

pavlovicnemanja: Maybe we can remove it, to keep code clean. This part of code was commented on April 7, 2015.

petrjasek:
:+1:

akintolga:
:+1:

client/app/scripts/superdesk/itemList/views/relatedItem-list-widget.html
fritzSF: Sure, remove it.

PR #1691, opened 2016-01-27
move away client files

petrjasek:
atm it's in a folder but to make use of it you have to npm install

requires won't change once this folder is in it's own repo, but atm
it's simpler to work with it this way.

- [x] specify dependencies
- [x] move specs
- [x] move grunt config
- [x] move styles
- [x] move fonts
- [x] move images
- [x] move scripts


petrjasek:
ping @superdesk/developers

mugurrus:
I think there's no point in reviewing this, we can't see the diff anyway. So I suggest to rebase and merge it.

petrjasek:
you are right, I will merge once it's green

PR #1715, opened 2016-02-03
chore(server): update gunicorn

petrjasek:
it supports new params in access log which makes it more readable


akintolga:
:+1:

ioanpocol:
:+1:

PR #1717, opened 2016-02-03
chore(core): update core versions

petrjasek:


petrjasek:
needs specs fix https://github.com/superdesk/superdesk-client-core/pull/9

PR #1724, opened 2016-02-04
SD-3067 Publish stories asynchronously

mugurrus:
Requires https://github.com/superdesk/superdesk-core/pull/151


petrjasek:
this code is there already

PR #1727, opened 2016-02-05
Update README.md & scripts

takeit:
Replaced `$(readlink -e $(dirname "$0"))` with `$(echo $(cd -P -- "$(dirname -- "$0")" && pwd -P))"` so we can run `scripts/docker-local-create-user.sh` and `scripts/docker-local-demo.sh` on OSX. (readlink with -e param causes some issues on mac)

When running Superdesk on Docker from OSX, Docker Toolbox needs to be installed. I may create the install guide on how to install superdesk locally and on docker using OSX, but that will go into separate PR.

### DIFF
--- a/docker/container_build.sh
+++ b/docker/container_build.sh
@@ -24,7 +24,7 @@ RUN_FRONTEND_UNIT=${bamboo_RUN_FRONTEND_UNIT:=1}
 RUN_E2E=${bamboo_RUN_E2E:=1}
 
 # install script requirements
-virtualenv -p python2 $SCRIPT_DIR/env
+virtualenv $SCRIPT_DIR/env
 set +u
 . $SCRIPT_DIR/env/bin/activate
 set -u
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Use [docker-compose](http://docs.docker.com/compose/ "") and the config from the
 ##### install system-wide dependencies
 
 ```sh
-$ sudo apt-get install python2.7 python-virtualenv git
+$ sudo apt-get install python-virtualenv git
 ```
 and install [the newest docker](https://docs.docker.com/installation/).
 and make sure you can run [docker without sudo](http://askubuntu.com/questions/477551/how-can-i-use-docker-without-sudo).
--- a/scripts/docker-local-create-user.sh
+++ b/scripts/docker-local-create-user.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
-SCRIPT_DIR=$(readlink -e $(dirname "$0"))
-(test -d $SCRIPT_DIR/env || virtualenv -p python2 $SCRIPT_DIR/env ) &&
+SCRIPT_DIR="$(echo $(cd -P -- "$(dirname -- "$0")" && pwd -P))"
+(test -d $SCRIPT_DIR/env || virtualenv $SCRIPT_DIR/env ) &&
 . $SCRIPT_DIR/env/bin/activate
 set -ue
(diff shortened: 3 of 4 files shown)


petrjasek:
too bad you broke e2e tests ;)
but are you sure it should be python3? i think it might have issues on some python3 versions, but python 2.7 should always work

takeit:
haha my bad ;)

well, I had some issues installing it on python 2.7. I will recheck again and provide more details. I think you should specify on which versions of python it should work, for instance it works on 3.5.1 but it doesn't on 3.5 etc. (maybe test against different python versions on travis?)

Also here https://github.com/superdesk/superdesk/tree/master/server#requirements it's written you support python 3.3+ but then in the main readme you use python 2.7, this is really confusing as at the end I dont know on which version it should work :)

ahilles107:
i can't run Superdesk on 2.7 for me it doesn't work also with 3.5.0 - i had upgrade python to 3.5.1 to run latest code.

petrjasek:
we're mixing 2 things here - superdesk runs on python3 indeed, but that's something you setup in virtualenv. what this docs is refering to is installation of tools to be able to create such environment.
and something where you can run docker-compose which afaik runs on python2.7. so probably the correct thing would be have there both python2 and python3 (to be able to create virtualenv for python3).

takeit:
@petrjasek that makes sense now. 

But still in this case, you need to install two versions of python: 
- 2.7 to create an environment
- 3 (3.3+)

I see that from version 1.5.0 of docker-compose it works with Python 3: `If you install Compose with pip or use it as a library, it now works with Python 3.` (https://github.com/docker/compose/releases/tag/1.5.0)

Additionally I tried to run superdesk on installed python 3 only (I removed python 2 totally).
I changed `python2` to `python3` in `scripts/docker-local-demo.sh` and ran it. All works properly on docker. I managed to run it.

I would get rid of installing python2 in this case as its uneeded dependency which you need to install just to set up env (which can be set with python 3 too) or am I missing something?

python2 is used only in some scripts (https://github.com/superdesk/superdesk/search?utf8=%E2%9C%93&q=python2) which can run using python3 as well.

/cc @actionless

actionless:
we can replace `python2.7` to just `python`

we don't need `2` since docker-compose supports `3`, thanks for noticing that

takeit:
Okay, PR is updated now

docker/container_build.sh
```diff
@@ -24,7 +24,7 @@ RUN_FRONTEND_UNIT=${bamboo_RUN_FRONTEND_UNIT:=1}
 RUN_E2E=${bamboo_RUN_E2E:=1}
 
 # install script requirements
-virtualenv -p python2 $SCRIPT_DIR/env
+virtualenv -p python $SCRIPT_DIR/env
```
actionless: you can replace it to just
```
virtualenv $SCRIPT_DIR/env
```

actionless:
:+1: 
please rebase and squash

petrjasek:
then we shouldn't install `python-virtualenv` but use `python3 -m venv` - it's not built in

actionless:
it was added in python 3.3

is any older python is still shipped in any popular currently supported distros?
if so it would be better to leave it as is because python-virtualenv still works for those newer python versions

petrjasek:
not sure which these are.. but ubuntu 14.04 lts and debian stable should be fine, both have python3.4.
so we could try with python3 only

actionless:
i've found:
debian 7 (supported until 2018, python3.2)
ubuntu 12.04 (supported until end of 2017, python3.2)

petrjasek:
ok then let's keep it as it is

PR #1729, opened 2016-02-05
added expected argument to -a option

takeit:
added `true` to `-a/--admin` option as it expects one argument

```
usage: manage.py users:create [-?] --username USERNAME --password PASSWORD
                              --email EMAIL [--admin ADMIN]
manage.py users:create: error: argument --admin/-a: expected one argument
```

btw I think its a bug in the command as it should not accept any argument, it should create an admin account when u put `-a` or `--admin`

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -104,5 +104,5 @@ This will give you list of available commands.
 This command will create an administrator user.
 
 ```sh
-$ python manage.py users:create -u <username> -p <password> -e <email> -a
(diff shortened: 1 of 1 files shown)


actionless:
@takeit please close this one if a linked patch works for you

takeit:
@actionless it doesn't work, still:

```
manage.py users:create: error: argument --admin/-a: expected one argument
```

actionless:
now it should be fine (i tested :) )

takeit:
:+1:

PR #1730, opened 2016-02-05
feat: generate installation docs

actionless:


petrjasek:
imo we will have to write this install docs manually, some steps in docker makes not much sense when you would install locally (like copy requirements files, install it in tmp folder, bower install with root).
also it doesn't mention required services (mongo/elastic..).
any other ideas? ping @superdesk/developers

actionless:
@petrjasek 

with manually-written docs we will lose the possibility to have ci-verified install guide

`--allow-root` arg does not require to run bower by root

and regarding documenting required services and (that's more important) environment variables -- i think we can come up with the similar auto-generated docs but not only for Dockerfile but docker-compose.yml too

petrjasek:
I know. we should probably figure out who are we targeting with this install docs.
so what's the idea? is this supposed to be for developers or end users?

actionless:
that's a tricky question -- manual docs are usually demanded by people who can't or don't want to use docker but can't understand Dockerfile's syntax

Michael-Gourlay:
I would guess that developers would be the main audience for a docker installation guide. Although some end users might deploy Superdesk using docker any production use at scale is unlikely to do so. Just my 2 cents guys.

actionless:
i hardly can imagine production sys.admin who will be confused understanding this syntax  https://github.com/superdesk/superdesk/blob/master/Dockerfile

i would suggest finding out list of issue with this revision of autogenerated install doc and starting to find a way to address them without writing docs manually (== having one more component to maintain)

however for stable releases writing docs could make sense if the quality of autogenerated doc will be still far from perfect

petrjasek:
but that's the thing, are we targeting sys admins with that? because if not we can have simpler docs with only steps that are required..

actionless:
for "normal users" (what does it mean btw, like a manager demoing superdesk from a laptop?) i don't see a point of not using a docker because it's reducing the possible failure reason to the only one service, ie the only one thing which could not work will be docker itself while there are plenty of things which could go wrong on the manual installation

ahilles107:
IMO it's wrong direction - we need other kind of docs here.

actionless:
1) who you gonna target with those docs
2) who will be updating docs to follow each superdesk change
3) how are you gonna track what docs became out-of-date
4) how are you going to verify those docs (ie what following those steps leads to some defined result)

ahilles107:
1) developers who don't use docker and prefer install superdesk manually on their laptop/server
2) deveopers who make that change (there shouldn't be feature/bc break/configuration PR without docs) 
3) look at point 2
4) q/a, people actually using this docs 

Recreating docker instructions gives us false sense of solving that issue.

actionless:
1) i can not imagine a developer, whatever language he is using in superdesk, js or python, who will be not possible to understand the syntax of that file (https://github.com/superdesk/superdesk/blob/master/Dockerfile). And even if it's possible, then what's wrong with those current auto-generated docs? I don't see any mistakes or misleading in them.
2,3) so your proposal instead of updating docs automatically to force other people doing that?
4) and more people to verify that..

are you gonna pay for it from your own pocket? or you are going to convince the bosses pay to satisfy marginal complains of docker haters

petrjasek:
I like the idea of automatically testing install docs. but this is not it.
basically I don't see any added value, there is no more information than in the docker file. so imo we can just add a link to dockerfile from the install docs and put there some extra info (if any).

actionless:
@petrjasek  that was the initial intention and it was like that for some decent amount of time but now @ahilles107 comes around and saying what devs can't understand those files' syntax so've tried to found some solutions for those people who can't understand in the context that `COPY` and `RUN` commands mean

petrjasek:
I see, well that's not it obviously. so closing this one for time being

PR #1734, opened 2016-02-09
[SD-3971] text archive import fix

marwoodandrew:


server/aap/import_text_archive/commands.py
```diff
@@ -63,16 +69,24 @@ class AppImportTextArchiveCommand(superdesk.Command):
...
+
+    def run(self, start_id, user, password, url, query, limit, direction):
         print('Starting text archive import at {}'.format(start_id))
         self._user = user
         self._password = password
         self._id = int(start_id)
         self._url_root = url
         self._query = urllib.parse.quote(query)
+        # direction True is forwards
+        self._direction = True
```
mugurrus: I would name this _direction_forward, as it is used in other places too. _direction suggest it's values can be 'forward', 'backward'

mugurrus:
:+1:

mdhaman:
:+1:

PR #1736, opened 2016-02-10
chore(client): update core

petrjasek:


petrjasek:
this will only work after https://github.com/superdesk/superdesk-client-core/pull/22 is merged

PR #1744, opened 2016-02-11
fix(workqueue): When an item is opened for correction it will show up on the bottom panel (SD-3556)

ioanpocol:


petrjasek:
rebase pls

pavlovicnemanja:
:+1:

PR #1747, opened 2016-02-12
feat(authoring): Implemented opening full-preview in new tab.

eozan:
SDPA-210

### DIFF
--- a/client/app/scripts/superdesk-authoring/authoring.js
+++ b/client/app/scripts/superdesk-authoring/authoring.js
@@ -903,7 +903,7 @@
                 });
 
                 $scope.fullPreview = false;
-                $scope.fullPreviewItem = null;
+                $scope.fullPreviewUrl = '/#/preview/' + $scope.origItem._id;
                 $scope.proofread = false;
                 $scope.referrerUrl = referrer.getReferrerUrl();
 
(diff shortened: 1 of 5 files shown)


mugurrus:
:+1: (e2e test would be nice)

client/app/scripts/superdesk-authoring/authoring.js
```diff
@@ -2518,6 +2533,26 @@
...
+    function FullPreviewDirective(api) {
+        return {
+            scope: {
+                item: '=',
+                itemId: '@',
+                closeAction: '='
+            },
+            templateUrl: 'scripts/superdesk-authoring/views/full-preview.html',
+            link: function(scope, elem) {
+                if (scope.itemId) {
```
petrjasek: is there any use for this one?

petrjasek:
in full preview there are a few strings without translate annotation, otherwise looks ok

client/app/scripts/superdesk-authoring/authoring.js
eozan: It makes it possible to preview with only id, but currently not used.

eozan:
Translations are fixed.

vied12:
:+1:

PR #1748, opened 2016-02-12
feat(app_init): Prepopulate deata for services, categories, genre, lo…

ioanpocol:
…cations and subjects


mugurrus:
:+1:

petrjasek:
I'm not sure we should have service reference per vocabulary item. it seems too granular. but if we go for this I wouldn't keep the service on vocabulary, there is no need to do this in 2 ways.

ioanpocol:
Services on vocabulary is not needed. When there are no options on vocabulary items we can hide the dropdown. I just wanted to keep the compatibility with pa-poc files.
But the current po-poc implementation when you select 2 services you will have multiple times territory dropdown, but you want to have different dropdowns for sports topic and news topic.
Probably both solutions are needed in order to have a more flexible solution (but still not sure if it is flexible enough).

petrjasek:
ok lets keep both then :+1:

PR #1749, opened 2016-02-15
Fix/translation path

vladnicoara:


vladnicoara:
Requires this to pass the tests https://github.com/superdesk/superdesk-client-core/pull/32

vied12:
:+1:

petrjasek:
:+1:

mugurrus:
:+1:

petrjasek:
without that `de` string you will have to update some e2e tests testing language picker..

vladnicoara:
@petrjasek that is why I mentioned that it needs this https://github.com/superdesk/superdesk-client-core/pull/32 to pass the tests. The file with the 'de' string moved into the superesk-clien-core repo. 
Locally it passes that test without issues. I'll check again

petrjasek:
ah right sorry @vladnicoara my bad
can you maybe update the client reference then? it should pass right?

vladnicoara:
@petrjasek right, I missed that :) rebased now. It should pass now here as well. Thanks!

PR #1754, opened 2016-02-16
SDPA - Editor imporvements after feedback

vied12:
- [x] upload and embed buttons are enabled when block is empty
- [x] shows the embed caption properly without `<br>` tags
- [x] + bar to add a new embed between 2 block has been removed, but the drop zone remains
- [x] Fix issue with duplicated extracted text


eozan:
:+1:

petrjasek:
:+1: btw `trans` works same like `translate` ?

vied12:
I don't know actually. I just renamed `trans` to `translate`

PR #1755, opened 2016-02-17
[SD-4048] - Generic macro command

akintolga:


akintolga:
@petrjasek Macro files appear in both superdesk and superdesk-core. Was that because we need to preserve the relative location of the folder to load them dynamically?

akintolga:
Depends on: https://github.com/superdesk/superdesk-core/pull/168

petrjasek:
right I've removed this in https://github.com/superdesk/superdesk/pull/1757

akintolga:
This is not needed anymore.

PR #1760, opened 2016-02-17
fix(authoring): Fixed topics in preview.

eozan:
SDPA-239

### DIFF
--- a/client/app/scripts/superdesk-authoring/views/full-preview.html
+++ b/client/app/scripts/superdesk-authoring/views/full-preview.html
@@ -23,7 +23,7 @@
                         <label translate>Topic</label>
                     </div>
                     <div class="preview-data">
-                        <span ng-repeat="topic in item.subject">
+                        <span ng-repeat="topic in item.subject|filter:'patopic'">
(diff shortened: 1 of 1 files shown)


petrjasek:
:+1:

client/app/scripts/superdesk-authoring/views/full-preview.html
```diff
@@ -23,7 +23,7 @@
                         <label translate>Topic</label>
                     </div>
                     <div class="preview-data">
-                        <span ng-repeat="topic in item.subject">
+                        <span ng-repeat="topic in item.subject|filter:'patopic'">
```
actionless: i am not sure if that's ok to hardcode pa-related stuff on this branch

mugurrus:
Agree with Yauhen

eozan:
I agree that this solution is not ideal, but actual fix will be done next sprint with other improvements (We need to add some flags to vocabularies and UI counterparts, etc.). In the meantime, this needs to be fixed quickly. I'm merging this with approval from @hlmnrmr .

PR #1771, opened 2016-02-22
Fix test javascript status code check.

naomiaro:


vied12:
:+1: and could be merged everywhere

hlmnrmr:
:+1:

petrjasek:
this was fixed in core already https://github.com/superdesk/superdesk-client-core/blob/merge-pa/spec/helpers/backend.js#L24

PR #1785, opened 2016-02-26
Updated vocabularies and content types for using urgency field.

eozan:
SDPA-76
SDPA-77

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -97,3 +97,5 @@ INSTALLED_APPS = [
     'pa.topics',
     'pa.pa_img',
 ]
+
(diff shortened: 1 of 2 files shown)


vied12:
:+1:

server/settings.py
```diff
@@ -97,3 +97,5 @@
     'pa.topics',
     'pa.pa_img',
 ]
+
+DEFAULT_URGENCY_VALUE_FOR_MANUAL_ARTICLES = None
```
actionless: should be wraped inside `env` function, see above in this module

petrjasek: it shouldn't, it's client specific config

actionless:
:+1: then

mugurrus:
:+1:

PR #1786, opened 2016-02-26
Table in editor [SDNTB-101]

vied12:


mugurrus:
You need to update e2e tests

vied12:
@mugurrus I don't get why, I only added a button in the editor toolbar.
Does the errors in e2e tests can be related to the other changes in superdesk-client-core ?
https://github.com/superdesk/superdesk-client-core/compare/749f9de...903efb7

mugurrus:
The errors happen on selection of certain GUI elements so you should check the corresponding tests and see what changed that broke the elements selection.

vied12:
Actually even the base branch doesn't pass the test
https://travis-ci.org/superdesk/superdesk/builds/112006969

ioanpocol:
@vied12  Please make the PR against ntb.
For core PRs please use master branches.
Sorry for late notice.

PR #1795, opened 2016-03-01
chore(core): Check master with latests cores

ioanpocol:


petrjasek:
replaced by https://github.com/superdesk/superdesk/pull/1797

PR #1799, opened 2016-03-02
SDPA-264 : Change 'slugline' back to 'keyword' and place under 'topic'

naomiaro:


petrjasek:
I think you have to also add `app-templates-cache.js` into index.html, it's not there yet. without this it won't work when you run grunt build

naomiaro:
ah ok I'll try that, didn't see that this was failing.

naomiaro:
<script src="templates-cache.js"></script>

this is currently in index.html

petrjasek:
it gets created when you run grunt build. also app-templates-cache.js, and that's missing in index.html, should go after templates-cache.js. btw when you do grunt build it will create those files in dist folder

vied12:
:+1:

petrjasek:
:+1:

PR #1805, opened 2016-03-04
Table in editor [SDNTB-101]

vied12:
new #1786, based on `ntb` instead of `ntbnew`


ioanpocol:
I think that you have to update the dependencies to cores

ioanpocol:
:+1:

PR #1831, opened 2016-03-11
fix(aap multimedia) a change in the core broke it.

marwoodandrew:
This should fix the failing test in superdesk/superdesk#1822


petrjasek:
:+1:

PR #1833, opened 2016-03-11
SDPA-139 : Superdesk doesn't pull in all images from the PA Images API

naomiaro:


mugurrus:
I would recommend some unit and behave tests here

PR #1837, opened 2016-03-15
feat(sign-off): Map signoff and add edit signoff for item(SDNTB-24, SDNTB-25)

ioanpocol:


petrjasek:
those changes in server/app.py are not needed, that boilerplate was moved into core

ioanpocol:
Fixed app.py.

petrjasek:
:+1:

PR #1838, opened 2016-03-15
fix(travis): avoid parallel builds on superdesk

petrjasek:
same reasons like on [client-core](https://github.com/superdesk/superdesk-client-core/pull/176)


akintolga:
:+1:

petrjasek:
same like https://github.com/superdesk/superdesk/pull/1846

PR #1839, opened 2016-03-15
SDPA-330 : Pull in metadata from image file for editing

naomiaro:


ahilles107:
where exif.js is used?

naomiaro:
@ahilles107 it's in client core which isn't merged yet

https://github.com/superdesk/superdesk-client-core/pull/177

ahilles107:
@petrjasek this is where subsplit would help so much. Now you need to follow 2PR for one change (and one of them looks like invalid).

naomiaro:
yeah I agree with this 2 PR thing being confusing, also I don't like how related changed can't be merged together.

petrjasek:
I'm not sure how it would work with multiple clients, any ideas? because here it's client specific version

ahilles107:
hmm... so it's only for AAP? It's not clear as it goes to `superdesk/superdesk` branch master. So it looks like default client + client core version.

petrjasek:
this is pa's branch, will be eventually superdesk-pa repository

naomiaro:
@petrjasek but if this library is being used in client core, it will need to be included in all client branches.

petrjasek:
but we will update master too eventually

petrjasek:
sure @naomiaro

ahilles107:
ahh i missed that branch (hat to check it in othere PR). I have no idea how to organize subtrees with this case (many clients and one core). From my point of view it's totally wrong - as we will build separate app foe every client. But it's not connected to this PR.

petrjasek:
yep exactly

PR #1841, opened 2016-03-16
SDNTB-97 Support for localization

vladnicoara:
updated the client-core referente to point to the translation file that
has the right languages


petrjasek:
you will have to add
```
<script src="node_modules/superdesk-core/scripts/superdesk-search/react.js"></script>
```
to index.html

vladnicoara:
@petrjasek Thanks! doing it now

ioanpocol:
Already updated core dependencies on PR https://github.com/superdesk/superdesk/pull/1837

PR #1847, opened 2016-03-17
Fixed topics in preview.

eozan:


petrjasek:
why another template here? we shouldn't overuse this. it should be last resort

eozan:
Changed from template overriding to config separation.

PR #1849, opened 2016-03-17
Fix for service preferences.

eozan:


client/app/index.html
```diff
@@ -171,7 +171,7 @@
     <script src="node_modules/superdesk-core/scripts/superdesk-authoring/versioning/versioning.js"></script>
     <script src="node_modules/superdesk-core/scripts/superdesk-authoring/versioning/versions/versions.js"></script>
     <script src="node_modules/superdesk-core/scripts/superdesk-authoring/versioning/history/history.js"></script>
-    <script src="node_modules/superdesk-core/scripts/superdesk-authoring/metadata/metadata.js"></script>
+    <script src="scripts/superdesk-authoring/metadata/metadata.js"></script>
```
petrjasek: this should be probably `app/scripts/...`

petrjasek: or does it work locally for you?

petrjasek: when using grunt build?

eozan: it works locally for me, but changed the path to app/scripts/..., and it still works, grunt build gives no errors.

petrjasek: it won't return error, warning maybe. so does it work?

PR #1850, opened 2016-03-18
disable aap macros

petrjasek:
it will make macros written for aap not available for pa

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -14,7 +14,7 @@ import os
 
 APPLICATION_NAME = 'Superdesk'
 
-MACROS_MODULE = 'superdesk.macros'
+MACROS_MODULE = 'macros'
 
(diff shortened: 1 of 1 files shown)


eozan:
:+1:

ahilles107:
Any context? Ticket? Description?

petrjasek:
@ahilles107 I don't know what to add. it will make macros written for aap not available for pa. there is no ticket. it does what it says

ahilles107:
I write this because it's really hard for someone outside main development scope to follow and understand changes. 

Just `it will make macros written for aap not available for pa.` would be enough. In this case it's probably not mandatory as pa-poc branch should be private repository. But most of PR's in superdesk repos have this same issue. And best example goes from top ;)

petrjasek:
fair enough :smile:

akintolga:
:+1:

PR #1860, opened 2016-03-22
fix(upload): add exif lib to index

petrjasek:
it's used to get metadata of an image when uploading


ahilles107:
I assume that library is used in `superdesk/client-core`?

naomiaro:
:+1:  yep it's used in client core @ahilles107

PR #1871, opened 2016-03-25
fix(content type): change the definition of editor settiong on conten…

ioanpocol:
…t types

The schema section from content type json is used as it is as data validator for Cerberus.
I created a new section named 'editor' where it will be all setting related to field editor.


Aca-jov:
:+1:

PR #1874, opened 2016-03-29
cv: thumbnail are no longer first in the renditions list

vied12:
Because now order matters for default rendition, it's important to put a better resolution first


ioanpocol:
:+1:

PR #1883, opened 2016-03-31
Content types updated for header changes.

eozan:
SDPA-355


eozan:
Should be merged together with https://github.com/superdesk/superdesk-client-core/pull/246

server/data/content_types.json
```diff
@@ -5,11 +5,12 @@
         "priority": 80,
         "enabled": true,
         "schema": {
-            "slugline": {"required": true},
-            "body_html": {"required": true},
-            "anpa_category": {},
-            "subject": {},
-            "ednote": {}
+            "anpa_category": {"order": 2, "sdWidth": "full"},
```
ioanpocol: This schema is sent to eve validator as it is and sdWidth is not supported.
Please change the definition of content types as in the following PR:
https://github.com/superdesk/superdesk/pull/1876/files
For all editor related settings there is created a new section called editor. All that is related to data validation remains on schema. The supported values supported for schema are detailed here:
http://docs.python-cerberus.org/en/stable/usage.html#validation-rules

ioanpocol:
:+1:

PR #1889, opened 2016-04-01
Fixed ranking.

eozan:


naomiaro:
Again I'd find a way to deal with this in CSS or we'll have too much garbage going on.

naomiaro:
https://github.com/superdesk/superdesk-client-core/pull/260/files

PR #1892, opened 2016-04-01
Set PA date time types for publish/embargo input.

naomiaro:


client/app/index.html
```diff
@@ -246,6 +246,12 @@
 
         config.previewSubjectFilterKey = 'patopic';
 
+        //date formats for publish/embargo entry
+        config.view = {
+            dateformat: 'D/M/YYYY',
+            timeformat: 'H:mm'
```
actionless: `HH:mm`?

client/app/index.html
```diff
@@ -246,6 +246,12 @@
 
         config.previewSubjectFilterKey = 'patopic';
 
+        //date formats for publish/embargo entry
+        config.view = {
+            dateformat: 'D/M/YYYY',
```
actionless: it was `MM/DD/YYYY` by default before

naomiaro: yeah, this time format is for input only in the angular view, the angular model reformats it to server format. The singular just allows them to type 4/4/2016 to save time 9:00 etc...

naomiaro: It's moment js format strings.

naomiaro: The model formats are

`model: {
       dateformat: 'DD/MM/YYYY',
       timeformat: 'HH:mm:ss'
},`

mugurrus:
:+1:

PR #1894, opened 2016-04-04
fix(data init): Reconfigure desks and their stages, update prepopulat…

ioanpocol:
…ion script for desks (SDNTB-152)


Aca-jov:
:+1:

PR #1895, opened 2016-04-04
Replaces `values` attribute by `qcode`

hlmnrmr:
Related issue https://dev.sourcefabric.org/browse/SDPA-368

Mainstream updates to controlled vocabularies expect a `qcode` attribute instead of `value`


petrjasek:
:+1:

PR #1902, opened 2016-04-05
SDNTB-135 Rename SERVICE back to CATEGORY

vladnicoara:
when https://github.com/superdesk/superdesk-client-core/pull/272 is
merged, the CATEGORY label will be changed to SERVICE( there is another
'Category' label in that header and it's confusing) so this is to make
sure that everything looks the same for AAP


petrjasek:
:+1:

mdhaman:
:+1:

PR #1905, opened 2016-04-05
Listing all cvs in content types to make sure they show up in PA.

naomiaro:
Service metadata needs to be listed here to show up now. It's all lumped together in each content type. I'm not sure of the ordering required. There doesn't exist a good way to make metadata marked for only a particular service to be required when needed.


ioanpocol:
:+1:

mugurrus:
:+1:

PR #1922, opened 2016-04-07
SDNTB-160 Localization: custom button in the editor

vladnicoara:
Renaming not needed as the main CATEGORY button was changed to SERVICE


ioanpocol:
:+1:

ioanpocol:
I think we can delete all those translations.

PR #1934, opened 2016-04-11
Ranking field removed.

eozan:
SDPA-371

### DIFF
--- a/server/data/vocabularies.json
+++ b/server/data/vocabularies.json
@@ -55,8 +55,8 @@
         ]
     },
     {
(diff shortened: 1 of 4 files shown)


server/data/vocabularies.json
```diff
@@ -55,8 +55,8 @@
         ]
     },
     {
-        "_id": "ranking",
-        "display_name": "Ranking",
+        "_id": "urgency",
+        "display_name": "Urgency",
```
hlmnrmr: display_name should be "Ranking"

PR #1939, opened 2016-04-12
Ranking field is set to be displayed only on selected services.

eozan:
SDPA-338


mugurrus:
:+1:

eozan:
Should be merged with https://github.com/superdesk/superdesk-client-core/pull/308

PR #1947, opened 2016-04-14
Timeformat fixed.

eozan:
SDPA-394


eozan:
This is a possible solution for SDPA-394, but because I'm unable to reproduce the problem, I can't verify. Let's keep it WIP for now.

naomiaro:
This shouldn't be changed

PR #1953, opened 2016-04-15
fix(data): setup product for subscribers

petrjasek:
this will require https://github.com/superdesk/superdesk-core/pull/300 to work


ioanpocol:
:+1:

PR #1963, opened 2016-04-18
feat(macros): Integrate a custom ntb macro.

ioanpocol:
The example ntb macro is not final yet, the integration is done

### DIFF
--- a/client/app/index.html
+++ b/client/app/index.html
@@ -260,6 +260,13 @@
                 'scheduled': true
         };
 
+       // enable ingest/content view in workspace sidebar
+       config.workspace = {
+            ingest: 0,
+            content: 0
+        };
+
+
         var modules = [
             'superdesk',
             'superdesk.settings',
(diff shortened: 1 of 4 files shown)


petrjasek:
:+1:

trondhuso:
So this is what it shall look like:
Valutakurser 15.4 (14.4)ID: 12276300
Representative markedskurser for valuta fra Norges Bank

US dollar (USD)	8,2457	(8,2332)
EURO (EUR)	9,3044	(9,2640)
Sveitsiske franc (CHF)	852,13	(851,63)
Danske kroner (DKK)	125,04	(124,49)
Britiske pund (GBP)	11,693	(11,644)
Svenske kroner (SEK)	101,24	(101,16)
Canadiske dollar (CAD)	6,4014	(6,4182)
Japanske yen (JPY)	7,5701	(7,5262)

What I've done is to store the values in an array. The next task is to output the values in the array into a table like the one above. 
Which is what I have not accomplished, and here I need som assistance.

ioanpocol:
After the macro is applied the editor is not refreshed.
It will be fixed by the following task:
https://dev.sourcefabric.org/browse/SD-4394

PR #1974, opened 2016-04-21
fix(server: data: subscribers): dont hardcode assets url

actionless:


vied12:
how does that work? how should we define `#ENV_PUBLICAPI_URL#` ?

actionless:
@vied12 
https://github.com/superdesk/superdesk-core/blob/4d573e12fc4dde1c333d2f7def76a5ecaf0795d0/apps/prepopulate/app_initialization_test.py#L63-L77

mugurrus:
:+1:

PR #1977, opened 2016-04-22
SDPA-421 : Update deployment scripts to set desk stages properly

naomiaro:
Fixing up desks and stages in prepop json to create draft/working stages in the style as seen in master branch json files.


ioanpocol:
:+1:

PR #1989, opened 2016-04-27
doc: added user guides for installing client & server

gbbr:
Addresses [issue SD-4309](https://dev.sourcefabric.org/browse/SD-4309).


README.md
```diff
@@ -19,11 +19,42 @@ different technologies.
...
 
-### Installation
+## Installation
+
+### Client
+
+1. Clone the repository
+2. Navigate to the folder where you've cloned this repository (if it's the main repo, go inside the `client` folder).
+2. Run `npm install && bower install` to install dependencies.
+3. Run `grunt server` to run the web server.
```
petrjasek: after this you could add a step to open browser and navigate to `localhost:9000`

README.md
```diff
@@ -19,11 +19,42 @@ different technologies.
...
+* ElasticSearch
+* Redis
+
+#### MacOS
+
+For the sake of this walkthrough, it is considered that your python binary is called `python3`. Please amend as needed for your own system.
+
+1. Install dependencies using `brew` (make sure you do a `brew update` first): `python3`, `mongodb`, `elasticsearch`, `redis`.
+2. Create a virtual envirnoment in a folder of your choice, for example `mkdir ~/pyvenv && pyvenv-3.5 ~/pyvenv`.
+3. Activate the environment with by running its activation script: `. ~/pyvenv/bin/activate`. All of the following steps need to run under the virtual environment.
```
petrjasek: after this you will also need to install dependencies, `pip install -r server/requirements.txt`

actionless:
yawghen [1:20 PM] 
it will be better to split both mac and normal install to two separate docs since it could mislead

so please reference to linux install guide without docker after this line, https://github.com/superdesk/superdesk/pull/1989/files#diff-04c6e90faac2675aa89e2176d2eec7d8R81


> For manual installation just follow the steps described in the [Dockerfile](./Dockerfile).
> Or check this [guide](./INSTALL_UBUNTU1404.md) for more details.
(edited)

yawghen [1:25 PM] 
it's better to hide that guide as far from users as possible because they always getting manual install guides wrong and asking lots of useless questions

for example the list of system-wide deps for the backend should be wider for non-docker installation
and if a user will skip some warning about lib not found superdesk won't be working and so on

that's one of the reasons why most of the usesr should be ​*forced*​ to use docker since it's making it much harder to do something wrong

new messages
gbbr [1:33 PM] 
True

[1:34] 
Would you mind commenting on the PR so others can give their input?

gbbr:
> it will be better to split both mac and normal install to two separate docs since it could mislead

Can you please explain why you would find that to be misleading? Most repositories I've seen online use this form, where they have a list of OS's and describe their installation in the same file. The only time I've seen a use of different files is when the steps for a certain OS were way too long to fit into the original.

> that's one of the reasons why most of the usesr should be ​forced​ to use docker since it's making it much harder to do something wrong

Why should they be forced to use Docker? Personally I try to not use Docker as much as possible as it has caused me many issues in the past. It's generally ok to give them the option to chose and take full control of their installation. I think it's safe to assume that people installing the server are mostly developers so I don't see any risk. Also, on MacOS docker is tricky to work with.

I'm happy to make the changes you are suggesting but I'm not 100% convinced they are the best choice :)

gbbr:
@petrjasek true, I've missed those steps. All addressed and commit was amended.

README.md
gbbr: Done. See below.

README.md
gbbr: Added

actionless:
you are underestimating how aggressively stupid people can be.
your assumption on target audience is not matching with the support forums contents.

when using docker you're just starting already pre-configured image so it's reducing the probability of human factor drastically

gbbr:
How about I just make a separate file called `INSTALL.md` as you've suggested and place all of that there? We could link to it from the main file? Would that be a fair compromise?

actionless:
not really, docker way should be prioritized among users because of the reasons i given above

gbbr:
Yes, that's what I'm saying. Keep the readme as is, and add my changes to
another file to which we can link from the main one.

actionless:
ah, that's fine by me, i've actually proposed the same in a comment before it, but it was added as an edit so you could not notice it

petrjasek:
I don't think we should hide install guide, or even force anyone to use docker. We can move things around to start with docker, but there should be still an install guide even for docker, not only how to run demo with docker, which is not the same thing..

actionless:
@petrjasek by actual installation you mean production deployment, right?

i think for that case we could even make a special document, because some stuff for production deployment is not the same as for demoing/development not depends if docker is used or not

petrjasek:
I mean deployment without demo data. Production deployment is yet another missing piece.

actionless:
i think demo data thing can be resolved by separating ./scripts/docker-local-create-user.sh to actually creating admin user and prepopulating

gbbr:
Ok, so another option would be to re-arrange (as @petrjasek suggested) and recommend users to use Docker for ease of installation. And if someone feels hacky they can do the manual steps. I also feel a bit that adding an extra file under the presumption that "_people are clueless_" is not necessarily a successful approach and might make it harder for some to find the steps.

Which solution should I go with?

@actionless would you be ok with that?

actionless:
> and might make it harder for some to find the steps.

that's the idea:

if a person have no problems with reading that link will be found and manual install guide will successfully located

if not — it will be better to make it harder the situation when somebody will just paste to the console every command from the readme in order of appearance for all the platforms (disregard for explanation given before) and expecting everything to work

actionless:
> And if someone feels hacky they can do the manual steps.

that good point to state in the text which will be referencing to the manual install guide

gbbr:
@actionless I think we can merge this after [discussing it](https://dev.sourcefabric.org/browse/SD-4309?focusedCommentId=82294&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-82294) in today's retrospective.

PR #1994, opened 2016-04-28
fix(content types): Fix the backend validation for publish

ioanpocol:


petrjasek:
maybe we can also set by default `type=text` before doing the validation?
:+1:

ioanpocol:
OK, the change will be on superdesk-core

Aca-jov:
:+1:

naomiaro:
Is this something that will need to be updated on our pa-poc branch to continue working @ioanpocol ?

ioanpocol:
@naomiaro If the validation works well on PA no changes are required.

PR #2020, opened 2016-05-09
set abstract not to use italics

petrjasek:
after https://github.com/superdesk/superdesk-client-core/pull/434 it would switch to italics


akintolga:
:+1:

PR #2023, opened 2016-05-10
[SDNTB-15] [SDNTB-159] Create Scanpix search provider

jerome-poisson:
Implement Scanpix Search provider with features such as black&white
 or clearEdge pictures search, and search by one or severals IDs

SDNTB-15

### DIFF
--- a/server/ntb/scanpix/scanpix_datalayer.py
+++ b/server/ntb/scanpix/scanpix_datalayer.py
@@ -1,0 +1,367 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014, 2015, 2016 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+import arrow
+import logging
+import re
+import requests
+import json
+from datetime import datetime
+
+from io import BytesIO
+from eve.io.base import DataLayer
+from eve_elastic.elastic import ElasticCursor
+from superdesk.upload import url_for_media
+
+from superdesk.errors import SuperdeskApiError, ProviderError
+from superdesk.media.media_operations import process_file_from_stream, decode_metadata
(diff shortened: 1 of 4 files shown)


server/ntb/scanpix/scanpix_datalayer.py
```diff
@@ -0,0 +1,365 @@
...
+
+        It adds predefined headers and auth token if available.
+
+        :param url
+        :param data
+        """
+        r = requests.post(url, data=json.dumps(data), headers=self._headers, auth=(self._user, self._password))
+
+        if r.status_code < 200 or r.status_code >= 300:
+            logger.error('error fetching url=%s status=%s content=%s' % (url, r.status_code, r.content or ''))
```
petrjasek: maybe it should rather raise an error, because it continues to parse results even if there is an error

server/ntb/scanpix/scanpix_datalayer.py
```diff
@@ -0,0 +1,365 @@
...
+            # set the version created to now to bring it to the top of the desk, images can be quite old
+            doc['versioncreated'] = utcnow()
+            inserted = [file_id]
+            file_type = content_type.split('/')[0]
+            rendition_spec = get_renditions_spec()
+            renditions = generate_renditions(out, file_id, inserted, file_type,
+                                             content_type, rendition_spec,
+                                             url_for_media, insert_metadata=False)
+            doc['renditions'] = renditions
+        except Exception as io:
```
petrjasek: would be good if there is some more specific error..

petrjasek:
:+1: looks good, just added some minor comments

server/ntb/scanpix/scanpix_datalayer.py
jerome-poisson: OK I'll do that (this line come from pa-img, so there is the same issue there)

ioanpocol:
:+1:

jerome-poisson:
I've raised an error for first Petr comment. For the second one the error is raised by storage backend so we can't really predict what kind of error can be thrown, so I have changed it to have a generic Exception on the media.put call only, and more specific exception catch on the second part.

If it's OK and travis agree, this can be merged

petrjasek:
:+1:

PR #2026, opened 2016-05-12
update cores

petrjasek:


mdhaman:
@petrjasek you would have to reindex the elastic for the keyword changes.

petrjasek:
:ok:

PR #2033, opened 2016-05-16
fixed description_text in Scanpix search result

jerome-poisson:
There was a bad check, I've replaced it with a try/except which avoid
this kind of mistake and which is more pythonic

SDNTB-15


petrjasek:
:+1:

ioanpocol:
:+1:

PR #2038, opened 2016-05-19
[SD-4491] Mark tabular content as format preserved and pre tag it

marwoodandrew:
Dependant on superdesk/superdesk-core#363


marwoodandrew:
Only required in the AAP branch

PR #2042, opened 2016-05-19
feat(monitoring): add single line config for monitoring lists

petrjasek:
SDNTB-189

this will make more sense once https://github.com/superdesk/superdesk-client-core/pull/458 gets merged, but it won't hurt to have it in place :wink:


ioanpocol:
:+1:

PR #2050, opened 2016-05-23
SDPA-456 Blocked and then repeat publishing caused by "Failed to move to archived." error

mugurrus:
depends on:
https://github.com/superdesk/superdesk-core/pull/366
https://github.com/superdesk/superdesk-client-core/pull/506

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -48,6 +48,7 @@ INSTALLED_APPS.extend([
     'superdesk.activity',
     'superdesk.vocabularies',
     'apps.comments',
+    'superdesk.profiling',
 
     'superdesk.io',
     'superdesk.io.feeding_services',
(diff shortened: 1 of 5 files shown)


server/settings.py
```diff
@@ -111,3 +112,5 @@
 
 DEFAULT_URGENCY_VALUE_FOR_MANUAL_ARTICLES = None
 DEFAULT_SOURCE_VALUE_FOR_MANUAL_ARTICLES = 'PA'
+
+ENABLE_PROFILING = True
```
actionless: mb better read it from env vars?

actionless: or even it remove it from here but read env var from superdesk-core default settings

vied12:
I push this branch on this repository to be able to resolve the conflict 
https://github.com/superdesk/superdesk/pull/2059

PR #2052, opened 2016-05-25
[SD-4585] Fix bad sources on import from text archive

marwoodandrew:


server/aap/import_text_archive/commands.py
```diff
@@ -198,7 +206,46 @@ def _process_bunch(self, x):
...
+                        item['body_html'] = '<p>' + story + '</p>'
+                    else:
+                        item['body_html'] = '<pre>' + story + '</pre>'
+                    try:
+                        item['word_count'] = get_text_word_count(item['body_html'])
+                    except:
+                        pass
+
                 item['source'] = self._get_head_value(doc, 'Agency')
+                if item['source'] is None:
```
akintolga: I think this should be the last line to set if nothing else matches

server/aap/import_text_archive/commands.py
```diff
@@ -198,7 +206,46 @@ def _process_bunch(self, x):
...
+                            sign_off = last_line.split(' ')
+                            if len(sign_off) > 0:
+                                item['source'] = sign_off[0].upper()
+
+                            else:
+                                item['source'] = sign_off.upper()
+                        else:
+                            if len(item['source']) == 25:
+                                item['source'] = 'AAP'
+                        if item['source'].startswith('AAP'):
```
akintolga: If we don't expect a fall-through then all needs to be `elif`s

server/aap/import_text_archive/commands.py
marwoodandrew: It is a different kind of wrong to the other cases, this is if the agency is missing all together rather than being there but wrong!

server/aap/import_text_archive/commands.py
marwoodandrew: Fall through is good! still needs to check what got parsed for validity

akintolga:
:+1:

mdhaman:
:+1:

petrjasek:
:+1:

PR #390, opened 2015-04-16
update(documentation): Updating documentation with nested dropdown menus

pavlovicnemanja92:
And creating additional structure for n+ nested menus.

PR #2062, opened 2016-05-29
[SD-4611] Move AAP specific marcos, formatters  to AAP branch.

mdhaman:
This core PR depends on [#390](https://github.com/superdesk/superdesk-core/pull/390)

Some of the AAP related parsers will be moved in a separate PR.


akintolga:
:+1:

PR #2073, opened 2016-06-01
fix(client): add missing module to index

petrjasek:


petrjasek:
@ioanpocol this will be an issue also on ntb, you can cherry pick

PR #2080, opened 2016-06-02
Adding pyodbc dependency.

mdhaman:


.travis.yml
```diff
@@ -31,6 +31,8 @@ before_install:
     - unzip chrome-linux.zip
     - export CHROME_BIN=`pwd`/chrome-linux/chrome
     - $CHROME_BIN --version
+    - sudo apt-get -qq update
```
petrjasek: if you want to use sudo on travis you have to set `sudo: true` at the beginning

PR #2088, opened 2016-06-07
Missing keyword notification on publish fixed.

eozan:
SDPA-414

### DIFF
--- a/client/app/translations.js
+++ b/client/app/translations.js
@@ -11,7 +11,8 @@ angular.module('gettext').run(['gettextCatalog', function (gettextCatalog) {
         "KEYWORDS": "TAGS",
         "Keywords": "Tags",
         "URGENCY": "RANKING",
-        "Urgency": "Ranking"
+        "Urgency": "Ranking",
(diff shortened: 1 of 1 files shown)


client/app/translations.js
```diff
@@ -11,7 +11,8 @@ angular.module('gettext').run(['gettextCatalog', function (gettextCatalog) {
         "KEYWORDS": "TAGS",
         "Keywords": "Tags",
         "URGENCY": "RANKING",
-        "Urgency": "Ranking"
+        "Urgency": "Ranking",
+        "'SLUGLINE is a required field'": "Keyword is a required field"
```
akintolga: Hi @eozan, is the inner single-quote needed because it is a multi-word key?

ioanpocol:
:+1:

client/app/translations.js
vied12: Yes it seems like. Or the maybe the key is actually `'SLUGLINE is a required field'`.
But it works fine, I merge it :)

PR #2091, opened 2016-06-07
Fixed metadata handling for NITF parser

jerome-poisson:
- category is now added to item['subject'] with a "category" scheme
- "genre_custom" scheme is now added to found subjects, which are also stored
  in item['subject']

there is some code here (mainly add_subjects_scheme) which is not ideal in
settings, this is temporary and this should be moved in a separate
module (with a NITFFeedParser subclass) once SD-4650 is fixed.

SDNTB-213
SDNTB-219

### DIFF
--- a/server/settings.py
+++ b/server/settings.py
@@ -185,13 +185,28 @@ SIGN_OFF_MAPPING = 'email'
 DEFAULT_CONTENT_TYPE = 'Standard'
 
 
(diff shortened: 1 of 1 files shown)


petrjasek:
:+1:

ioanpocol:
:+1:

mugurrus:
:+1:

PR #2100, opened 2016-06-09
[Scanpix] Simplified rendition images selection

jerome-poisson:
the former way of choosing rendition images was to do a best guest
based on actual rendition sizes in settings, and it was not settings
correclty all renditions keys.

As renditions sizes have little chance to change, the code was
simplified and Superdesk renditions to Scanpix preview mapping is now
done with a simple dict containing a list of preview to get in order of
preference.

SDNTB-210


ioanpocol:
:+1:

PR #2104, opened 2016-06-10
[SDNTB-215] Change "subscriptions" in Scanpix search

jerome-poisson:
'all' subscription value is handled by removing the subscription filter.


petrjasek:
:+1:

PR #2114, opened 2016-06-15
fix(server): avoid errors when new relic license is not set

petrjasek:


naomiaro:
👍  cool, this helps then what Jerome and I probably had problems with?

akintolga:
:+1:

PR #2117, opened 2016-06-16
Handles Vidible embed codes SDPA-454

vied12:
- add in backend an endpoint to fetch Vidible video metadata
- define in frontend a special handler for Vidible embed code

requires https://github.com/superdesk/superdesk-client-core/pull/610

and https://github.com/superdesk/superdesk-core/pull/432 to be fully complete


petrjasek:
:+1:

PR #2135, opened 2016-06-22
[NITF] fixed parsing of services (anpa_category)

jerome-poisson:
This commit fix the handling of services for NTB, qcode and language
found in vocabularies are used (or name is used as qcode if not found).

fix SDNTB-233


petrjasek:
:+1:

PR #2143, opened 2016-06-27
[scanpix] create 4 instances of Scanpix search provider

jerome-poisson:
The 4 instances are hardcoded as it is currently not possible to set
dynamically search providers in frontend.

fix SDNTB-217


petrjasek:
:+1:

PR #2149, opened 2016-06-29
SDPA-436 SDPA-437 SDPA-438 SDPA-439 SDPA-440 - content types

mugurrus:


petrjasek:
:+1:

server/data/content_types.json
```diff
@@ -204,5 +204,171 @@
...
+            "abstract": {"required": true},
+            "byline": {},
+            "feature_image": {"type": "picture", "required": true},
+
+            "topic_news": {"required": true},
+            "age_range": {},
+            "territory": {"required": true},
+            "reader_type": {},
+            "assets": {},
+            "relatedItems": {}
```
gbbr: Some of these items in the schema need to have translations on the client ([see here](https://github.com/superdesk/superdesk-client-core/blob/master/scripts/superdesk-workspace/content/content.js#L533-L555)). For example, `relatedItems` and `assets` isn't there at the moment. 

Unfortunately, this is not a very maintainable solution and a better one would be required. I would suggest adding another field in the backend, named `label`, which would describe what the editor should show. Any other suggestions or solutions are welcome.

gbbr:
Besides the missing translations, this LGTM 👍

server/data/content_types.json
mugurrus: Thanks, will do

PR #2156, opened 2016-07-05
angular-history.js added to index.html

vied12:
Depends on https://github.com/superdesk/superdesk-client-core/pull/671


vied12:
It is not needed anymore now we have webpack working with SD

PR #2165, opened 2016-07-11
SD-4886 Metadata validation on content profiles does not work correctly

mugurrus:
metadata validation and upgrade for PA


vied12:
:+1: 

(I wonder how much we could factorize in server/data/content_types.json )

ioanpocol:
:+1:

PR #2168, opened 2016-07-13
[NITF] NITF Publisher with customizations for NTB

jerome-poisson:
This commit implement a new NTBNITF subscribe type and publish items
in NITF format with customizations for NTB.

SDNTB-229

### DIFF
--- a/server/ntb/publish/ntb_nitf.py
+++ b/server/ntb/publish/ntb_nitf.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8; -*-
+#
+# This file is part of Superdesk.
+#
+# Copyright 2013, 2014 Sourcefabric z.u. and contributors.
+#
+# For the full copyright and license information, please see the
+# AUTHORS and LICENSE files distributed with this source code, or
+# at https://www.sourcefabric.org/superdesk/license
+
+from superdesk.metadata.item import ITEM_TYPE, CONTENT_TYPE
+from superdesk.publish.formatters.nitf_formatter import NITFFormatter
+from xml.etree.ElementTree import SubElement
+import logging
+import re
+logger = logging.getLogger(__name__)
+
+EMBED_RE = re.compile(r"<!-- EMBED START ([a-zA-Z]+ {id: \"(?P<id>.+?)\"}) -->.*"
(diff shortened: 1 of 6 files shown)


jerome-poisson:
https://github.com/superdesk/superdesk-core/pull/462 need to be merged before this one can be merged

mugurrus:
:+1:

server/ntb/publish/ntb_nitf.py
```diff
@@ -0,0 +1,129 @@
...
+        :param str format_type:
+        :param dict article:
+        :return: True if article can formatted else False
+        """
+        return format_type == 'ntbnitf' and article[ITEM_TYPE] == CONTENT_TYPE.TEXT
+
+    def _format_tobject(self, article, head):
+        category = ''
+        for s in article.get('subject', []):
+            if s.get('scheme') == 'category':
```
ioanpocol: Category is a custom field added for NTB. Can we avoid this check and make it more generic?
For example we can test is scheme is not subject to get qcode.

server/ntb/publish/ntb_nitf.py
```diff
@@ -0,0 +1,129 @@
...
+                category = s['qcode']
+        return SubElement(head, 'tobject', {'tobject.type': category})
+
+    def _format_docdata(self, article, docdata):
+        super()._format_docdata(article, docdata)
+        if 'slugline' in article:
+            SubElement(docdata, 'du-key', attrib={'version': '1', 'key': article['slugline']})
+
+    def _format_subjects(self, article, tobject):
+        subjects = [s for s in article.get('subject', []) if s.get("scheme") == "subject_custom"]
```
ioanpocol: Subject_custom is a scheme added for NTB only. We have to avoid to use custom fields in code.

ioanpocol:
:+1: 

If the references to custom fields are avoided, this class can be added to core.

jerome-poisson:
@mugurrus indeed, I'll add some before merging
@ioanpocol  normally this class is only handling NTB specific stuff, the generic one are in the core NITFFormatter class. But that true that the regex and method in _format_body_content could be useful for future subclasses.

PR #2170, opened 2016-07-14
[scanpix] change subscription behaviour for ntbtema

jerome-poisson:
as requested by NTB (see linked ticket), "subscription" filter is
replaced by "punchcard" if set to "subscription" ("inside subscription"
in the frontend) only for ntbtema resource.

fix SDNTB-250


mugurrus:
:+1:

PR #2171, opened 2016-07-14
[scanpix] allow "*" in text search

jerome-poisson:
regex parsing queries was ignoring "*", but it is needed to handle
wildcards.

Also fixed a small docstring issue

fix SDNTB-240


mugurrus:
:+1:

PR #2181, opened 2016-07-21
[NITF Formatter] fixed featuremedia and body_html handling

jerome-poisson:
only "featureimage" was used, now "featuremedia" is checked too if
"featureimage" is not found

body_html was added to NITF's <body.content/> using the inherited
method from core's NITF Formatter, but this was not getting all the
elements.
To work around this, the inherited method is not used anymore, and
BeautifulSoup is used to transform the HTML to XML, which is then added
to body.content

This commit fixes the issues related in feedback of SDNTB-229 (in the
comments)


petrjasek:
:+1: but some unittest would be great

jerome-poisson:
@petrjasek I have made a test for the formatter, I'll update it.

PR #2182, opened 2016-07-22
chore(cores): Update supedesk cores for NTB

ioanpocol:


petrjasek:
I guess this will need an update too @gbbr

gbbr:
Yes, I will take care of updating client core today on both NTB and PA. If this is urgent, maybe remove the core update and merge the rest...

PR #2202, opened 2016-07-28
[NITF] added evloc element in NTB NITF

jerome-poisson:
<evloc/> element was missing, this patch add it

SDNTB-229


petrjasek:
:+1:

PR #2208, opened 2016-08-03
Update core (master)

gbbr:
Reverts https://github.com/superdesk/superdesk/pull/2194 and depends on https://github.com/superdesk/superdesk-client-core/pull/788


gbbr:
I am merging this as it has already previously been reviewed and thoroughly tested.

PR #2210, opened 2016-08-04
Update core (NTB)

gbbr:
Reverts https://github.com/superdesk/superdesk/pull/2199 and depends on  superdesk/superdesk-client-core#788


Aca-jov:
:+1:

jerome-poisson:
Hi @gbbr can you explain the revert? And actually what you are doing globally? I'm reviewing the code but I don't get why so many files are removed, specially things like LICENSE.txt or .dockerignore.

Thanks :)

gbbr:
Hey @jerome-poisson - good question!

This is not actually a revert in the sense that you might think. The original PR which was equal to this one was breaking deployment with Bamboo so I had to revert it. This PR reverts that revert (inception).

As for what it actually does. The main idea is that this repository should only contain a configuration file that allows to define how the client-side app builds based on customer needs, using the `superdesk.config.js` file. Now, when you run `grunt build` (or `server`), the build that is generated is based on the configuration from the file. On some builds you might get some features, on others you won't.

As for the files that were removed, those were moved into the client repository where they belong. A lot of them were not even used, and most of the testing files should be with the client core, where they belong.

I hope this clears your questions a bit, if not, I am happy to further explain.

gbbr:
P.S. Basically the point of this work was to remove code duplication. We had `index.html`, `main.js` duplicated both in the client core and this repository, making it hard to consolidate changes many times, and have stuff breaking because of small inconsistencies people involved don't normally consider. I strongly feel that client app code should live in client app repository and not in two places because that is not a good direction to go towards.

jerome-poisson:
@gbbr OK got it, this seems sane, thanks for the explanations.

:+1:

ioanpocol:
:+1:

gbbr:
Thanks for reviewing, whenever you want to update client-core next time on NTB, please merge this PR first.

PR #2225, opened 2016-08-10
Update client core (PA-POC)

gbbr:
Bower has been removed from the codebase, making these files no longer needed. See superdesk/superdesk-client-core#809 for more detailed updates.


pavlovicnemanja:
:+1:

PR #2237, opened 2016-08-16
Clean-up unused files

gbbr:
- Updates client core
- Removes left-over bower files
- Removes extra `.gitignore` (only one in the root folder is enough)


thnkloud9:
:thumbsup:

petrjasek:
:+1:

PR #2252, opened 2016-08-29
refact(search) - Duplicate search parameters that exist in filters removed

akintolga:
- anpa_category, genre, priority and urgency exist in aggregations and they can be used from "Filters"

### DIFF
--- a/client/superdesk.config.js
+++ b/client/superdesk.config.js
@@ -55,12 +55,8 @@ module.exports = function(grunt) {
         },
 
         search_cvs: [
-            {id: 'anpa_category', name:'Service', field: 'anpa_category', list: 'categories'},
             {id: 'category', name:'Category', field: 'subject', list: 'category'},
-            {id: 'genre', name:'Genre', field: 'genre', list: 'genre_custom'},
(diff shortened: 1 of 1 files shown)


akintolga:
pinging @ioanpocol

akintolga:
depends on https://github.com/superdesk/superdesk-client-core/pull/903

ioanpocol:
Even if those items appears on facets, NTB requested to have them on filters too. Probably Karel can show again the new refactored search to NTB and probably they will change their mind.

akintolga:
@ioanpocol thanks for the response. 

When I was testing in https://ntb.dev.superdesk.org/#/ I noticed that the way those parameters work also different than the way that they work in filters. This affects the way the query and tags. 

I agree with you that they will probably change their mind, even if not this area needs to be revisited to make it similar to the filters.

ioanpocol:
@akintolga At the moment we implemented it worked well, but I will check it again.

akintolga:
Karel has approved these changes for NTB today.

ioanpocol:
:+1:

mugurrus:
:+1:

PR #2256, opened 2016-08-30
[NITF] [formatter] added new elements

jerome-poisson:
the following elements have been added:
	- meta timestamp, ntb-dato, NTBUtDato, NTBStikkord, NTBBilderAntall, NTBDistribusjonsKode, NTBKanal, NTBID, NTBIPTCSequence, NTBEditor
	- date.issue now follows NTB format
	- added key-list/keyword in docdata
	- added tagline/a in body.end
	- added distributor/org in body.head
	- added baselang in <nitf> root element
	- doc-id follow new NTB specifications

following elements have been removed:
	- NTBPrioritet
	- date.expire

abstract in <p class="lead" lede="true"> element now use text with HTML
elements stripped.

if there is a paragraph in body_html using class="lead", it's changed
    for "txt", and lede attribute is removed if it exists.

mime_type had default values of "image/jpeg" for images and
"video/mpeg" for videos.

date has been removed from body.head/dateline

mimetype guessing in Scanpix search provider has been fixed and is now based on originalFileName
then fileFormat

requested in SDNTB-254

### DIFF
--- a/server/ntb/publish/ntb_nitf.py
+++ b/server/ntb/publish/ntb_nitf.py
@@ -10,13 +10,13 @@
 
 from superdesk.metadata.item import ITEM_TYPE, CONTENT_TYPE
 from superdesk.publish.formatters.nitf_formatter import NITFFormatter
-from xml.etree.ElementTree import SubElement
 import re
 from bs4 import BeautifulSoup
 from xml.etree import ElementTree as ET
 from superdesk.publish.publish_service import register_file_extension
 from superdesk.errors import FormatterError
 import superdesk
+from datetime import datetime
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


server/ntb/publish/ntb_nitf.py
```diff
@@ -148,6 +225,16 @@ def repl_embedded(match):
         # and etree from stdlib doesn't have a proper HTML parser
         soup = BeautifulSoup(html, 'html.parser')
         html_elts = ET.fromstring('<div>{}</div>'.format(soup.decode(formatter='xml')))
+        # <p class="lead" lede="true"> is used by NTB for abstract
```
petrjasek: afaik it's part of nitf standard, maybe they want to have it there. (not the class, but the `lede=true`)

jerome-poisson: the lede="true" is used on the first <p> and they want the abstract there (with class="lead").

We can also get <p class="lead"> on the first paragraph from client (it may come from the ingest, not sure if it's the client or the ingest), so I need to remove to avoid 2 "lead". But there will still be a <p class="lead" lede="true"> as first paragraph, just it will come from the abstract and not the body_html.

petrjasek: I see

petrjasek:
:+1:

naspeh:
👍 

BTW: 

What about to use lxml?
1. for getting text
     - http://lxml.de/lxmlhtml.html#html-element-methods `.text_content()`
     - http://lxml.de/lxmlhtml.html#cleaning-up-html
2. for etree
     `from lxml import etree`
     http://lxml.de/tutorial.html
     seems it has the same iterface

or it is not worth dependency?

jerome-poisson:
@naspeh I'm also for using lxml because it's fast and efficient (and ElemenTree is based on it), but BeautifulSoup is already used here, and there is no sense to add a dependency. But I would be in favor of replacing BeautifulSoup by lxml if we use it eveywhere, as BS engine can be used in lxml anyway. Maybe we can discuss it at SourceCamp? @petrjasek what do you think?

PR #2264, opened 2016-09-01
[NITF] [FORMATTER] use new specification for subject

jerome-poisson:
"subject" meta element now uses the new specifications with a nyX-
prefix for article updates.

SDNTB-254


ioanpocol:
:+1:

PR #2266, opened 2016-09-01
[content types] added default value for body footer

jerome-poisson:
body footer has now a default value of "©NTB"

SDNTB-261


ioanpocol:
:+1:

PR #2267, opened 2016-09-01
[NITF] [FORMATTER] added body footer

jerome-poisson:
body footer is added a new paragraph at the end of the body_content

SDNTB-263

### DIFF
--- a/server/ntb/publish/ntb_nitf.py
+++ b/server/ntb/publish/ntb_nitf.py
@@ -263,6 +263,14 @@ class NTBNITFFormatter(NITFFormatter):
                 del p_lead.attrib['lede']
             except KeyError:
                 pass
+
+        try:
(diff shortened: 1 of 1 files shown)


server/ntb/publish/ntb_nitf.py
```diff
@@ -263,6 +263,14 @@ def repl_embedded(match):
                 del p_lead.attrib['lede']
             except KeyError:
                 pass
+
+        try:
```
ioanpocol: why not a shorter version:
if 'body_footer' in article: ...

jerome-poisson: it's the ask forgiveness not permission pattern. it's slightly better is the key is supposed to be available.
I used it often by habit, there is not much difference with the other way on such a small thing, but there is nothing wrong by using it.

PR #2269, opened 2016-09-02
[NITF] [FORMATTER] use local time for filename generation

jerome-poisson:
improvment on SDNTB-264

### DIFF
--- a/server/ntb/publish/ntb_nitf.py
+++ b/server/ntb/publish/ntb_nitf.py
@@ -117,7 +117,7 @@ class NTBNITFFormatter(NITFFormatter):
         example: 2016-08-16_11-07-46_Nyhetstjenesten_Innenriks_ny1-rygge-nedgang.xml
         """
         metadata = {}
-        metadata['date'] = article['versioncreated'].strftime("%Y-%m-%d_%H-%M-%S")
+        metadata['date'] = article['versioncreated'].astimezone().strftime("%Y-%m-%d_%H-%M-%S")
         try:
             metadata['service'] = article['anpa_category'][0]['name']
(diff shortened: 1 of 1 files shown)


server/ntb/publish/ntb_nitf.py
```diff
@@ -117,7 +117,7 @@ def _get_filename(self, article):
         example: 2016-08-16_11-07-46_Nyhetstjenesten_Innenriks_ny1-rygge-nedgang.xml
         """
         metadata = {}
-        metadata['date'] = article['versioncreated'].strftime("%Y-%m-%d_%H-%M-%S")
+        metadata['date'] = article['versioncreated'].astimezone().strftime("%Y-%m-%d_%H-%M-%S")
```
petrjasek: is it using server timezone?

jerome-poisson: It's using server's local time zone yes. Actually it was not specified and UTC was used here, but as local timezone is used in other NTB metadata, I've changed it too for consistency.

ioanpocol:
:+1:

server/ntb/publish/ntb_nitf.py
petrjasek: imo it would be better then to use zone from config - there is `DEFAULT_TIMEZONE` setting

jerome-poisson: oh I didn't knew there was this in settings, yes it makes definitely more sense then, I'll update the P.R.

jerome-poisson:
following P.R. will replace this one: https://github.com/superdesk/superdesk/pull/2282, so I'm closing it.

PR #2271, opened 2016-09-05
[NITF] [FORMATTER] moved <distributor/> element creation in new method

jerome-poisson:
<distributor/> element is now added in its own method so it is added in
the right location.

SDNTB-280


ioanpocol:
:+1:

petrjasek:
:+1:

PR #2273, opened 2016-09-05
feat(macro): Extract Html Macro

ioanpocol:
Delete from body_html all html tags except links


jerome-poisson:
it works but I wonder if it would not be better to go into the tree and remove non link elements manually instead of using a replace with special string.

anyway I'm :+1:

PR #2275, opened 2016-09-06
Fix WebSocket url for dev.superdesk.org instances

naspeh:
Seems something changes and instead of `localhost` in app.bundle.js we have `0.0.0.0`.


petrjasek:
:+1:

PR #2282, opened 2016-09-07
[NITF] [FORMATTER] use DEFAULT_TIMEZONE for localized date elements

jerome-poisson:
DEFAULT_TIMEZONE from settings is now used, so admin can choose if she
wants to use local timezone (if not set) or any other one.

SDNTB-285


petrjasek:
:+1:

PR #2286, opened 2016-09-07
[NITF] [FORMATTER] added pubdata element

jerome-poisson:
this commit add pubdata with date.publication attribute.

SDNTB-286


ioanpocol:
:+1:

petrjasek:
:+1: but would be good to have a test

PR #2296, opened 2016-09-12
[NITF] [FORMATTER] fixed dateline element

jerome-poisson:
SDNTB-280


petrjasek:
:+1: test would be great

jerome-poisson:
@petr yes I was actually writting some

jerome-poisson:
@petrjasek I've changed test to avoid formatting for each test, so if you can give a second look it would be great, thanks :)

petrjasek:
:+1:

PR #2298, opened 2016-09-13
[NITF] [FORMATTER] escape non unicode characters

jerome-poisson:
previously NTB NITF formatter was encoding unicode XML to requested
iso-8859-1 format, replacing invalid characters in this format by "?",
because XML was supposedly not wanted.

It seems that XML encoding is accepted at the end, so this commit
change this behaviour to encode directly in iso-8859-1, and XML escape invalid chars.

A unicode version is kept in formatted_item, as it may be useful for
previewing.

SDNTB-280


naspeh:
👍 looks good

PR #2304, opened 2016-09-16
Added e2e docker-compose specific configuration

lnogues:
This file is adding a docker specific configuration for e2e tests.

I've update my gist accordingly: https://gist.github.com/lnogues/09e6f1161d853b3523286c8774d42ffc

### DIFF
--- a/docker/start-e2e.sh
+++ b/docker/start-e2e.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+#cd /opt/superdesk &&
+#bash ./scripts/fig_wrapper.sh honcho start
+
+cat >/tmp/testenv <<EOF
+SUPERDESK_TESTING=true
+
+EOF
+
+cd /opt/superdesk/client-core &&
+npm link &&
+cd /opt/superdesk/client &&
+npm link superdesk-core &&
+grunt --force server --server='http://localhost:5000/api' --ws='ws://localhost:5100' &
+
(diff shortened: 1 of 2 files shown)


docker/start-e2e.sh
```diff
@@ -0,0 +1,18 @@
...
+EOF
+
+cd /opt/superdesk/client-core &&
+npm link &&
+cd /opt/superdesk/client &&
+npm link superdesk-core &&
+grunt --force server --server='http://localhost:5000/api' --ws='ws://localhost:5100' &
+
+cd /opt/superdesk &&
+bash ./scripts/fig_wrapper.sh honcho -e /tmp/testenv start &&
```
naspeh: You could probably do:
```
bash SUPERDESK_TESTING=true ./scripts/fig_wrapper.sh honcho -e /tmp/testenv start
```
insead of this `/tmp/testenv`

docker/start-e2e.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#cd /opt/superdesk &&
+#bash ./scripts/fig_wrapper.sh honcho start
```
naspeh: Please remove comments here.

docker/start-e2e.sh
lnogues: Although I agree with you, it doesn't work when written that way

PR #2321, opened 2016-10-04
[NITF] [FORMATTER] filter <p> elements to set NTB requested classes

jerome-poisson:
<p> element class is now set like this:
- "lead" is used for abstract
- "txt" for <p> immediately following <hl2>
- "txt-ind" in all other cases

SDNTB-304

### DIFF
--- a/server/ntb/publish/ntb_nitf.py
+++ b/server/ntb/publish/ntb_nitf.py
@@ -18,6 +18,7 @@ from superdesk.publish.publish_service import PublishService
 from superdesk.errors import FormatterError
 import superdesk
 from datetime import datetime
+import copy
 import pytz
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


jerome-poisson:
this PR can pass travis until https://github.com/superdesk/superdesk-core/pull/599 is merged

server/ntb/publish/ntb_nitf.py
```diff
@@ -66,6 +72,20 @@ def format(self, article, subscriber, codes=None, encoding="us-ascii"):
...
             raise FormatterError.nitfFormatterError(ex, subscriber)
 
+    @staticmethod
+    def p_filter(root_elem, p_elem):
+        """modify p element to have 'txt' or 'txt-ind' attribute
+
+        'txt' is only used immediatly after "hl2" elem, txt-ind in all other cases
+        """
+        parent = next((p for p in root_elem.iter() if p_elem in p))
+        children = list(parent)
```
naspeh: seems it is better to rewrite this like:
```
+        children = [p for p in root_elem.iter() if p_elem in p]
```

naspeh: sorry, I just confused with names `parent`, then `children = list(parent)` :)

PR #2322, opened 2016-10-04
Update development setup guide

djKooks:
It seems there are some link name error on guide. Could you review this change?


petrjasek:
thx for the fix!

PR #2341, opened 2016-10-12
[NITF] [FORMATTER] remove styles

jerome-poisson:
use the new attr_reset from html2nitf to remove "style" attributes

SDNTB-311


jerome-poisson:
this P.R. can't pass travis before https://github.com/superdesk/superdesk-core/pull/615 is merged

PR #2353, opened 2016-10-17
SDESK-223: Scanpix search: Display archived-date in search results

thnkloud9:
update scanpix datalayer to use archivedTime as versioncreated

### DIFF
--- a/server/ntb/scanpix/scanpix_datalayer.py
+++ b/server/ntb/scanpix/scanpix_datalayer.py
@@ -213,7 +213,7 @@ class ScanpixDatalayer(DataLayer):
             new_doc['original_source'] = new_doc['source'] = doc['credit']
         except KeyError:
             pass
-        new_doc['versioncreated'] = utcnow()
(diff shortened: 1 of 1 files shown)


server/ntb/scanpix/scanpix_datalayer.py
```diff
@@ -213,7 +213,7 @@ def _parse_doc(self, doc):
             new_doc['original_source'] = new_doc['source'] = doc['credit']
         except KeyError:
             pass
-        new_doc['versioncreated'] = utcnow()
+        new_doc['versioncreated'] = self._datetime(doc['archivedTime'])
         new_doc['firstcreated'] = self._datetime(doc['archivedTime'])
```
jerome-poisson: these 2 lines could be merged to avoid calling 2 times the same method: 
  new_doc['versioncreated'] = new_doc['firstcreated'] = self._datetime(doc['archivedTime'])

thnkloud9: ah, right, thanks!

PR #2363, opened 2016-10-19
Remove "server/data" directory [SDESK-199]

naspeh:
- check if data directory exists before setting `INIT_DATA_PATH`
- update `logging_config.yml` (is somebody using logstash now?)

Related to https://github.com/superdesk/superdesk-core/pull/631


petrjasek:
just wondering - vocabularies and validators will be in core?

naspeh:
yes, it uses data from core by default, so there are in core, but for example, if you create `data` directory for NTB branch, you should put them there for sure

PR #2367, opened 2016-10-20
SDNTB-205 Added config for angular locale version

vladnicoara:


naspeh:
Pretty simple. @gbbr is it ok?

gbbr:
Yes, this should cut it.

PR #2378, opened 2016-10-25
(swimlane) - swimlane view config

sjunaid:
[SDESK-288]

- This change will enable swimlane view in master instance with four columns


sjunaid:
This depends on https://github.com/superdesk/superdesk-client-core/pull/1028

PR #2384, opened 2016-10-27
feat(publish): For tree type vocabularies add the parent if a child i…

ioanpocol:
…s present

### DIFF
--- a/server/ntb/publish/ntb_nitf.py
+++ b/server/ntb/publish/ntb_nitf.py
@@ -21,6 +21,7 @@ from datetime import datetime
 import copy
 import pytz
 import logging
+from copy import deepcopy
 logger = logging.getLogger(__name__)
 tz = None
 
@@ -46,7 +47,9 @@ class NTBNITFFormatter(NITFFormatter):
         """
         return format_type == 'ntbnitf' and article[ITEM_TYPE] == CONTENT_TYPE.TEXT
 
-    def format(self, article, subscriber, codes=None, encoding="us-ascii"):
+    def format(self, original_article, subscriber, codes=None, encoding="us-ascii"):
+        article = deepcopy(original_article)
+        self._populate_metadata(article)
         global tz
         if tz is None:
             # first time this method is launched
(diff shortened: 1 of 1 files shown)


server/ntb/publish/ntb_nitf.py
```diff
@@ -73,6 +76,31 @@ def format(self, article, subscriber, codes=None, encoding="us-ascii"):
...
+        vocabularies = list(superdesk.get_resource_service('vocabularies').get(None, None))
+        fields = {'place': 'place_custom', 'subject': 'subject_custom'}
+        for field in fields:
+            vocabulary = self._get_list_element(vocabularies, '_id', fields[field])
+            if not vocabulary:
+                continue
+            vocabulary_items = vocabulary.get('items', [])
+            field_values = article.get(field, [])
+            for value in list(field_values):
+                if value.get('parent', None) and not self._get_list_element(field_values, 'qcode', value['parent']):
```
naspeh: Will be better to use one more variable for: `self._get_list_element(field_values, 'qcode', value['parent'])`

ioanpocol: done

server/ntb/publish/ntb_nitf.py
```diff
@@ -73,6 +76,34 @@ def format(self, article, subscriber, codes=None, encoding="us-ascii"):
...
+            if not vocabulary:
+                continue
+            vocabulary_items = vocabulary.get('items', [])
+            field_values = article.get(field, [])
+            for value in list(field_values):
+                if not value.get('parent', None):
+                    continue
+                parent = self._get_list_element(field_values, 'qcode', value['parent'])
+                if not parent:
+                    field_values.append(self._get_list_element(vocabulary_items, 'qcode', value['parent']))
```
naspeh: sorry, seems I missed that it different calls:
```
self._get_list_element(field_values, 'qcode', value['parent'])
self._get_list_element(vocabulary_items, 'qcode', value['parent'])
```
if they'll be the same, it'll be more sence to add externall variable.

So please revert the last change and just merge PR.

Sorry and thanks.

ioanpocol: It's ok, usually is not a good practice to put call functions with many parameters on if condition.

PR #2402, opened 2016-11-08
Submodules

naspeh:
Let's use git submodules for cores, benefits:
- github shows them (in this PR or inside code tree https://github.com/naspeh-sf/superdesk/tree/submodules)
- updating cores: go to `client-core` or `server-core` and `git pull` then commit new version of submodules in the main repo
- all sourcefabric code in specific place (not in `node_modules` or inside `virtualenv`)

**As developer:**
I can just clone core repos to these directories and develop inside without using `git submodule update --init`. Anyway, I still need to use `npm link` for `client-core`.

**As devops:**
I can use the same tree to test cores, just clone the main repo and then fetch specific commit for cores into these directories.

P.S. [Git Submodules basic explanation](https://gist.github.com/gitaarik/8735255)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "client-core"]
+	path = client-core
+	url = https://github.com/superdesk/superdesk-client-core.git
+[submodule "server-core"]
+	path = server-core
+	url = https://github.com/superdesk/superdesk-core.git
--- a/README.md
+++ b/README.md
@@ -21,6 +21,38 @@ Find more information about the client configuration in the README file of the r
 
 ## Installation
 
+### Cloning the repo
+
+#### If you want just install it
+```bash
+git clone https://github.com/superdesk/superdesk.git
+cd superdesk
+git submodule update --init
+```
+
+#### If you going to develop cores
+Please don't run `git submodule update --init`, please do
+```bash
+git clone git@github.com:superdesk/superdesk.git
+cd superdesk
+git clone git@github.com:superdesk/superdesk-core.git server-core
+git clone git@github.com:superdesk/superdesk-client-core.git client-core
+```
+Than you can update cores and commit and still can update the main repo without dealing with submodules.
+
+#### If you want to change core references
+Please use a separate local repo for that.
+```bash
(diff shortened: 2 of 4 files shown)


hlmnrmr:
Not sure about this. I understand the idea behind, but with submodules you really need to know what you are doing, otherwise you may get into trouble. Let's see what others say about this @superdesk/superdesk-core @superdesk/superdesk-editorial @superdesk/superdesk-data

naspeh:
@hlmnrmr please share the source of
> you really need to know what you are doing, otherwise you may get into trouble

We point exact commit in our references, you can do it with submodules - they are here for that...

petrjasek:
I don't see this very useful for my workflow, not sure about others (especially devops).
also not sure how I would profit from those listed benefits:

> updating cores: go to client-core or server-core and git pull then commit new version of submodules in the main repo

we have a script for that

> all sourcefabric code in specific place (not in node_modules or inside virtualenv)

can you explain why this is good?

having said that, I'm not against this. can we use tags for submodules, or just commit ids?

hlmnrmr:
@petrjasek afaik, you can also use tags

naspeh:
Updated [README a little](https://github.com/naspeh-sf/superdesk/blob/748f32ae276b15e6eb0bdafd1435c0de0b6b745a/README.md#cloning-the-repo):

@petrjasek: you can use it with branches too, for example 1.0 releases

```
cd client-core
git checkout 1.4
cd server-core
git checkout 1.4
# and then "git pull origin HEAD" for updates
```

> can you explain why this is good?

I'm going to update installation docs with "Unify deployment" stuff.

For example if I'm going to contribute to `superdesk-core` I'll still use this main repo with the same installation, then I  change the code in `server-core` directory and virtualenv will see the changes...

The same for `client-core` to use one installation instruction.

gbbr:
In my experience people generally avoid using git submodules. Their implementation is faulty and they can cause a lot of problems in your workflow. I don't have solid evidence to back this up, but I can tell you for sure that it is something generally avoided. Not much to add other than this.

naspeh:
Seems I need to find another solution without submodules. Thanks, guys. I'm closing this PR.

PR #2415, opened 2016-11-22
fix(docker): remove `SUPERDESK_TESTING` from env

petrjasek:


petrjasek:
@ioanpocol you might need those `docker/compose.yml` changes in ntb too

PR #2460, opened 2017-01-19
update core

jerome-poisson:
this commit update superdesk-core and Dockerfile because libxml and libxslt are now needed to build lxml

SDNTB-328

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ build-essential libffi-dev git \
 libtiff5-dev libjpeg8-dev zlib1g-dev \
 libfreetype6-dev liblcms2-dev libwebp-dev \
 curl libfontconfig nodejs npm nginx \
+libxml2-dev libxslt1-dev \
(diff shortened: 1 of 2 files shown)


petrjasek:
heads-up @mdhaman this might need some changes in your stack when you start using 1.6+
 also in deploy scripts @naspeh

jerome-poisson:
@petrjasek I've already merged it for `superdesk-ntb` because the built was broken so it was urgent (it was passing travis, so I realized afterwards). Cc @naspeh

petrjasek:
@jerome-poisson :ok: but afaik we don't use docker for deployments, so not sure if that helps

jerome-poisson:
@petrjasek the bamboo build was passing after that.

petrjasek:
yep we probably still use docker for superdesk-ntb, but not for master, and not for ntb.superdesk.pro
anyway @naspeh should be aware now :wink:

naspeh:
thanks, I saw this `lxml` change, `sd*.test` is working well

PR #2463, opened 2017-01-23
Update server core

gbbr:
Update server core to superdesk/superdesk-core@6ecd2b4f8f74b2161fa78c20579b595467ea5a9d and https://github.com/superdesk/superdesk-core/pull/749

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -2,4 +2,4 @@ gunicorn==19.6.0
 honcho==0.6.6
 newrelic>=2.66,<2.67
 
-git+git://github.com/superdesk/superdesk-core.git@108d3d1#egg=Superdesk-Core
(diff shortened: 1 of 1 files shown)


server/settings.py
```diff
@@ -46,6 +46,13 @@ def env(variable, fallback_value=None):
 WS_HOST = env('WSHOST', '0.0.0.0')
 WS_PORT = env('WSPORT', '5100')
 
+LANGUAGES = [
+	{'language': 'nb-NO', 'label': 'Bokmål', 'source': True, 'destination': False},
```
petrjasek: issue is with this indentation, it should be 4 spaces each

gbbr: These changes I've added accidentally 👎 
I'm removing them...

PR #2480, opened 2017-02-03
Update installation section in README

naspeh:
@hlmnrmr, @petrjasek what do you think about such update?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,89 +10,28 @@ distribution and publishing platform developed and maintained by Sourcefabric
 with the sole purpose of making the best possible software for journalism. It
 is scaleable to suit news organizations of any size. See the [Superdesk website] (http://www.superdesk.org) for more information.
 
-Looking to stay up to date on the latest news? [Subscribe] (http://eepurl.com/bClQlD) to our monthly newsletter. 
+Looking to stay up to date on the latest news? [Subscribe] (http://eepurl.com/bClQlD) to our monthly newsletter.
 
-The Superdesk server provides the API to process all client requests. The client 
-provides the user interface. Server and client are separate applications using 
+The Superdesk server provides the API to process all client requests. The client
+provides the user interface. Server and client are separate applications using
 different technologies.
 
 Find more information about the client configuration in the README file of the repo:
-[https://github.com/superdesk/superdesk-client-core](https://github.com/superdesk/superdesk-client-core "") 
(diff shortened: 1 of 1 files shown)


petrjasek:
imo we should still provide information how to install it without any tool, so users get an idea what are dependencies etc.

naspeh:
@petrjasek this why I give a link to [details](https://github.com/superdesk/fireq/blob/master/files/superdesk/install) where you can see all steps in straightforward bash. And more options later.

The reason to get rid this Docker mentions  and to start pointing to [superdesk/fireq](https://github.com/superdesk/fireq).

petrjasek:
it's cool to remove docker, but what you link there is not user friendly.

naspeh:
but you can copy-past any command from this. I'm going to update documentation and add more user-friendly description, but for now I think we should remove this docker stuff from README, it's not working now

petrjasek:
yep you can copy paste, but still this is not something I would consider user friendly.
it's way to complicated. would make it hard to debug if someone has install issues. so imo we should provide an info as simple as possible :tm: 
but if this current info is wrong lets get rid of it, sure.

hlmnrmr:
We should keep it real and simple at this very moment. The current README does not work, that's a fact. So, removing Docker and MacOS info is fine, and even if those would work well, we should keep that in a separate file or place (e.g. CONTRIBUTING.md, Wiki) as they are mainly resources for developers. In this README we should keep info on the platform we officially support (i.e. Ubuntu).

@naspeh I agree with @petrjasek only linking to the `fireq` docs doesn't work too well. Installation steps should live here and in the simplest possible way, and yes... for further details we can reference `fireq`. This is a good start though.

naspeh:
@hlmnrmr just updated README regarding your comment. It's ok for now?

baltedewit:
installation was smooth using this script on Google Cloud Computer Engine.

definitely missing out a lot of documentation on how to configure it properly. what happened to aws for storage etc.

naspeh:
@baltedewit, if you are using [fireq](https://github.com/superdesk/fireq) scripts, specify your aws settings in `/etc/superdesk.sh` file, list of settings are here: https://superdesk.readthedocs.io/en/latest/settings.html#amazon-s3-settings

and then run `supervisorctl restart all`

baltedewit:
@naspeh thanks, that's exactly what I was looking for. superdesk.readthedocs.io should be mentioned in the readme btw.

PR #2488, opened 2017-02-07
fix(localization): Fix the service vs category issue (SDNTB-374)

ioanpocol:


ioanpocol:
Moved the settings on client core

PR #2503, opened 2017-02-16
[SDESK-769] Adding reference to superdesk.macros module

nrvikas:


nrvikas:
Closing this as it is handled in superdesk-core module.

PR #2510, opened 2017-02-20
Upgrade Elasticsearch to 2.4

vied12:


docker/common.yml
```diff
@@ -19,7 +19,7 @@ kibana:
   restart: always
 
 elastic:
-  image: elasticsearch:1.5
+  image: elasticsearch:2.4
```
petrjasek: imo 1.7 would be better for now

PR #2620, opened 2017-06-05
[Data] set of backend data to initialize TheSource instance

jerome-poisson:
SDTS-44


jerome-poisson:
WIP please don't merge now, there are trouble when using these data.

PR #2624, opened 2017-06-06
cores now use 1.8rc3

jerome-poisson:


client/package.json
```diff
@@ -2,6 +2,6 @@
     "name": "superdesk",
     "license": "GPL-3.0",
     "dependencies": {
-        "superdesk-core": "superdesk/superdesk-client-core#e100036"
+        "superdesk-core": "superdesk-core": "1.8.0-rc.1"
```
petrjasek: imo this won't work, just use `"superdesk-core": "1.8.0-rc.1"`

jerome-poisson: oh, I have to blame either my finger or my copy/paste ability :)

PR #2662, opened 2017-07-12
Use new publisher module

jerome-poisson:


jerome-poisson:
node waiting for https://github.com/superdesk/superdesk-core/pull/985 to update core, else this P.R. will fail

PR #2699, opened 2017-09-19
chore(cores): Update superdesk cores

ioanpocol:


petrjasek:
I think you will have to remove analytics from packages, otherwise it will get old webpack version

naspeh:
also needs to remove `analytics` here [client/superdesk.config.js#L9](https://github.com/superdesk/superdesk/blob/master/client/superdesk.config.js#L9)

PR #2809, opened 2018-01-31
Always build the latest client on master instance

tomaskikutis:


client/package.json
```diff
@@ -3,7 +3,7 @@
     "license": "GPL-3.0",
     "dependencies": {
         "superdesk-analytics": "superdesk/superdesk-analytics#963c6e1",
-        "superdesk-core": "superdesk/superdesk-client-core#6c00eb6",
+        "superdesk-core": "superdesk/superdesk-client-core",
```
petrjasek: I think you can keep `#master` there to make it more explicit

PR #2855, opened 2018-03-29
[MACROS] planning item workflow state change on assign id

jerome-poisson:
This macro change the workflow of planning item from "assigned" to
"in_progress" if the item as a corresponding "assign_id"

SDANSA-131


jerome-poisson:
moved to core after checking with @karelpetrak

PR #2890, opened 2018-05-23
[ROUTING] remove all marks when closing desk

jerome-poisson:
this patch modify previous one by removing all marks of items in a desk,
not only the one of closed destination, because destination can change
in case of chaining, and user expect to see all mark for the desk
removed.

SDANSA-154


petrjasek:
does it leave marks done by users?

jerome-poisson:
@petrjasek now yes ;)

PR #2965, opened 2018-09-26
feat(autorefresh): Activate autorefresh feature (SDNTB-520)

ioanpocol:


client/superdesk.config.js
```diff
@@ -38,8 +40,10 @@ module.exports = function(grunt) {
             preview: 1,
             swimlane: {columnsLimit: 4},
             editor3: true,
-            editorHighlights: true
+            editorHighlights: true,
+            autorefresh: true,
```
petrjasek: imo the name could indicate what is auto refreshed, like monitoringAutorefresh or something

ioanpocol: autorefreshContent

PR #3001, opened 2018-11-21
SDESK-3548 Available publishing formats

ride90:


server/data/vocabularies.json
```diff
@@ -0,0 +1,151 @@
...
+                    {
+                        "name": "Email",
+                        "qcode": "email"
+                    },
+                    {
+                        "name": "Adobe InDesign (IDML)",
+                        "qcode": "idml"
+                    },
+                    {
+                        "name": "AAP NewsML 1.2",
```
petrjasek: we should probably call it just `NewsML 1.2`

ride90: @petrjasek ok

PR #3003, opened 2018-11-22
[DOCKER] set names to containers

jerome-poisson:
Docker is now adding an unique id at the end of container names. This
patch set fixed names to containers, so they are more easy to manipulate.


jerome-poisson:
@ride90 well, it's obviously a copy/paste mistake, I'll fix it thanks :)

PR #3023, opened 2019-01-06
#3003 Fix wrong path in Dockerfile when install client. Fix start.sh remove command install lib when docker compose up container, only run process. Add missing docker images in common.yml

vokye700:


docker/common.yml
```diff
@@ -10,6 +10,18 @@ elastic:
...
+kibana:
+  image: kibana:5
+  container_name: kibana
+
+logstash:
+  image: logstash:5
+  container_name: logstash
+
+pubapi:
+  image: willypublicitas/pubapi
```
petrjasek: what is this pubapi?

docker/start-dev.sh
```diff
@@ -1,10 +1,6 @@
 #!/bin/bash
-cd /opt/superdesk/client-core &&
-npm link &&
```
petrjasek: without linking this won't be dev setup

docker/common.yml
nickdbush: Any ideas?

PR #3038, opened 2019-02-01
chore(pictures): Set status in-progress for existing pictures

ioanpocol:


ioanpocol:
Not needed here because of:
https://github.com/superdesk/superdesk/pull/3040

PR #3048, opened 2019-02-13
chore(macros): Update chores in order to get new macros (SDANSA-222)

ioanpocol:


petrjasek:
we should stick to those branches, I will pick the fixes there, I don't want to use master in production

PR #3050, opened 2019-02-14
feat(macro): Add translate macro for ANSA (SDANSA-222)

ioanpocol:


server/ansa/macros/translate_text.py
```diff
@@ -0,0 +1,35 @@
...
+
+"""
+    Translate text from body_html
+"""
+
+
+def translate(text=''):
+    URL_TRANSLATION = app.config["ANSA_TRANSLATION_URL"]
+    headers = {'Content-Type': 'application/x-www-form-urlencoded'}
+    data = {'text': text, 'lang': 'en', 'target': 'it'}
```
petrjasek: they might also translate the other way around, would make sense to translate to it if `item.language` is en and vice versa

PR #3051, opened 2019-02-18
Update cores

takeit:


takeit:
@petrjasek any idea why the fireq fails?

PR #3052, opened 2019-02-18
Update Superdesk cores and enable nestedItemsInOutputStage

takeit:


petrjasek:
there is a fix for nested items which is not merged yet, you should wait a bit

takeit:
okay, would it be available today?

takeit:
fix ref. https://github.com/superdesk/superdesk-client-core/pull/2838

PR #3088, opened 2019-05-16
keep original item guid in the output

petrjasek:
so on update it will keep guid in newsml

SDANSA-237


server/ansa/formatters/ansa_newsml_g2_formatter.py
```diff
@@ -270,8 +270,22 @@ def _format_item_set(self, article, item_set, item_type):
...
+            item.set('guid', self._get_original_guid(article))
+            item.set('version', str(article.get('rewrite_sequence') + 1))
+        else:
+            item.set('version', str(1))
         return item
 
+    def _get_original_guid(self, article):
+        guid = article['rewrite_of']
+        for i in range(article.get('rewrite_sequence', 1)):
+            prev = superdesk.get_resource_service('archive').find_one(req=None, _id=guid)
```
ride90: just a small note,
I would add a projection here:
```
req = ParsedRequest()
req.projection = '{"rewrite_of": 1}'
```

petrjasek: will handle it in another PR

PR #3100, opened 2019-06-25
[DOCKER] added grammalecte Dockerfile + added the service to docker-c…

jerome-poisson:
…ompose files

SDBELGA-131

### DIFF
--- a/docker/common.yml
+++ b/docker/common.yml
@@ -17,6 +17,10 @@ postfix:
   - maildomain=mail.sourcefabric.org
   - smtp_user=user:pwd
 
+grammalecte:
+  build: grammalecte/
+  container_name: grammalecte
+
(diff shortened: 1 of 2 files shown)


petrjasek:
shouldn't this be done on superdesk/superdesk-belga repo?

jerome-poisson:
@petrjasek I think a French grammar/spellchecker is generally useful, not only for Belga.

jerome-poisson:
@petrjasek so what shoudl I do? An other option is to keep the grammalecte Dockerfile here, but not adding it to `docker-compose-services.yml` nor `docker-compose.yml` , and only adding it for custom projects which need it (i.e. Belga).

petrjasek:
sounds good :+1:

PR #3108, opened 2019-08-06
Ansa add buttons for adding media from monitoring list items

tomaskikutis:
SDANSA-270

demo: https://drive.google.com/open?id=1n-eNuO_P9svHR2rC0aYnBY1V4LgGFNEC


client/extensions/imageShortcuts/src/get-widgets.tsx
```diff
@@ -0,0 +1,108 @@
...
+            this.state = {
+                // components can be remounted even if `articleEditStart` /  `articleEditEnd` haven't fired
+                // for example if view type is being toggled between list/grid or more items are loaded.
+                isDisplayed: typeof contentProfile !== 'undefined',
+            };
+
+            this.show = (article: IArticle) => {
+                if (getContentProfilePromise === undefined) {
+                    getContentProfilePromise = superdesk.entities.contentProfile.get(article.profile).then((profile) => {
+                        if ((profile.schema.hasOwnProperty('feature_media') || profile.schema.hasOwnProperty('media-gallery'))) {
```
petrjasek: that media gallery field they use has id `photoGallery`

petrjasek: would be good to use constants for both

PR #3119, opened 2019-09-24
Fixes for adding items to gallery and feature media

tomaskikutis:
SDANSA-270

Media added from related items widget didn't perform validation. Existence of the same item or max upload count was not being taken into account. I changed the code to use the extensions API so the same logic is used as in the core.


client/superdesk.config.js
```diff
@@ -143,5 +143,9 @@ module.exports = function(grunt) {
             imageShortcuts: 1,
             ansaIptc: 1,
         },
+
+        enabledExtensions: {
```
petrjasek: I don't think this should be here

tomaskikutis: Removed :)

PR #3134, opened 2019-11-25
Set SD stable version & added vocabularies.json

takeit:
Superdesk version will be the latest stable.
Superdesk Publisher version will be the latest one from `master` branch


takeit:
@petrjasek what's wrong with the failing tests? can it be safely merged?

petrjasek:
looks like an issue with superdesk-publisher dependencies, so no idea

takeit:
@tomaszrondio can you take a look?

PR #3135, opened 2019-11-26
fix(build): Make master a vanilla Superdesk install

MarkLark86:
Optional Superdesk modules such as Planning, Analytics and Publisher should use separate branches from the Vanilla Superdesk builds

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -2,8 +2,6 @@
     "name": "superdesk",
     "license": "GPL-3.0",
     "dependencies": {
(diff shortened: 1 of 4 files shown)


MarkLark86:
@petrjasek I propose that the master branch of this repository should be reserved for a vanilla version of Superdesk and optional modules should use a separate branch

petrjasek:
@hlmnrmr @karelpetrak any feedback on this?

PR #3145, opened 2020-02-25
Update planning publisher

ride90:


client/extensions/helloWorld2/npm-debug.log
```diff
@@ -0,0 +1,48 @@
+0 info it worked if it ends with ok
```
petrjasek: this file shouldn't be there we can probably ignore `*.log`

tomaskikutis: `npm-debug.log` shouldn't be committed either

server/Procfile
```diff
@@ -3,3 +3,4 @@ wamp: python3 -u ws.py
 work: celery -A worker worker
 beat: celery -A worker beat --pid=
 capi: gunicorn -c gunicorn_config.py content_api.wsgi
+highcharts: python3 -u -m analytics.reports.highcharts_server
```
petrjasek: seems like analytics is not installed so this won't work

client/extensions/helloWorld2/npm-debug.log
ride90: my bad, I did not notice it somehow..

server/requirements.txt
```diff
@@ -2,6 +2,5 @@ gunicorn==19.7.1
 honcho==1.0.1
 newrelic>=2.66,<2.67
 
--e git+git://github.com/superdesk/superdesk-core.git@f38aaee#egg=Superdesk-Core
--e git+git://github.com/superdesk/superdesk-planning.git@master#egg=superdesk-planning
-
+-e git+git://github.com/superdesk/superdesk-core.git@v1.32.5#egg=Superdesk-Core
```
petrjasek: you can use `superdesk-core>=1.32.5,<1.33` for this one. but maybe it should just use latest stable

PR #3159, opened 2020-04-28
Use released version of planning (ansa)

MarkLark86:


petrjasek:
actually shouldn't it be rather master? here we need stable version

PR #3160, opened 2020-04-28
Use master branch for planning (planning-publisher)

MarkLark86:


petrjasek:
here I think we need stable

PR #3174, opened 2020-06-01
process legal archive items and remove its media

petrjasek:
only keep viewImage and only if that item is no longer in archived

### DIFF
--- a/server/ansa/commands/remove_expired_media.py
+++ b/server/ansa/commands/remove_expired_media.py
@@ -8,20 +8,20 @@ from ansa.remove_expired_media import remove_expired_media
 class RemoveExpiredMediaCommand(superdesk.Command):
 
     option_list = (
-        superdesk.Option('--days', dest='days', default=50),
-        superdesk.Option('--skip', dest='skip', default=0),
-        superdesk.Option('--limit', dest='limit', default=0),
+        superdesk.Option('--days', dest='days', type=int, default=50),
+        superdesk.Option('--skip', dest='skip', type=int, default=0),
+        superdesk.Option('--limit', dest='limit', type=int, default=0),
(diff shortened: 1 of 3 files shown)


server/ansa/commands/remove_expired_media.py
```diff
@@ -0,0 +1,36 @@
...
+
+from datetime import datetime, timedelta
+from ansa.remove_expired_media import remove_expired_media
+
+
+class RemoveExpiredMediaCommand(superdesk.Command):
+
+    option_list = (
+        superdesk.Option('--days', dest='days', default=50),
+        superdesk.Option('--skip', dest='skip', default=0),
```
ride90, line 12: I think we can use `type` argument for `skip`, `limit` and `days`

server/ansa/commands/remove_expired_media.py
```diff
@@ -0,0 +1,36 @@
...
+            .skip(int(skip)) \
+            .limit(int(limit))
+
+        archived_service = superdesk.get_resource_service('archived')
+
+        i = 0
+        for item in cursor:
+            if item.get('associations') or item.get('renditions'):
+                i += 1
+                doc = item.copy()
```
ride90, line 32: why do we need a copy here?

ride90:
@petrjasek how we preserve `viewImage`?

petrjasek:
> @petrjasek how we preserve `viewImage`?

that's in https://github.com/superdesk/superdesk/blob/8bf8a6615abaf38407818b9709437a6b10655696/server/ansa/remove_expired_media.py#L13

server/ansa/commands/remove_expired_media.py
petrjasek: that was copy on write..

PR #3175, opened 2020-06-01
add migration to fix missing desk on published items (#3173)

petrjasek:
and attach these to image desk

SDANSA-346


server/data_updates/00023_20200529-155751_archive.py
```diff
@@ -0,0 +1,47 @@
...
+class DataUpdate(DataUpdate):
+
+    resource = 'archive'
+
+    def forwards(self, mongodb_collection, mongodb_database):
+        service = get_resource_service(self.resource)
+        desk = mongodb_database['desks'].find_one({'name': re.compile(r'^XFV')})  # image desk
+        if not desk:
+            logger.warning('desk XFV not found')
+        ids = list(mongodb_collection.find({'state': 'in_progress', 'task.desk': None}, {'_id': True}))
```
ride90, line 29: `mongodb_collection` points to `archive`, I think in this case we need a `published` collection.
Why do query `in_progress` items?

petrjasek: good catch, that makes no sense, I think I was testing and didn't have any more published items left for it.. :man_facepalming:

PR #3179, opened 2020-06-09
Send to Panel changes. [SDANSA-388]

GyanP:


client/superdesk.config.js
```diff
@@ -91,6 +91,9 @@ module.exports = function(grunt) {
         ui: {
             italicAbstract: false,
             sendAndPublish: true,
+            sendPublishSchedule: true,
+            sendEmbargo: true,
```
petrjasek, line 95: I think they want these 2 disabled

PR #3181, opened 2020-06-10
implement ansa photo archive management

petrjasek:
there is new delete action available for ansa
photo archive items.

SDANSA-402


client/extensions/ansa-archive/src/extension.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const extension: IExtension = {
+    id: 'ansa-archive',
+    activate: (superdesk: ISuperdesk) => {
+        const result: IExtensionActivationResult = {
+            contributions: {
+                entities: {
+                    article: {
+                        getActionsBulk: (articles) => {
+                            const actions: IArticleActionBulk[] = [];
```
tomaskikutis, line 15: It'd be nice to stick to `Array<IArticleActionBulk>` format. It's enforced by a linter on the client, but not here.

petrjasek: what's the difference?

petrjasek:
> Is spiking allowed for these external ansa items? It's a bit weird that external items appear in our internal lists, they probably shouldn't or should they? Does their structure match `IArticle`?

it's not our spiking, but using their api to delete those external files. I suppose they don't have UI to manage that archive, or it's in the old system they are trying to replace

client/extensions/ansa-archive/src/extension.ts
tomaskikutis: None, it's just a formatting choice.

tomaskikutis:
It's a bit weird that external items appear in our internal lists, they probably shouldn't or should they? Does their structure match `IArticle`?

petrjasek:
it does match. this is items we have available via custom search provider, it's like a proxy to external archive

client/extensions/ansa-archive/src/extension.ts
petrjasek: I see, ok makes sense to be consistent

PR #3186, opened 2020-07-08
Ninjs formatter - DateCreated field is not populated with data. [SDANSA-423]

GyanP:
Since I can't reproduce this issue, so I don't know the actual source of the issue.


client/extensions/ansaIptc/src/extension.ts
```diff
@@ -61,7 +61,7 @@ const extension: IExtension = {
                         const created = parseDatetime(data.DateCreated, data.TimeCreated);
                         const release = parseDatetime(data.ReleaseDate, data.ReleaseTime);
 
-                        if (created) {
+                        if (created && created !== '') {
```
petrjasek, line 64: I think `''` is falsy so this change doesn't do much

PR #3191, opened 2020-09-22
Fix migration for single selection vocabularies

pablopunk:
SDESK-5445

This migration was making usageterms to be multiselection


server/data_updates/00005_20181114-090110_vocabularies.py
```diff
@@ -17,7 +17,9 @@ class DataUpdate(DataUpdate):
     def forwards(self, mongodb_collection, mongodb_database):
         vocabularies_service = get_resource_service('vocabularies')
         for vocabulary in vocabularies_service.get(req=None, lookup=None):
-            if vocabulary.get('single_value', False):
+            if hasattr(vocabulary, 'selection_type'):
```
petrjasek, line 20: I would just skip it, like:
```py
if vocabulary.get('selection_type'):
    continue
```

petrjasek (approved):

PR #3204, opened 2020-10-21
Use develop deps on server

pablopunk:
notice is for branch `ansa-develop`

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -7,5 +7,5 @@ flask-babel==0.12.2
 pymongo==3.8.0
 werkzeug<1.0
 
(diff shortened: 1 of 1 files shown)


petrjasek:
why we need this?

pablopunk:
@petrjasek well this branch is `ansa-develop` and those deps are not updated to the latest develop changes. Also, I'm trying to get [SDANSA-422](https://dev.sourcefabric.org/browse/SDANSA-422) tested by QA on an ansa instance so this could do it, but there's some issue in the deployment

petrjasek:
I see, I think that should be already available on sd-ansa, imo we should rather drop this branch

pablopunk:
@petrjasek sure if we don't need it go ahead

PR #3210, opened 2020-11-10
update docker config

petrjasek:
- add service for superdesk-server based on python
- use multi stage build for client
- test docker-compose build via github actions


docker/docker-compose.yml
```diff
@@ -1,61 +1,86 @@
-logstash:
-  extends:
-    file: common.yml
-    service: logstash
-  links:
-  - elastic
+version: "2.4"
```
jerome-poisson, line 1: if we do a cleaning, why not starting directly with latest version (3.x)?

docker/docker-compose.yml
```diff
@@ -1,61 +1,86 @@
...
+    - smtp_user=user:pwd
 
-postfix:
-  extends:
-    file: common.yml
-    service: postfix
+  superdesk-server:
+    build: ../server/
+    container_name: superdesk-server
+    links:
```
jerome-poisson, line 38: if working with docker-file 3.x links are not necessary anymore (we just need to set `depents_on`, and the default network setup should be fine).

jerome-poisson (approved):
looks good, just thinking that we should move to more recent version of docker.

PR #3213, opened 2020-11-19
Fix the validation

GyanP:


server/ansa/validate.py
```diff
@@ -208,9 +208,10 @@ def validate(sender, item, response, error_fields, **kwargs):
     if user and is_user_external(user) and not mask.get(Validators.PRODUCT_ALLOWED):
         response.append("Products not allowed to external User")
 
-    desk = superdesk.get_resource_service('desks').find_one(
-        req=None, _id=item['task']['desk']
-    )
+    if item.get('task') and item['task'].get('desk'):
+        desk = superdesk.get_resource_service('desks').find_one(
+            req=None, _id=item['task']['desk']
+        )
```
petrjasek, line 214: you should add `desk = None` before the if or into else so that variable will be set always

PR #3220, opened 2021-02-10
Update extensions registration

tomaskikutis:


petrjasek:
will fix those server issues, but what about those client errors?

tomaskikutis:
I forgot to remove the id from the sample extension. Client tests are passing now.

PR #3222, opened 2021-02-16
Simplify Docker configuration

nickdbush:
As the new containers have been released to Dockerhub, making a `docker-compose.yml` file easily available seems like the next best step. I've replaced the current compose file, which builds the containers from source, with a new compose file which uses the prebuilt images. The old configuration is retained for development purposes in `docker-compose-dev.yml`.


nickdbush:
It would be helpful if someone could run the `docker-compose.yml` file on their machine just so we can have peace of mind that it will work for (hopefully) everyone.

petrjasek:
I was doing something similar but didn't manage to finish it, you can check it on [docker](https://github.com/superdesk/superdesk/tree/docker) branch. so instead of 2 docker-compose files there is override which just sets the build, the rest is more or less the same. can you update your PR with those changes?

nickdbush:
Updated @petrjasek

PR #3259, opened 2021-09-15
build tools 2.0.1

tomaskikutis:


client/package.json
```diff
@@ -9,6 +9,6 @@
         "build": "npx @superdesk/build-tools build-root-repo ./"
     },
     "devDependencies": {
-        "@superdesk/build-tools": "^1.0.18"
+        "@superdesk/build-tools": "2.0.1"
```
petrjasek, line 12: I would still use `^` so we don't have to change it on every update

PR #3276, opened 2022-02-16
testing superdesk-ui-framework 2.4.18

dev-tsoni:


client/package.json
```diff
@@ -3,7 +3,7 @@
     "name": "superdesk",
     "license": "AGPL-3.0",
     "dependencies": {
-        "superdesk-core": "github:superdesk/superdesk-client-core#develop"
+        "superdesk-core": "github:superdesk/superdesk-client-core#1df1b6d9966f52f2e7dce101b22131e26df2a2b9"
```
MarkLark86, line 6: It's best of keeping this on `develop`, otherwise we will need to update this when there are any changes done in `cilent-core`.

In this scenario, we would wait for your `client-core` PR to be merged then we will rebuild the `sams-develop` instance

PR #3278, opened 2022-04-09
Update docker-compose.yml

Mahdi-Ba:
set env SUPERDESK_URL because upload URL media to correct store and fetch in Mongo DB .
if don't do this upload media starting with => localhost:5000


petrjasek:
thanks!

PR #3494, opened 2024-11-07
Add dependabot on release branch

tomaskikutis:


petrjasek:
it only works on 1 branch per repo, so with that change it would stop working on develop

tomaskikutis:
https://github.com/dependabot/dependabot-core/issues/388

Are you sure? I read about people being able to configure it. Also it appears the feature was merged in 2018

petrjasek:
we can test this, don't remember now if it wasn't working for security updates or version updates

client/index.ts
```diff
@@ -3,10 +3,6 @@ import {startApp} from 'superdesk-core/scripts/index';
 setTimeout(() => {
     startApp(
         [
-            {
-                id: 'broadcasting',
-                load: () => import('superdesk-core/scripts/extensions/broadcasting'),
```
petrjasek, line 8: is it broken?

tomaskikutis: it's duplicated, the same thing is imported on line 23, except with custom configuration so I chose to keep that

PR #3506, opened 2024-11-19
Fix `docker compose` commands.

jerome-poisson:
- `compose` is now a sub-command of `docker`, so `docker compose` should be used instead of `docker-compose`.

- The initialization commands used `docker-compose run` which starts a new container, while at this point the container should be already running, so `exec` should be used instead.

- Removed erroneous `run` commands used before `python manage.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Find more information about the client configuration in the README file of the r
 
 ## Run Superdesk locally using Docker
 
-You can start superdesk using the `docker compose.yml` file:
+You can start superdesk using the `docker-compose.yml` file:
 
 ```sh
 $ docker compose up -d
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,20 +18,20 @@ Find more information about the client configuration in the README file of the r
 
 ## Run Superdesk locally using Docker
 
-You can start superdesk using the `docker-compose.yml` file:
+You can start superdesk using the `docker compose.yml` file:
```
petrjasek, line 21: I think here it should actually stay

jerome-poisson: Oh good catch.

petrjasek (approved):

PR #3701, opened 2025-05-23
Update planning

thecalcc:
SDESK-7639


petrjasek:
there is newer version already

PR #3760, opened 2025-06-26
update package-lock

devketanpro:


petrjasek:
pls don't use node 14 to update package-lock, that would break dependabot updates in future, you can trigger dependabot update via insights (or wait for tomorrow)

PR #3828, opened 2025-08-06
fix docker build

petrjasek:
use docker build instead of docker compose build


tomaskikutis (approved):
what was the issue with `docker compose build`?

petrjasek:
looks like issue in docker compose https://github.com/docker/compose/issues/12998

petrjasek:
actually it should be fixed in https://github.com/actions/runner-images/releases/tag/ubuntu24%2F20250804.2

PR #4291, opened 2026-09-21
Add Graylog logging via graypy GELF UDP handler

IvanJelicSF:
### Summary

Sends server logs to Graylog over GELF UDP using [graypy](https://pypi.org/project/graypy/). The root logger now writes to both the console and Graylog, so logs from the API, the websocket server and the Celery workers all end up in Graylog.

### Changes

- **`server/requirements.in` / `requirements.txt`**: add `graypy` (pinned to `2.1.0`).
- **`server/logging_config.yml`**:
  - New `graylog` handler (`graypy.GELFUDPHandler`) at `INFO` level. It reads host, port and facility from `settings` via `ext://`.
  - New `graylog` formatter (`%(message)s`). The traceback is appended to the message so it shows up in the Graylog message list. graypy still sends it separately in `full_message`.
  - The root logger's handlers are now `[console, graylog]`.
- **`server/settings.py`**:
  - New settings, read from env variables:
    | Env variable       | Default     |
    |--------------------|-------------|
    | `GRAYLOG_HOST`     | `localhost` |
    | `GRAYLOG_PORT`     | `12201`     |
    | `GRAYLOG_FACILITY` | `superdesk` |
  - `CELERY_WORKER_HIJACK_ROOT_LOGGER = False`, so Celery workers keep the root handlers from `logging_config.yml` instead of replacing them with Celery's own.

### Notes

- The Graylog handler is always on. It uses UDP, which doesn't wait for a reply, so an instance with no Graylog listening keeps running normally and the log packets are simply dropped.
- Deployments with their own `LOG_CONFIG_FILE` are unaffected.
- With `CELERY_WORKER_HIJACK_ROOT_LOGGER` turned off, worker log output follows `logging_config.yml` rather than Celery's default format.

Resolves [SDESK-8032](https://sofab.atlassian.net/browse/SDESK-8032)

[SDESK-8032]: https://sofab.atlassian.net/browse/SDESK-8032?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


petrjasek:
I would probably do it in superdesk-core and only configure it when those env variables are set, or is it working like that automatically?

MarkLark86:
> I would probably do it in superdesk-core and only configure it when those env variables are set, or is it working like that automatically?

Agree'd. I think it would be better as a core feature, and handled via configs.
[configure_logging](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/logging.py#L36) is called from app_factory using [configure_logging(app.config["LOG_CONFIG_FILE"])](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/factory/app.py#L644).

Maybe we could modify the `configure_logging` function to accept the full log dictionary, and use the configs to determine if we enable the graylog handlers there (and `graypy` is installed).

IvanJelicSF:
> > I would probably do it in superdesk-core and only configure it when those env variables are set, or is it working like that automatically?
> 
> Agree'd. I think it would be better as a core feature, and handled via configs. [configure_logging](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/logging.py#L36) is called from app_factory using [configure_logging(app.config["LOG_CONFIG_FILE"])](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/factory/app.py#L644).
> 
> Maybe we could modify the `configure_logging` function to accept the full log dictionary, and use the configs to determine if we enable the graylog handlers there (and `graypy` is installed).

I actually think it's good to have both, in case we have an older version we need this for, so we can apply without backporting the core change.

MarkLark86:
> > > I would probably do it in superdesk-core and only configure it when those env variables are set, or is it working like that automatically?
> > 
> > Agree'd. I think it would be better as a core feature, and handled via configs. [configure_logging](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/logging.py#L36) is called from app_factory using [configure_logging(app.config["LOG_CONFIG_FILE"])](https://github.com/superdesk/superdesk-core/blob/develop/superdesk/factory/app.py#L644).
> > 
> > Maybe we could modify the `configure_logging` function to accept the full log dictionary, and use the configs to determine if we enable the graylog handlers there (and `graypy` is installed).
> 
> I actually think it's good to have both, in case we have an older version we need this for, so we can apply without backporting the core change.

Fair enough.

So step 1 is enable it here (and copied to customer repos where required)
Step 2 would be make it a core feature

petrjasek:
I'm just wondering what happens when we merge it here and then add it to core, would it start logging twice?
might be better to put it somewhere where we're not using latest core

IvanJelicSF:
> I'm just wondering what happens when we merge it here and then add it to core, would it start logging twice?
> might be better to put it somewhere where we're not using latest core

I wouldn't do it both places, it's either Core or distro. For example, we do it in Core for 3.7+, others in the distro, or something like that. I'll have the PR for the Core later today.

petrjasek:
yep exactly, so wouldn't merge it here because here it will get it from core once we merge your PR

IvanJelicSF:
> yep exactly, so wouldn't merge it here because here it will get it from core once we merge your PR

Oh yes, absolutely, specially as there will be another PR here if we want web socket logging in Graylog

